PR #18, opened 2020-12-17
Use fetch_member instead of get_member for auto-converting

hpenney2:
Using `get_member` will return `None` if the bot doesn't have the server members intent or hasn't (fully) chunked the guild the member came from. `fetch_member` doesn't require intents, so it's a better choice. This also requires `process_options` to be async so it can call and get the return values of coroutines (in this case, just `fetch_member`). However, this shouldn't cause any issues.

### DIFF
--- a/discord_slash/client.py
+++ b/discord_slash/client.py
@@ -5,6 +5,7 @@ from discord.ext import commands
 from . import http
 from . import model
 from .utils import manage_commands
+from inspect import iscoroutinefunction
 
 
 class SlashCommand:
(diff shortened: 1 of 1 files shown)


eunwoo1104:
The problem with `fetch` things is that if bot fails fetching it will raise exception instead of returning `None`. However getting from cache gives `None` if it failed finding a user. But I'll reconsider before closing or merging.

eunwoo1104:
Merging to new branch, thank you for contributing!

PR #22, opened 2020-12-18
Dispatch an event on valid command

PredaaA:
That PR add a dispatch `slash_command` event, with command name, and registered data about the command.


eunwoo1104:
There actually is some problem with this changes in what I saw.
1. It dispatches __before__ checking the guild id.
2. `selected_cmd` can be almost empty when the command is subcommand.

These should be fixed before merging.

PredaaA:
I fixed both things you mentionned. I'm just unsure if how I handled subcommands dispatch is what you'd prefer.

eunwoo1104:
It looks fine now. Merging this pull request.
Thank you for contributing!

PR #28, opened 2020-12-19
Make README Easier to read

stellaurora:
**Reason:**
I felt like the Note wasn't the most "grammatically" correct

i have no idea what to put here
big word make me feel smart


eunwoo1104:
Actually I'm not that good at English, so I think there was some grammar mistakes. Thank you for contributing!

stellaurora:
haha, well I don't believe I'm the best at it either hopefully I didn't make too many mistakes!

PR #33, opened 2020-12-22
fix: retry bug in get_all_commands utility

maduck:
Fixes potential crash on 429 when attempting to retrieve all commands.


eunwoo1104:
Well, my mistake. Thank you for contributing!

PR #36, opened 2020-12-24
Fix error raised when value ctx.guild is type int

4surix:


eunwoo1104:

4surix:
You're welcome, I am happy to help !

PR #41, opened 2020-12-28
Add functions to removes all commands.

4surix:
Hello !  
  
I made two functions to delete all commands everywhere or just in one area (global or guild).  

### Why ?
I did this because I wanted to delete all the commands from a server and then put back only the ones I wanted, so that I could remove the ones that no longer exist and add the new ones.  
  
### Exemple :
```python
import discord
import discord_slash


bot = discord.Client()
slash = discord_slash.SlashCommand(bot)

@bot.event
async def on_ready(self):

    await slash.remove_all_commands()

    # Useless, because `remove_all_commands` already removes the global commands,
    #  just for the exemple.
    await slash.remove_all_commands_in('global')

    # As explained I then call up the existing `register_all_commands` function 
    #  to add the commands I want.
    await slash.register_all_commands()

bot.run(token)
```

Thank you for reading, have a nice day !

### DIFF
--- a/discord_slash/client.py
+++ b/discord_slash/client.py
@@ -169,7 +169,7 @@ class SlashCommand:
 
     async def remove_all_commands(self):
         """
-        Remove all slash commands except subcommands to Discord API.
+        Remove all slash commands.
         """
 
         await self.remove_all_commands_in('global')
@@ -181,7 +181,7 @@ class SlashCommand:
 
     async def remove_all_commands_in(self, area):
         """
-        Remove all slash commands in area except subcommands to Discord API.
+        Remove all slash commands in area.
 
         :param area: 'global' or guild ID where removing all commands.
         :type area: Union[str, int]
--- a/discord_slash/client.py
+++ b/discord_slash/client.py
@@ -4,6 +4,7 @@ import discord
 from discord.ext import commands
 from . import http
 from . import model
+from . import error
 from .utils import manage_commands
 from inspect import iscoroutinefunction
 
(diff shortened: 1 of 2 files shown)


eunwoo1104:
This is good function, but there are some questions.
1. This doesn't affect the SlashCommand itself. This should be in the `.utils.manage_commands`.
2. Why did you use `for x in self.commands.keys():`? This isn't necessary, and it is wasting some times.

4surix:
Actually as `register_all_commands` was in `SlashCommand`,  I thought it was the right place because `remove_all_commands` does the opposite of `register_all_commands`.  
  
Yes, `for x in self.commands.keys():` is absolutely useless here.  
I think I should go to sleep one day.

eunwoo1104:
> Actually as `register_all_commands` was in `SlashCommand`, I thought it was the right place because `remove_all_commands` does the opposite of `register_all_commands`.

Well... `register_all_commands` reads commands registered in SlashCommand. However, yours read commands list from discord API. That's why I said this should be in the `.utils.manage_commands`.

4surix:
Okay I see, you are right,  
  
So I moved into `.utils.manage_commands` and this is what it looks like :

```python
import discord
import discord_slash
from discord_slash.utils import manage_commands


bot = discord.Client()
slash = discord_slash.SlashCommand(bot)

@bot.event
async def on_ready(self):
    await manage_commands.remove_all_commands(slash)
    await manage_commands.remove_all_commands_in(slash, 'global')
    await slash.register_all_commands()

bot.run(token)
```

It is good ? If yes, i commit this.

eunwoo1104:
Looks good!

eunwoo1104:
\+ I'll trim some code to fit the code style.

PR #45, opened 2021-01-01
Add faq to docs

quirky-bluejay:
This adds the faq in the #faq channel in the discord server to the docs.

I've done a bit more than just copy them over and format them.
I'll try and list all of what i changed compared to the #faq channel
- Expand on `How to add a command` by adding `Why don't my commands show up`
- Add a note on using `auto_delete` and how it makes a request for every guild
- Reformat `Difference between ctx.send and ctx.channel.send?`
  - Seperate it into two sections and expanding on them
  - Add a warning about if the bot is not in the guild or has permissions
- Add a link to the discord server at the bottom

There's one thing that i could not quite figure out how to do 
When referencing other functions, like the utils functions i used `:py:func:` and `:py:attr:` etc.
When i build it with make html, those references are not links, and according to the sphinx docs i looked up there should be a link :confused:
Is this the read the docs theme doing this?

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -21,7 +21,7 @@ Creating a slash command is a two part process.
 
 Create a slash command on Discord  
 ---------------------------------
-Set :py:attr:`SlashCommand.auto_register` to ``True`` when you register the slash command client.
+Set :py:attr:`auto_register <client.SlashCommand.auto_register>` to ``True`` when you register the slash command client.
 
 .. code-block:: python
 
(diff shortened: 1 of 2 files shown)


eunwoo1104:
> When i build it with make html, those references are not links, and according to the sphinx docs i looked up there should be a link

Try using `:func:` or `:class:` or `:attr:` instead. It worked for me.
Anyways it looks great! I'll merge this if you fix this or maybe I'll fix this.

quirky-bluejay:
I tried replacing `:py:func:` with `:func:` and `:py:attr:` with `:attr:`, etc
It still does not seem to work.
I saw where you used `:func:` in client.py and i'm using pretty much the same format.
Not sure what's going wrong.

I also noticed that i put `manage_commands.delete()` instead of `manage_commands.remove_slash_command()` and i've fixed that and a few spelling mistakes

eunwoo1104:
Hmm.... Not sure then. Should I just ignore that and merge?

quirky-bluejay:
I've managed to make all of the internal references work except `auto_register` and `auto_delete`, it turned out that I needed to give the full path.
I can't get the discord references to work.
If you think it's ok to merge then yeah, I'll keep trying to figure this out anyway.

eunwoo1104:
I think it is fine now. Merging this pull request. Thank you for contributing!

PR #55, opened 2021-01-06
Use command docstring as default description

spinnerich:
# Description

Added command docstring as a fallback for the description. This feature is also implemented in discordpy.
(+ fixed a typo)

# Usage

```py
@slash.slash()
async def ping(self, ctx: SlashContext):
    """Wow that's handy!"""
    await ctx.send(content="Pong!")
```
image

# Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

### DIFF
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -1,0 +1,8 @@
+version = 1
+
+[[analyzers]]
+name = "python"
+enabled = true
+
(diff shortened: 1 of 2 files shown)


eunwoo1104:

Issue #52, opened 2021-01-06
Guild id required


quirky-bluejay:
**The Issue**
In the `utils.manage_commands` functions for some commands `None` can be passed for the `guild_id` parameter to specify using a global scope, but the parameter is not optional
It would be expected that if you don't pass a guild id that it would use the global scope, instead of having to pass `None`

**Ideal solution**
Make the guild_id parameter optional

**Expected behaviour**
`utils.manage_commands.get_all_commands(id, token)` should work
ATM you have to do (id, token, None)

eunwoo1104:
Applying that to all of the functions would be hard, since it will require reformatting sequence of parameters. However `get_all_commands` seems possible, don't know why I didn't already apply that to it.

eunwoo1104:
Fixed on #57

PR #57, opened 2021-01-07
Quality improvements

spinnerich:
# Description

This brings general quality improvements.

- Merging isinstance calls
- Removing unnecessary dict.keys()
- Removed unnecessary __init__ of RequestFailure (because it does not change base classes init)
- More Specific IncorrectFormat error message
- Simplifying check for send_type
- Inline comments should contain 2 spaces before # (according to PEP)
- guild_id defaults to None (get_all_commands) #52


eunwoo1104:

PR #59, opened 2021-01-09
Added auto-generation of command options (+ fixes)

spinnerich:
# Description

Added auto-generation of command options. This works through argument type hints (similar behavior to [discord.py's converters](https://discordpy.readthedocs.io/en/latest/ext/commands/commands.html?highlight=converters#converters))
This feature is a fallback if no options are passed directly. Note that the choices option and individual description for arguments are currently not supported.

Additional changes:
- Now SlashCommandOptionType is used more often instead of numbers
- using "or" clause instead of "var1 if var1 else var2"
- Adding conversion of type (hints) to SlashCommandOptionType if possible
- Fixing: IncorrectFormat would be raised if no embeds were passed
- Updating copyright year
- Auto description is now available for cog_ext
- Create a function for the generation of auto_dict from options

# Usage

```py
@slash.slash()
async def mention(ctx: SlashContext, user: discord.User, message: typing.Optional[str] = ""):
    """Mention a given user with an optional message"""
    await ctx.send(content=f"{user.mention} {message}")
```

image

# Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

### DIFF
--- a/discord_slash/client.py
+++ b/discord_slash/client.py
@@ -286,7 +286,7 @@ class SlashCommand:
         :type auto_convert: dict
         :param guild_ids: List of Guild ID of where the command will be used. Default ``None``, which will be global command.
         :type guild_ids: List[int]
-        :param options: Options of the slash command.
+        :param options: Options of the slash command. This will affect ``auto_convert`` and command data at Discord API. Default ``None``.
         :type options: list
         :param has_subcommands: Whether it has subcommand. Default ``False``.
         :type has_subcommands: bool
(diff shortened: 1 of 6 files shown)


eunwoo1104:

eunwoo1104:
Hey, I was testing with this version, and I found out that `typing.get_origin`, etc. used at this pull request is not supported for Python 3.7 or lower, and some other things used at this raises some exception. Sorry, but it looks like I need to revert this pull request to fit the version requirements.

spinnerich:
> Hey, I was testing with this version, and I found out that `typing.get_origin`, etc. used at this pull request is not supported for Python 3.7 or lower, and some other things used at this raises some exception. Sorry, but it looks like I need to revert this pull request to fit the version requirements.

Okay then. Since I'm on 3.8, I did not notice that. I am going to find a compatible way of doing this and then create a pull request again.

PR #63, opened 2021-01-11
Update client.py

ghost:
## Changes

    - Issue:
- [ ] (If required) Document string is updated/added.

## About this pull request

Warning at global command creation (no change for cogs for now), and deprecated methods removed


eunwoo1104:
(Same comment as the discord server.)
The problem is that this will block using global command, even if you intended to. So because of this, sorry but closing this pull request.

PR #72, opened 2021-01-13
gettingstarted.rst

i0bs:
## Changes

- [X] I checked this pull request runs on `Python 3.6.X`.
    - Issue:
- [X] (If required) Document string is updated/added.
- [X] This changes anything except python code. (README, docs, etc.)

## About this pull request

This is a new page for the docs that allows you to learn more than the quickstart "basics" that are presented. It also adds a small update to the "quickstart.rst" and "index.rst" pages to show this page as well now.

### DIFF
--- a/docs/gettingstarted.rst
+++ b/docs/gettingstarted.rst
@@ -39,12 +39,52 @@ make a JSON example (although we won't be using it) in order to explain how this
 works:
 
 .. code-block:: python
+
   {
     "name": "argOne",
     "description": "description of first argument",
     "type": 3, # STRING type,
     "required": True
   }
+  
+With this very basic understanding in mind, now we are able to begin programming
+a simple Python script that will allow us to utilize this ability through one of
+the many subclasses offered in *discord-py-slash-command*.
+
+.. code-block:: python
+
+  import discord
+  from discord_slash import SlashCommand
+  from discord_slash.utils import manage_commands # Allows us to manage the command settings.
+
+  client = discord.Client(intents=discord.Intents.all())
+  slash = SlashCommand(client, auto_register=True)
+
+  guild_ids = [789032594456576001]
+
+  @client.event
+  async def on_ready():
+      print("Ready!")
+
+  @slash.slash(
+    name="test",
+    description="this returns the bot latency",
+    options=[manage_commands.create_option(
+      name = "argOne",
+      description = "description of first argument",
+      option_type = 3, # STRING type,
+      required = True
+    )],
+    guild_ids=guild_ids
+  )
+  async def _test(ctx, argOne: str):
+      await ctx.send(content=f"You responded with {argOne}.")
+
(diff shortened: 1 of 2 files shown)


docs/gettingstarted.rst
```diff
@@ -0,0 +1,110 @@
...
+    description="this returns the bot latency",
+    options=[manage_commands.create_option(
+      name = "argOne",
+      description = "description of first argument",
+      option_type = 3, # STRING type,
+      required = True
+    )],
+    guild_ids=guild_ids
+  )
+  async def _test(ctx, argOne: str):
```
MrSpinne, line 80: I would suggest using `arg_one` instead of `argOne` as it is pythons naming convention.

MrSpinne, line 80:
```suggestion
  async def _test(ctx, arg_one: str):
```

docs/gettingstarted.rst
```diff
@@ -0,0 +1,110 @@
...
+
+  @client.event
+  async def on_ready():
+      print("Ready!")
+
+  @slash.slash(
+    name="test",
+    description="this returns the bot latency",
+    options=[manage_commands.create_option(
+      name = "argOne",
```
MrSpinne, line 73:
```suggestion
      name = "arg_one",
```

docs/gettingstarted.rst
```diff
@@ -0,0 +1,110 @@
...
+      description = "description of first argument",
+      option_type = 3, # STRING type,
```
spinnerich:
```suggestion
      option_type = "STRING", # We can pass the name of the type directly
```
May be more conveniert.

spinnerich (commented):
Maybe we should pass the name of the type? I think it might be easier for new users to handle.
Btw great work 🎉

docs/gettingstarted.rst
```diff
@@ -0,0 +1,110 @@
...
+The main changes that you need to know about are with the lines calling the import
+of ``manage_commands``, as well as the ``options = [] ...`` code within the ``@slash.slash()``
+context coroutine. This will now create a new option called "argOne" when shown for
+the slash command.
+
+.. _quickstart: https://discord-py-slash-command.readthedocs.io/en/latest/quickstart.html
+.. _ApplicationCommandOptionChoice: https://discord.com/developers/docs/interactions/slash-commands#applicationcommandoptionchoice
+.. _ApplicationCommandOption: https://discord.com/developers/docs/interactions/slash-commands#applicationcommandoption
+
+.. toctree::
```
eunwoo1104: We don't actually need toctree and `Indices and tables` for extra docs page.
It is only necessary for the index page.

i0bs: Ah okay, I will remove it then.

eunwoo1104:
Merging again, and thank you again!

PR #75, opened 2021-01-15
Automatically call get_cog_commands for all cogs

Geekid812:
## Changes

    - Issue:
- [x] This changes anything except python code. (README, docs, etc.)

## About this pull request
This removes the necessity to call ``get_cog_commands()`` on each cog by automatically doing so in ``register_all_commands()``. It also displays a temporary warning that the call is no longer required. README and docs have been updated to reflect this.

Also, first PR hype?

### DIFF
--- a/discord_slash/client.py
+++ b/discord_slash/client.py
@@ -60,9 +60,12 @@ class SlashCommand:
                 self.get_cog_commands(cog)
             self._discord.add_cog = override_add_cog
             default_remove_function = self._discord.remove_cog
-            def override_remove_cog(cog: commands.Cog):
-                default_remove_function(cog)
+            def override_remove_cog(name: str):
+                cog = self._discord.get_cog(name)
+                if cog is None:
+                    return
                 self.remove_cog_commands(cog)
+                default_remove_function(name)
             self._discord.remove_cog = override_remove_cog
 
     def get_cog_commands(self, cog: commands.Cog):
--- a/discord_slash/client.py
+++ b/discord_slash/client.py
@@ -62,6 +62,10 @@ class SlashCommand:
         :param cog: Cog that has slash commands.
         :type cog: discord.ext.commands.Cog
         """
+        if hasattr(cog, '_slash_registered'): # Temporary warning
+            return self.logger.warning("Calling get_cog_commands is no longer required "
+            "to add cog slash commands. Make sure to remove all calls to this function.")
+        cog._slash_registered = True # Assuming all went well
         func_list = [getattr(cog, x) for x in dir(cog)]
         res = [x for x in func_list if isinstance(x, (model.CogCommandObject, model.CogSubcommandObject))]
         for x in res:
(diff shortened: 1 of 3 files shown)


i0bs:
LTGM!

Geekid812:
What's wrong?

spinnerich:
Looks good at first glance 👍

eunwoo1104:
There is one problem: it can't handle cog reload.

Geekid812:
Okay then, new approach: overriding bot.add_cog and bot.remove_cog to register slash commands. This now makes it fully reload compatible. (docs not updated yet)

discord_slash/client.py
```diff
@@ -54,14 +54,31 @@ def __init__(self,
         else:
             self._discord.add_listener(self.on_socket_response)
             self.has_listener = True
+            default_add_function = self._discord.add_cog
+            def override_add_cog(cog: commands.Cog):
+                default_add_function(cog)
+                self.get_cog_commands(cog)
+            self._discord.add_cog = override_add_cog
+            default_remove_function = self._discord.remove_cog
+            def override_remove_cog(cog: commands.Cog):
```
eunwoo1104: `remove_cog`'s parameter is `str`, not `commands.Cog`. Which means we can't remove the cog slash command with this way.

discord_slash/client.py
```diff
@@ -54,14 +54,31 @@ def __init__(self,
...
+                default_remove_function(cog)
+                self.remove_cog_commands(cog)
+            self._discord.remove_cog = override_remove_cog
 
     def get_cog_commands(self, cog: commands.Cog):
         """
         Gets slash command from :class:`discord.ext.commands.Cog`.
 
+        .. note::
+            Since version ``1.0.X``, this gets called automatically during cog initialization.
```
eunwoo1104:
```suggestion
            Since version ``1.0.9``, this gets called automatically during cog initialization.
```
We should put specific version.

discord_slash/client.py
```diff
@@ -101,9 +118,14 @@ def remove_cog_commands(self, cog):
         """
         Removes slash command from :class:`discord.ext.commands.Cog`.
 
+        .. note::
+            Since version ``1.0.X``, this gets called automatically during cog deinitialization.
```
eunwoo1104:
```suggestion
            Since version ``1.0.9``, this gets called automatically during cog deinitialization.
```
Same as the above.

eunwoo1104 (changes requested):

discord_slash/client.py
Geekid812: I left this out on purpose because I was not sure which version to fill it in with. Should've mentioned that.

discord_slash/client.py
Geekid812: Same answer as the above.

eunwoo1104 (approved):

eunwoo1104:

PR #78, opened 2021-01-23
Add cmd management to http.py

quirky-bluejay:
## About this pull request
This is needed because previously `utils.manage_commands` functions were used in `auto_register` and `auto_delete` and those functions created an aiohttp session for each function, which to quote aiohttp's docs "making a session for every request is a **very bad** idea."
This creates functions in `http.SlashCommandRequest` that use discord's session to make requests.

## Changes
Adds `add_slash_command`, `remove_slash_command`, `get_all_commands` and `command_request` to `SlashCommandRequest`
Changes `register_all_commands` and `delete_unused_commands` to use the new functions.
    - Issue:

### DIFF
--- a/discord_slash/client.py
+++ b/discord_slash/client.py
@@ -208,35 +208,39 @@ class SlashCommand:
                         options.append(base_dict)
                 if selected.allowed_guild_ids:
                     for y in selected.allowed_guild_ids:
-                        await manage_commands.add_slash_command(self._discord.user.id,
-                                                                self._discord.http.token,
-                                                                y,
-                                                                x,
-                                                                selected.description or "No Description.",
-                                                                options)
+                        await self.req.add_slash_command(
+                            self._discord.user.id,
+                            y,
+                            x,
+                            selected.description or "No Description.",
+                            options,
+                        )
                 else:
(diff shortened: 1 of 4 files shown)


eunwoo1104:
Leaving this as a external comment since it requires a lot of changes.
We already have discord client object. So we just can get bot's id as `self._discord.user.id` and remove all `bot_id` param.

quirky-bluejay:
I did notice that, but i thought that because all the other functions in `SlashCommandRequest` took `bot_id` as a parameter i assumed that i should for these functions.
I've:
- Removed `bot_id` parameter from all functions in `SlashCommandRequest`
- Updated all places where http functions were called
- Removed `bot_id` from `SlashMessage` as it was only used for http functions

eunwoo1104 (approved):

eunwoo1104:
> I did notice that, but i thought that because all the other functions in `SlashCommandRequest` took `bot_id` as a parameter i assumed that i should for these functions.

Oh, I thought I fixed that, but seems like I completely forgotten about it. Apology for that.
Anyways looks fine now. Merging this pull request, thank you for contributing!

PR #80, opened 2021-01-27
Implement PUT requests in auto_register

quirky-bluejay:
## About this pull request

This implements PUT requests for registering commands when `auto_register` and `auto_delete` are both true.
This will reduce the amount of requests made to discord, only one per guild/global compared to one request for each command for each scope.

## Changes 

- Added a `to_dict` to `SlashCommand` function to get all commands into a dictionary, this reduces duplicated code from `register_all_commands` and `sync_all_commands`
- Added `sync_all_commands` to `SlashCommand` which uses PUT requests to update the commands.
- Automatically call `sync_all_commands` when both `auto_register` and `auto_delete` are `True`
- Only call `register_all_commands` when `auto_register` is `True` and `auto_delete` isn't.
- Only call `delete_unused_commands` when `auto_register` is `True` and `auto_delete` isn't.
- Add `put_slash_commands` to http

## Checklist

    - Issue:

### DIFF
--- a/discord_slash/client.py
+++ b/discord_slash/client.py
@@ -159,13 +159,27 @@ class SlashCommand:
                         else:
                             del self.commands[x.base]
 
-    async def register_all_commands(self):
+    async def to_dict(self):
         """
-        Registers all slash commands to Discord API.\n
-        If ``auto_register`` is ``True``, then this will be automatically called.
+        Converts all commands currently registered to :class:`SlashCommand` to a dictionary.
+        Returns a dictionary in the format:
+
+        .. code-block:: python
+
+            {
+                "global" : [], # list of global commands
+                "guild" : {
+                    0000: [] # list of commands in the guild 0000
+                }
+            }
+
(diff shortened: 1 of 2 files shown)


eunwoo1104 (changes requested):
It looks like `to_dict` has `wait_until_ready`, but `register_all_commands` hasn't. This can be a problem if user just want to get a dict of the commands but can't use the coroutine. So we should also change the position of `wait_until_ready`.

quirky-bluejay:
I decided to move `wait_until_ready` to `to_dict` because i thought that someone using it would want to make sure that all the commands are loaded.
If you think that the pros of not being an asyncronous functions outweigh that, i'll change it back.

eunwoo1104:
> I decided to move `wait_until_ready` to `to_dict` because i thought that someone using it would want to make sure that all the commands are loaded.
> If you think that the pros of not being an asyncronous functions outweigh that, i'll change it back.

Ah, yeah you're right. I didn't think of that. I'll then accept this pull request, since anything else seems fine.

eunwoo1104 (approved):

eunwoo1104:
Merging this pull request, thank you always for contributing!

PR #86, opened 2021-02-03
Resolves #84 - Pass type to create_option

### LINKED ISSUE #84: Allow passing of types to discord_slash.utils.manage_commands.create_option
It would be nice to be able to pass a type, such as `str` or `discord.Role` to the `option_type` argument of the function `discord_slash.utils.manage_commands.create_option`
There may be other functions within that module which this could apply to, I haven't checked in depth. All it would need to do is map the type to an integer. For example: `str` would be `3`

quirky-bluejay:
## About this pull request

Adds a call to `SlashCommandOptionType.from_type` if the type passed was not an int

## Changes

- Restructure `SlashCommandOptionType.from_type` because currently if `bool` is passed it will return the value for an integer (this is because bool is a subclass of integer)
- Adds a call to `SlashCommandOptionType.from_type` if the type passed was not an int
- Adds an error `IncorrectType` which is raised if `from_type` doesn't return anything

## Checklist

    - Issue: #84 
- [x] (If required) Relavent documentation has been updated/added.

### DIFF
--- a/discord_slash/error.py
+++ b/discord_slash/error.py
@@ -48,3 +48,8 @@ class CheckFailure(SlashCommandError):
     """
     Command check has failed.
     """
+
+class IncorrectType(SlashCommandError):
+    """
+    Type passed was incorrect
+    """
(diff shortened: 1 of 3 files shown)


eunwoo1104 (approved):

eunwoo1104:
Well looks like the issue is resolved now. Thank you for contributing!

PR #96, opened 2021-02-12
Fixed #88

### LINKED ISSUE #88: Subcommands are registered to guilds where there are not allowed
I've noticed that when I use `guild_ids` on a subcommand for a command which is shared between multiple guilds, that subcommand is registered to all guilds where one of its subcommands is allowed.

## Steps
1. Register a subcommand `s1` for command `c`  for **Guild A**: 
   ```python
      @slash.subcommand(
         base="c",
         name="s1",
         guilds_ids=[GUILD_A_ID],
      )
      def _handle(ctx):
         # ....
     ```
1. Register a subcommand `s2` for command `c`  for **Guild B**: 
   ```python
      @slash.subcommand(
         base="c",
         name="s2",
         guilds_ids=[GUILD_A_ID],
      )
      def _handle(ctx):
         # ....
     ```

## Expected behavior
**Guild A** has `/c s1` command only and **Guild B** has `/c s2` command only.

## Actual behavior
**Guild A** has `/c s1` and `/c s2` but only can use `/c s1`, and **Guild B** has `/c s1` and `/c s2` but only can use `/c s2`.

eunwoo1104:
Restructured both API and SlashCommand command registering method.

### DIFF
--- a/discord_slash/client.py
+++ b/discord_slash/client.py
@@ -257,11 +257,11 @@ class SlashCommand:
                         if sub_sub.allowed_guild_ids:
                             for i in sub_sub.allowed_guild_ids:
                                 if i not in queue:
-                                    queue[i] = base_dict
+                                    queue[i] = copy.deepcopy(base_dict)
                                 queue[i]["options"].append(_dict)
                         else:
                             if "global" not in queue:
-                                queue["global"] = base_dict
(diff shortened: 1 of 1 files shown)


eunwoo1104:
We can just ignore codacy, just waring about unassigned list comprehension and too complex `to_dict()`.

PR #98, opened 2021-02-18
Makes temporary_auto_convert Case Insensitive 

LordOfPolls:
## About this pull request

As the title above, case insensitivity for the temporary fix. Discussed briefly on discord 

## Changes

``lower()`` was added where the arg name was being processed

## Checklist

    - Issue:

### DIFF
--- a/discord_slash/client.py
+++ b/discord_slash/client.py
@@ -649,6 +649,7 @@ class SlashCommand:
 
             # This is to temporarily fix Issue #97, that on Android device
             # does not give option type from API.
+            print(temporary_auto_convert)
             if "type" not in x:
                 x["type"] = temporary_auto_convert[x["name"]]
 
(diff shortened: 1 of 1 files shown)


eunwoo1104 (changes requested):
It looks like this only handles for subcommands. To make this also work with normal commands, changes also should be in `on_socket_response`.

LordOfPolls:
Sure thing, i only really considered subcommands, ill amend it now

quirky-bluejay:
Will this be able to pass the option of the arg to the function if the option is not all lower case?

LordOfPolls:
Isn't the dispatcher case-insensitive already? afaik the only hiccup was with the temporary fix 

at least through my, limited, testing, the args pass correctly

LordOfPolls:
image
image

eunwoo1104 (approved):

eunwoo1104:

PR #104, opened 2021-02-19
Define subcommand_name

LordOfPolls:
## About this pull request

``ctx.subcommand_name`` is defined in Context, but is never set, instead ``ctx.subcommand`` is set, but this is not present in the base class of Context. This correctly defines ``ctx.subcommand_name``

## Changes

``ctx.subcommand`` is removed, ``ctx.subcommand_name`` is set

This will break any code that is referencing ``ctx.subcommand``, which should not exist regardless 

## Checklist

    - Issue:

### DIFF
--- a/discord_slash/client.py
+++ b/discord_slash/client.py
@@ -764,7 +764,8 @@ class SlashCommand:
         sub_name = sub["name"]
         if sub_name not in base:
             return
-        ctx.subcommand = sub_name
+        if sub_name is not None:
+            ctx.subcommand_name = sub_name        
         sub_opts = sub["options"] if "options" in sub else []
         for x in sub_opts:
             if "options" in x or "value" not in x:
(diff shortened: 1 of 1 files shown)


discord_slash/client.py
```diff
@@ -764,7 +764,8 @@ async def handle_subcommand(self, ctx: context.SlashContext, data: dict):
         sub_name = sub["name"]
         if sub_name not in base:
             return
-        ctx.subcommand = sub_name
+        if sub_name is not None:
+            ctx.subcommand_name = sub_name
```
eunwoo1104: `sub_name` is always present, therefore we don't need to check if that exists.

eunwoo1104 (changes requested):

LordOfPolls:
Removed the check

eunwoo1104 (approved):

eunwoo1104:

PR #105, opened 2021-02-22
Fixes and optimisations mainly for guilds with only the applications.commands scope

MichaelJHarvey:
## About this pull request

This PR fixes some bugs introduced in v1.0.9 for guilds where the bot is not a member (i.e. bot scope) but has been added under the applications.commands scope. It also includes an optimisation that applies to these guilds along with guilds where the bot is a member but cannot view messages in the channel where a command was executed.

## Changes

Each commit in the PR includes an in-depth description of the change.
- 36ce1f6: Fix SlashMessage creation to use SlashContext.guild_id instead of SlashContext.guild as id when generating a discord generic object for the channel.
- 9cb4bb3: Fix issue in client where SlashContext.guild is used instead of SlashContext.guild_id to check whether a command is available for a guild.
- 7812a1d: Avoid waiting to set SlashContext.message in cases where the bot will always time out because it cannot read messages.

## Checklist

    - Issue:

### DIFF
--- a/discord_slash/client.py
+++ b/discord_slash/client.py
@@ -722,11 +722,8 @@ class SlashCommand:
 
             selected_cmd = self.commands[to_use["data"]["name"]]
 
-            if selected_cmd.allowed_guild_ids:
-                guild_id = ctx.guild.id if isinstance(ctx.guild, discord.Guild) else ctx.guild
-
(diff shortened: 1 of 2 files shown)


eunwoo1104 (approved):

eunwoo1104:

PR #111, opened 2021-03-04
make lib able to import with discord.ext.slash

sevenc-nanashi:
## About this pull request

Many discord.py extension can be imported like this script:
```python
from discord.ext import something
```

So I made this lib able to be imported like this.

## Changes

- Added discord/ext/slash.py

## Checklist

    - Issue:
- [x] (If required) Relavent documentation has been updated/added.

### DIFF
--- a/discord/ext/slash.py
+++ b/discord/ext/slash.py
@@ -1,0 +1,1 @@
+from discord_slash import *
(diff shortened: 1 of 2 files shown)


quirky-bluejay:
I personally have a few issues with doing this
The main one being that i think it'll make people confuse it more with `ext.commands`
And tbh i don't really see much point?

i0bs:
Don't see the need honestly, it's counterintuitive to the current design and could also impose future problems if Danny ever considers implementing slash commands into discord.py using that path. This is just an extension library for now, this library is not official.

eunwoo1104:
This doesn't look that necessary. Sorry, but closing this pull request.

PR #122, opened 2021-03-12
Update manage_commands.py

boehs:
## About this pull request

a note about how a non required field returns absolutely nothing if not filled, and how to resolve this. I tried to make it as nonspecific as possible, at the cost of clarity. I recommend if you have an idea to improve the note you do so, as it may be confusing in its current state?

## Changes

Added another note to the docstring

## Checklist

    - Issue:

### DIFF
--- a/discord_slash/utils/manage_commands.py
+++ b/discord_slash/utils/manage_commands.py
@@ -156,7 +156,7 @@ def create_option(name: str,
     :param required: Whether this option is required.
     :param choices: Choices of the option. Can be empty.
     :return: dict
-    
+
     .. note::
         a field with the value of ``False`` for ``required`` will return NOTHING if it is not filled. if you wish to have it default to something if not filled,_
         append ``argname = None`` to the ``async def`` portion of the command.
--- a/discord_slash/utils/manage_commands.py
+++ b/discord_slash/utils/manage_commands.py
@@ -156,7 +156,11 @@ def create_option(name: str,
     :param required: Whether this option is required.
     :param choices: Choices of the option. Can be empty.
     :return: dict
-
+    
+    .. note::
+        a field with the value of ``False`` for ``required`` will return NOTHING if it is not filled. if you wish to have it default to something if not filled,_
+        append ``argname = None`` to the ``async def`` portion of the command.
+        
     .. note::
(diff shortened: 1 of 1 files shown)


discord_slash/utils/manage_commands.py
```diff
@@ -157,6 +157,10 @@ def create_option(name: str,
     :param choices: Choices of the option. Can be empty.
     :return: dict
 
+    .. note::
+        a field with the value of ``False`` for ``required`` will return NOTHING if it is not filled. if you wish to have it default to something if not filled,_
+        append ``argname = None`` to the ``async def`` portion of the command.
```
quirky-bluejay:
```suggestion
    .. note::
        An option with ``required=False`` will not pass anything to the command function if the user doesn't pass that option when invoking the command.
        You must set the the relevant argument's function to a default argument, eg ``argname = None``.
```

boehs:
@AnotherCat perfect, so much better. I do wonder if its worth noting *where* the default should be set?

i0bs:
As long as you're returning a default that opts to `None`.

boehs:
> As long as you're returning a default that opts to `None`.

Not sure who you are replying to or what you mean /shrug. just hoping for some clarification!

quirky-bluejay:
> @AnotherCat perfect, so much better. I do wonder if its worth noting _where_ the default should be set?

 The one of the functions arguments, that has the same name as the arg with `required = False` needs to be a default

quirky-bluejay (approved):

boehs:
I wonder if changing 

> You must set the the relevant argument's function to a default argument, eg ``argname = None``.

to

> You must set the the relevant argument's function to a default value, eg ``argname = None``.

would help readability?

i0bs (approved):
LTGM!

eunwoo1104 (approved):

eunwoo1104:

quirky-bluejay:
> I wonder if changing 
> 
> > You must set the the relevant argument's function to a default argument, eg ``argname = None``.
> 
> to
> 
> > You must set the the relevant argument's function to a default value, eg ``argname = None``.
> 
> would help readability?

It's been merged but, i don't think that would be a good idea anyways because they are called default arguments in python docs.

PR #125, opened 2021-03-13
Adds support for cooldown and max_conc decorators

LordOfPolls:
## About this pull request
Adds support for *discord.py/ext* `cooldown` decorators and **possibly** `max_concurrency` decorators (tho i have not been able to test that one), and for once, im not making a breaking change

## Changes

- `received_at` was added to context. This stores the time the context object was created, and is used to calc cooldowns
- cooldown related functions were ported from *discord.py* into `model.py` 
- Added additional call to `invoke` function that throws a `CommandOnCooldown` exception

## Checklist

    - Issue:

### DIFF
--- a/discord_slash/context.py
+++ b/discord_slash/context.py
@@ -5,6 +5,7 @@ from datetime import datetime
 import discord
 from contextlib import suppress
 from discord.ext import commands
+from discord.utils import snowflake_time
 from . import http
 from . import error
 from . import model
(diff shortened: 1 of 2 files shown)


discord_slash/context.py
```diff
@@ -57,6 +59,7 @@ def __init__(self,
             self.author = discord.User(data=_json["member"]["user"], state=self.bot._connection)
         else:
             self.author = discord.User(data=_json["user"], state=self.bot._connection)
+        self.received_at = datetime.now()  # the time this context was created (used for cooldowns)
```
phenom4n4n: I think for consistency with dpy, this should be renamed to `self.created_at`. Additionally, rather than relying on the system time which could be inaccurate, and may not be in UTC time, a better option would be to use `discord.utils.snowflake_time` as is [used across the discord.py lib](https://github.com/Rapptz/discord.py/blob/a8f44174bafed3989ec2959a62b89006f4a9e9a1/discord/role.py#L244).

LordOfPolls: Agreed, committing

LordOfPolls:
Closing due to incoming pr that would do this better

PR #139, opened 2021-03-30
Implement files in editing messages

quirky-bluejay:
## About this pull request

This implements files while editing messages, and sending files in the initial response

## Changes

What changes were made?

Changes to `SlashMessage.edit` and edit in http to allow editing for files
Changed the behavior of `SlashContext.send` to defer and then patch if files are included for the initial response.

## Checklist

    - Issue:

### DIFF
--- a/discord_slash/context.py
+++ b/discord_slash/context.py
@@ -171,15 +171,15 @@ class SlashContext:
         initial_message = False
         if not self.responded:
             initial_message = True
-            if files:
-                raise error.IncorrectFormat("You cannot send files in the initial response!")
+            if files and not self.deffered:
+                await self.defer(hidden=hidden)
             if self.deffered:
                 if self._deffered_hidden != hidden:
                     self._logger.warning(
                         "Deffered response might not be what you set it to! (hidden / visible) "
                         "This is because it was deffered in a different state"
                     )
-                resp = await self._http.edit(base, self.__token)
+                resp = await self._http.edit(base, self.__token, files = files)
                 self.deffered = False
             else:
                 json_data = {
--- a/discord_slash/context.py
+++ b/discord_slash/context.py
@@ -193,10 +193,8 @@ class SlashContext:
         initial_message = False
         if not self.responded:
             initial_message = True
-            if files and not self.defered:
+            if files and not self.deferred:
                 await self.defer(hidden=hidden)
-            if self.defered:
-                if self._defered_hidden != hidden:
             if self.deferred:
                 if self._deferred_hidden != hidden:
                     self._logger.warning(
@@ -204,8 +202,6 @@ class SlashContext:
                         "This is because it was deferred in a different state"
                     )
                 resp = await self._http.edit(base, self.__token, files = files)
-                self.defered = False
-                resp = await self._http.edit(base, self.__token)
                 self.deferred = False
             else:
                 json_data = {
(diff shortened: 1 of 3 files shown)


discord_slash/context.py
```diff
@@ -193,14 +193,18 @@ async def send(self,
...
         if not self.responded:
             initial_message = True
-            if files:
-                raise error.IncorrectFormat("You cannot send files in the initial response!")
+            if files and not self.defered:
+                await self.defer(hidden=hidden)
+            if self.defered:
+                if self._defered_hidden != hidden:
             if self.deferred:
                 if self._deferred_hidden != hidden:
```
eunwoo1104, line 199: That is a small mistake I guess?

discord_slash/context.py
```diff
@@ -193,14 +193,18 @@ async def send(self,
...
                     self._logger.warning(
                         "deferred response might not be what you set it to! (hidden / visible) "
                         "This is because it was deferred in a different state"
                     )
+                resp = await self._http.edit(base, self.__token, files = files)
+                self.defered = False
                 resp = await self._http.edit(base, self.__token)
                 self.deferred = False
```
eunwoo1104, line 205: Another duplicate code.

discord_slash/context.py
```diff
@@ -216,6 +220,9 @@ async def send(self,
             self.responded = True
         else:
             resp = await self._http.post_followup(base, self.__token, files=files)
+        if files:
+            for file in files:
+                file.close()
```
eunwoo1104, line 221: Why would we need to close file objects? If I'm remembering correct it should be handled in http part.

discord_slash/model.py
```diff
@@ -194,39 +194,59 @@ def __init__(self, *, state, channel, data, _http: http.SlashCommandRequest, int
...
-            _resp["allowed_mentions"] = allowed_mentions.to_dict() if allowed_mentions else \
-                self._state.allowed_mentions.to_dict() if self._state.allowed_mentions else {}
-
-            await self._http.edit(_resp, self.__interaction_token, self.id)
-
-            delete_after = fields.get("delete_after")
-            if delete_after:
-                await self.delete(delay=delete_after)
+        if "file" in fields or "files" in fields:
+            await self._slash_edit(**fields)
```
eunwoo1104, line 244: Everything should be wrapped to try except block since this is for trying discord.py's original function, then if we get 403, which means only slash commands are added, it will use slash command webhook.

eunwoo1104 (changes requested):

discord_slash/model.py
quirky-bluejay: d.py doesn't support files while editing messages yet, so i'm just skipping it because it'll error every time

discord_slash/context.py
quirky-bluejay: d.py implements it in `send()` https://github.com/Rapptz/discord.py/blob/757dd92272763d2b481685e50c3b3784cb3594b5/discord/abc.py#L1060

discord_slash/context.py
quirky-bluejay: Seems like I introduced it while merging in new changes :frowning:

eunwoo1104 (approved):

LordOfPolls (approved):

eunwoo1104:
Merging this pull request!

PR #142, opened 2021-03-31
Removes getdoc  from command registration

LordOfPolls:
## About this pull request

If the description param isn't specified, the library will fall back to using doc strings as the command description. If the doc string is over the char limit for descriptions, it will cause syncing to fail. 

Expected behaviour of the library would be to use the description param, or "No description".

Technically breaking as someone could be using this behaviour
## Changes

Removed getdoc

## Checklist

    - Issue:

### DIFF
--- a/discord_slash/client.py
+++ b/discord_slash/client.py
@@ -413,7 +413,6 @@ class SlashCommand:
         subcommand_group = subcommand_group.lower() if subcommand_group else subcommand_group
         name = name or cmd.__name__
         name = name.lower()
-        description = description
         guild_ids = guild_ids if guild_ids else []
 
         if base in self.commands:
--- a/discord_slash/client.py
+++ b/discord_slash/client.py
@@ -2,7 +2,7 @@ import copy
 import logging
 import typing
 import discord
-from inspect import iscoroutinefunction, getdoc
+from inspect import iscoroutinefunction
 from contextlib import suppress
 from discord.ext import commands
 from . import http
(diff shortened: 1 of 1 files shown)


i0bs (approved):
In support of this.

LordOfPolls:
Doc-strings are intended as comments, people treat them as developer side code commenting. Due to this, someone new to the library would neglect to specify a description param to the decorator, causing an error thats confusing to debug.

Theres no easy way for us to be able to determine if a doc-string is being used to comment code, or be a description for a command, as such i feel its better to just ignore them and use the description param.

artem30801:
I use docstrings as it takes a lot less space and more intuitive (and also behaviour in discord.py). I think many people would want to use them too. Can you add param (like use_docstrings=True) to SlashCommand init call to enable docstring usage?

LordOfPolls:
Converting this PR into a draft as more people than i assumed use doc-strings for descriptions, and id rather minimise breaking changes. Feel free to put your thoughts into the `lib-discussions` channel on the Discord

circuitsacul:
I use docstrings for all command descriptions (in discord.py), I never use the `description` parameter because that looks really messy in code.

spinnerich:
Note: Discord.py also uses docstrings for command descriptions.

LordOfPolls:
We know. Its also been said multiple times in this thread

i0bs:
The goal of a library extension should be to extend off of the original discord.py branch, and by that logic we're superseding a normal programming habit by using docstrings not for _documentation_ but for command descriptions. This comes with various drawbacks due to the design choices that were made at that time for this. 

My suggestion is we offer a compromise for this as this does have the ability to already break our existent library code, and it's being written off as "well it's already a thing so keep it."

PR #155, opened 2021-04-19
Fixing wrong behaviour when using guild_ids

mgiacopu:
## About this pull request

When using `guild_ids` in the `@slash.slash` decorator the handling of incoming interactions halts prematurely [on line 772 in the client.py file](https://github.com/eunwoo1104/discord-py-slash-command/blob/d5cba045a60aa5f9e3e16d16e7a4520fbdd31e3c/discord_slash/client.py#L772) and it prevents the bot from interacting with users.
This is due to `ctx.guild_id` being an _integer_ and `selected_cmd.allowed_guild_ids` a _list of strings_: obviously an integer will never be in a string list.

Casting `ctx.guild_id`  to a _string_ solves the problem.

## Changes

Casted `ctx.guild_id`  to a _string_

## Checklist

    - Issue:

### DIFF
--- a/discord_slash/client.py
+++ b/discord_slash/client.py
@@ -730,7 +730,7 @@ class SlashCommand:
 
             selected_cmd = self.commands[to_use["data"]["name"]]
 
-            if selected_cmd.allowed_guild_ids and ctx.guild_id not in selected_cmd.allowed_guild_ids:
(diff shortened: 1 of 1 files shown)


quirky-bluejay:
`allowed_guild_ids` should be a list of integers.

mgiacopu:
Whooops.
I was using environment variables (which are strings) as guild ids.
Thank you, it just flew over my mind.

PR #156, opened 2021-04-21
Add Application Permissions Support

benwoo1110:
## About this pull request

Adds support for the slash command permissions. I wanted it for my own bot, got it into a workable state, so I figure I can open a PR to add its feature for everyone if accepted. Feel free to review and test. Let me know of any feedback, happy to accept critics and improve the code 😄.

## Changes

* Add setting of `default_permission` and `permissions` option for slash command.
* Add `get_all_command_permissions` and `put_command_permissions` method to http module.
* Add support for storing permission data in `CommandObject`.
* Add `create_guild_permissions` and `create_permission` utils method to facility defining permissions.
* Add `PermissionsData` class used for checking for any permission updates since last sync, similar to how `CommandData` works.
* Add logic to registers permissions in the `sync_all_commands` method.
* Add `SlashCommandPermissionsType` enum class to match discord api's [ApplicationCommandPermissionType](https://discord.com/developers/docs/interactions/slash-commands#applicationcommandpermissiontype).
* Add relevant documentation needed for the code changes.

## Example decorator with subcommand:
```py
@slash.subcommand(
    guild_ids=[750556940127436880, 823454213089787914],
    base="test", 
    base_desc="test",  
    subcommand_group="group",
    sub_group_desc="group",
    name="sub",
    description="sub"
)
@slash.permission(823454213089787914, generate_permissions(allowed_roles=[823851472982376468, 823851440232857600]))
@slash.permission(750556940127436880, generate_permissions(allowed_roles=[778838673666998292]))
async def test_sub(ctx):
    await ctx.send("test_sub")
```

## Checklist

    - Issue:


discord_slash/utils/manage_commands.py
```diff
@@ -235,3 +235,23 @@ def create_choice(value: str, name: str):
...
+
+
+def create_guild_permissions(applicable_guilds: typing.List[int], permissions: typing.List[dict]):
+    return {
+        "applicable_guilds": applicable_guilds,
+        "permissions": permissions
+    }
+
+
+def create_permission(id: typing.Union[str, int], id_type: int, permission: bool):
```
quirky-bluejay: snowflakes should always be integers

discord_slash/utils/manage_commands.py
```diff
@@ -235,3 +235,23 @@ def create_choice(value: str, name: str):
         "value": value,
         "name": name
     }
+
```
quirky-bluejay, line 316: I think another function should be created to do:
```py
funcname(
    allowed_roles: List[int],
    allowed_users: List[int],
    disallowed_roles: List[int],
    disallowed_users: List[int]
) 
``` 
This would make it more user friendly, reducing the amount of repetition required.

discord_slash/http.py
```diff
@@ -54,6 +54,24 @@ def get_all_commands(self, guild_id=None):
         """
         return self.command_request(method="GET", guild_id=guild_id)
 
+    def get_all_command_permissions(self, guild_id):
```
quirky-bluejay:
```suggestion
    def get_all_guild_commands_permissions(self, guild_id):
```

discord_slash/http.py
```diff
@@ -54,6 +54,24 @@ def get_all_commands(self, guild_id=None):
...
+        """
+        Sends a slash command get request to Discord API for all permissions of a guild.
+
+        :param guild_id: ID of the target guild to get registered command permissions of.
+        :return: JSON Response of the request.
+        """
+        return self.command_request(method="GET", guild_id=guild_id, url_ending="/permissions")
+
+    def put_command_permissions(self, guild_id, perms_dict):
```
quirky-bluejay:
```suggestion
    def update_guild_commands_permissions(self, guild_id, perms_dict):
```

quirky-bluejay (changes requested):
This is a great start to implementing this! 
I've outlined a couple of things that I noticed, but there are a few things with the overall implementation that I would like to see.

`permissions` to be structured by guild id on the command object
ie 
```py
{
  guildid: [ list of guild permission objects],
  guild2id: [list of guild permissions obects],
}
``` 
This is because I don't think the current list of dicts that contain a list of guild ids represents how permissions work on the api, it's set per guild, not in a group of guilds and can differ between guilds.
The current approach could also lead to some setting one role id for a list of objects, which best case will fail silently, worst case discord errors because the role doesn't exist on that guild.

To reflect this I think that:
- The `permissions` parameter of `@slash`(etc) changed to accept a dict of guild ids to permissons
- Also create a decorator that can be repeated for each guild ie `@permission(guild_id = 1, permissions [...])`

 Also functions should be added to `utils.manage_commands` that make requests to the api for the three permission related endpoints, as this is the case for command creation

discord_slash/client.py
```diff
@@ -132,6 +132,7 @@ def get_cog_commands(self, cog: commands.Cog):
                         "auto_convert": {},
                         "guild_ids": x.allowed_guild_ids.copy(),
                         "api_options": [],
+                        "api_permissions": None,
```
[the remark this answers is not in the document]
benwoo1110: Technically no difference as later in the code there checking for that `cmd["api_permissions"] or []` when creating a CommandObject based on the dict, but yea, will change to `[]` for consistency.

discord_slash/utils/manage_commands.py
benwoo1110: What should we name this function? 😄 But yea good idea, I will implement it soon.

Edit: maybe `generate_permissions`?

Edit again: implemented it!

discord_slash/client.py
```diff
@@ -220,8 +221,13 @@ async def to_dict(self):
...
             command_dict = {
                 "name": x,
-                "description": selected.description or "No Description.",
-                "options": selected.options or []
+                "cmd": {
+                    "name": x,
+                    "description": selected.description or "No Description.",
+                    "options": selected.options or [],
+                    "default_permission": selected.default_permission
+                },
```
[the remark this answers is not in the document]
benwoo1110: I was thinking of something like this:
```py
{
  "global": # global cmds here
  "guild": {
    guildid1: {
      cmdname1: {
        "cmd": # commands for that guild here
        "perms": # permissions for that guild here
      }
    }
  }
}
```

bcu the actual command object sent to discord doesn't have permissions in them, so it should be separate. Is there another way you want to go about doing this?

discord_slash/client.py
```diff
@@ -220,8 +221,13 @@ async def to_dict(self):
...
+                "perms": selected.permissions or []
             }
             if selected.allowed_guild_ids:
```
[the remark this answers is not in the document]
benwoo1110: Yea I get what you mean, my only query is how about global commands? Bcu it's possible for a global command to have permission in multiple guilds, but not for guild specific commands.

discord_slash/client.py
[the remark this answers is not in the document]
quirky-bluejay: Flatten the dict so it's 
```py
{
   "name": name,
   "description":description,
   "options":options,
   "default_permission":default_permission,
   "permissions": permissions
}
```
I don't see any reason to nest the command data separately, you are correct that the api does not return permission data when requesting commands, but that doesn't mean that we should separate them.  
`to_dict` is a documented function and therefore changing the structure of what it's returning is a breaking change.

discord_slash/client.py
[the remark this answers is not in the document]
quirky-bluejay: I'm not 100% sure but I think this has been resolved by changing the `permissions` to a dict structure.

discord_slash/client.py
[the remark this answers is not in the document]
benwoo1110: https://github.com/eunwoo1104/discord-py-slash-command/blob/d5cba045a60aa5f9e3e16d16e7a4520fbdd31e3c/discord_slash/client.py#L322-L336

It was due to the fact that that dict itself is supposed to be the exact thing that is sent to discord API, so I decided to split it. But its definitely possible to just do something like this in the code for loop code snippet shown above to map it to the dict required for discord API:
```
{
   "name": command.name,
   "description": command.description,
   "options": command.options,
   "default_permission": command.default_permission
}
```
Since we should not be including the permissions on the request for putting slash commands to discord API.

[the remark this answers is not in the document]
quirky-bluejay: imo flatten it and remove the permissions key in `sync_all_commands`. `to_dict` is meant to provide a representation of the commands in a dictionary format, not necessarily commands which are exactly in the format discord wants.
Also the api doesn't care if you include extra fields, just tested that. (still should remove `permissions` before sending the request solely to reduce data sent.)
I just think that we should avoid breaking changes :shrug:

benwoo1110:
Hi, I have implemented all the suggestions and changes. Hope you can take a look again soon!

(except documentation 😅 )

discord_slash/http.py
```diff
@@ -167,4 +185,3 @@ def delete(self, token, message_id="@original"):
         """
         req_url = f"/messages/{message_id}"
         return self.command_response(token, True, "DELETE", url_ending = req_url)
-
```
quirky-bluejay, line 170: Why was this line removed?

discord_slash/model.py
```diff
@@ -358,3 +360,37 @@ async def wrap():
...
+class PermissionsData:
+    """
+    Slash permissions data object
+
+    :ivar id: Command id, provided by discord.
+    :ivar permissions: List of permissions dict.
+    """
+    def __init__(self, id, permissions, **kwargs):
+        self.id = id
+        self.permissions = permissions
```
quirky-bluejay: I think this should create a list of another class, see how `CommandData` does options.

discord_slash/utils/manage_commands.py
```diff
@@ -142,6 +142,84 @@ async def remove_all_commands_in(bot_id,
...
+async def get_all_guild_commands_permissions(bot_id,
+                                             bot_token,
+                                             guild_id):
+    """
+    A coroutine that sends a slash command get request to Discord API.
+
+    :param bot_id: User ID of the bot.
+    :param bot_token: Token of the bot.
+    :param guild_id: ID of the guild to get permissions.
+    :return: JSON Response of the request.
```
quirky-bluejay: Maybe link to the api docs for what it returns?

discord_slash/utils/manage_commands.py
```diff
@@ -235,3 +313,38 @@ def create_choice(value: str, name: str):
         "value": value,
         "name": name
     }
+
+
+def create_permission(id: typing.Union[int], id_type: int, permission: bool):
```
quirky-bluejay:
```suggestion
def create_permission(id:int, id_type: int, permission: typing.Union[bool, SlashCommandPermissionsType]):
```

discord_slash/utils/manage_commands.py
```diff
@@ -235,3 +313,38 @@ def create_choice(value: str, name: str):
...
+    if not isinstance(id_type, int) or isinstance(id_type, bool): #Bool values are a subclass of int
+        original_type = id_type
+        id_type = SlashCommandPermissionsType.from_type(original_type)
+        if id_type is None:
+            raise IncorrectType(f"The type {original_type} is not recognized as a type that Discord accepts for slash command permissions.")
+    return {
+        "id": str(id),
```
quirky-bluejay:
```suggestion
        "id": id,
```
Snowfakes are integers

discord_slash/client.py
```diff
@@ -470,6 +551,8 @@ def add_subcommand(self,
             "description": base_description,
             "guild_ids": guild_ids.copy(),
             "api_options": [],
+            "default_permission": base_default_permission,
+            "api_permissions": base_permissions,
```
AnotherCat, line 555: Subcommands don't have permissions, permissions are only on the 'base' command

quirky-bluejay (changes requested):
Left a few comments, and `add_subcommand` should ideally attempt to 'merge' new permissions from other subcommands instead of just ignoring them.

discord_slash/http.py
benwoo1110: I think my code editor auto-format the file such that there is one empty line at the end of the file. Looks like the file had 2 empty lines at the end that's why one line is removed. I can add back if you want?

discord_slash/client.py
```diff
@@ -560,6 +564,8 @@ def add_subcommand(self,
             "sub_group_desc": subcommand_group_description,
             "guild_ids": guild_ids,
             "api_options": options,
+            "default_permission": base_default_permission,
+            "api_permissions": base_permissions,
```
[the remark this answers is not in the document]
benwoo1110: The reason I needed is as `SubcommandObject` inherits `CommandObject` which means it will throw an error if those 2 fields are not present. .But thinking of it, I can just pass None value.

discord_slash/http.py
quirky-bluejay: Oh oops sorry, for some reason i thought this removed the newline at the end of the file, which is required by pep8 ~~not that we follow it all the time anyways~~

discord_slash/client.py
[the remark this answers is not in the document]
quirky-bluejay: Maybe it shouldn't inherit from CommandObject. :thinking: 
cc @eunwoo1104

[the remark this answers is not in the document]
benwoo1110: 93b8865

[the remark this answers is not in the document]
quirky-bluejay: Honestly don't like this. It implies that the permissions do exist for subcommands but are always empty, which isn't the case. 
How about instead creating a base command class which then the normal command class inherits, and subcommands inherent?

discord_slash/http.py
benwoo1110: Ah I see, no worries all good!

discord_slash/client.py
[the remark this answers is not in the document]
benwoo1110: Sounds like a better option, just that it will have breaking change since SubcommandObject now inherits a different class.

discord_slash/model.py
benwoo1110: 7815254

discord_slash/client.py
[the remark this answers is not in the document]
benwoo1110: 7e16da9

I made `CommandObject` the common class which `BasecommandObject` and `SubcommandObject` inherits from.

docs/gettingstarted.rst
```diff
@@ -229,7 +229,59 @@ a string or integer. Below is an implementation of this design in the Python cod
...
+In this case, we are setting 2 permissions for guild with id of ``12345678``. Firstly, we are allowing
+role with id ``99999999`` and disallowing user with id ``88888888`` from running the slash command.
+
+.. code-block:: python
+
+  @slash.slash(name="test",
+              description="This is just a test command, nothing more.",
+              permissions={
+                12345678: [
+                  create_permission(99999999, 1, True),
```
i0bs: There is no mention of importing this. Is this meant to refer to the required line needed below?

`from discord_slash.utils.manage_commands import create_permission`

docs/gettingstarted.rst
```diff
@@ -229,7 +229,59 @@ a string or integer. Below is an implementation of this design in the Python cod
...
+following table shows the supported id types for permissions:
+
++-------------------+-----------+
+| **Name**          | **Value** |
++-------------------+-----------+
+| Role              | 1         |
++-------------------+-----------+
+| User              | 2         |
++-------------------+-----------+
+
```
i0bs, line 261: JSON response object should additionally be shown here to exemplify the table. Some developers of this library end up using kwarg "plug-and-play" for coding slash commands (as I know I do) so they may use the JSON route as well.

i0bs (changes requested):
Looking for just a couple more changes to the general outlook of the .RST documentation before going into further details.

docs/gettingstarted.rst
```diff
@@ -229,7 +229,59 @@ a string or integer. Below is an implementation of this design in the Python cod
...
+| **Field**   | **Type**                                   | **Description**                                                                                     |
++-------------+--------------------------------------------+-----------------------------------------------------------------------------------------------------+
+| id          | int                                        | Snowflake value of type specified. Represents the target to apply permissions on.                   |
++-------------+--------------------------------------------+-----------------------------------------------------------------------------------------------------+
+| type        | int                                        | An `ApplicationCommandPermissionType`_.                                                             |
++-------------+--------------------------------------------+-----------------------------------------------------------------------------------------------------+
+| permission  | boolean                                    | true to allow, false, to disallow.                                                                  |
```
i0bs: `true` to allow, `false` to disallow

This is per the Discord v8 docs to keep consistency.

i0bs (changes requested):
Last changes needed before .RST is of a satisfactory state.

docs/gettingstarted.rst
LilSpazJoekp: Just been following this..does the table need that much empty space?

i0bs: Yes due to RST formatting, that's how they have to be structured.

docs/gettingstarted.rst
```diff
@@ -229,7 +229,59 @@ a string or integer. Below is an implementation of this design in the Python cod
...
++-------------+--------------------------------------------+-----------------------------------------------------------------------------------------------------+
```
LilSpazJoekp:
```suggestion
+------------+----------+----------------------------------------------+
| **Field**  | **Type** | **Description**                              |
+------------+----------+----------------------------------------------+
| id         | int      | Snowflake value of type specified. Represent |
|            |          | the target to apply permissions on.          |
+------------+----------+----------------------------------------------+
| type       | int      | An `ApplicationCommandPermissionType`_.      |
+------------+----------+----------------------------------------------+
| permission | boolean  | true to allow, false, to disallow.           |
+------------+----------+----------------------------------------------+
```
This would be better optimized.

docs/gettingstarted.rst
LilSpazJoekp: Not necessarily see below suggestions.

docs/gettingstarted.rst
```diff
@@ -229,7 +229,59 @@ a string or integer. Below is an implementation of this design in the Python cod
...
+set permissions to apply for User or Role, `ApplicationCommandPermissionType`_ is used. It's a number and
+following table shows the supported id types for permissions:
+
++-------------------+-----------+
+| **Name**          | **Value** |
++-------------------+-----------+
+| Role              | 1         |
++-------------------+-----------+
+| User              | 2         |
++-------------------+-----------+
```
LilSpazJoekp, line 260:
```suggestion
+----------+-----------+
| **Name** | **Value** |
+----------+-----------+
| Role     | 1         |
+----------+-----------+
| User     | 2         |
+----------+-----------+
```

docs/gettingstarted.rst
i0bs: I don't think this is really needed, the official .RST file for this has them purposefully big with space for copy-paste reasons. (Slash Commands are still WIP albeit not being officially in beta anymore)

docs/gettingstarted.rst
i0bs: I don't think this is really needed, the official .RST file for this has them purposefully big with space for copy-paste reasons. (Slash Commands are still WIP albeit not being officially in beta anymore)

mike1808:
Are you planning to add permissions to `cog_subcommand`?

benwoo1110:
Must have forgotten to add the permissions parameter for cogs, will push to add it later

Edit: 
I have added support for it. btw if anyone wants to test, you can do 
`pip install --force git+https://github.com/benwoo1110/discord-py-slash-command@permissions`

mike1808:
@benwoo1110 I set `base_default_permission=False` to `@cog_ext.cog_subcommand` and it failed with: 

```
  File "/bot/slash.py", line 114, in _setup_wiki_commands
    await self.slash.sync_all_commands()
  File "//bot-env/lib/python3.9/site-packages/discord_slash/client.py", line 310, in sync_all_commands
    cmds = await self.to_dict()
  File "/bot-env/lib/python3.9/site-packages/discord_slash/client.py", line 220, in to_dict
    "default_permission": selected.default_permission,
AttributeError: 'BasecommandObject' object has no attribute 'default_permission'
```

mike1808:
I also wondering, how can I programmatically add more permissions to commands or subcommand? Imagine I set `default_permission=False` to one command and then I want to add permissions from another slash command. I was thinking to add permissions to `slash.commands[COMMAND].permissions` and then force resync. But that looks like hack then a real solution.

benwoo1110:
> @benwoo1110 I set `base_default_permission=False` to `@cog_ext.cog_subcommand` and it failed with:
> 
> ```
> Traceback (most recent call last):
>   File "/bot/slash.py", line 114, in _setup_wiki_commands
>     await self.slash.sync_all_commands()
>   File "//bot-env/lib/python3.9/site-packages/discord_slash/client.py", line 310, in sync_all_commands
>     cmds = await self.to_dict()
>   File "/bot-env/lib/python3.9/site-packages/discord_slash/client.py", line 220, in to_dict
>     "default_permission": selected.default_permission,
> AttributeError: 'BasecommandObject' object has no attribute 'default_permission'
> ```

Just fixed it, was a typo of `default_permissions` instead of `default_permission` for some attributes.



> I also wondering, how can I programmatically add more permissions to commands or subcommand? Imagine I set `default_permission=False` to one command and then I want to add permissions from another slash command. I was thinking to add permissions to `slash.commands[COMMAND].permissions` and then force resync. But that looks like hack then a real solution.

Yea it's a good idea, tho it probably can be done after this PR, since currently all command attributes can't be updated programmatically without syncing commands as well.

discord_slash/model.py
```diff
@@ -176,6 +175,28 @@ async def can_run(self, ctx) -> bool:
         return False not in res
 
 
+class BasecommandObject(CommandObject):
```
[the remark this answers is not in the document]
benwoo1110: Previously it was`SubcommandObject` for sub commands, so naturally I named it `BasecommandObject` for base command. But if the devs wants me to change it, then sure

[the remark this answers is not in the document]
quirky-bluejay: "Subcommand" is considered one word, that's what the API docs use. "BaseCommand" is two words.

discord_slash/model.py
```diff
@@ -185,16 +185,16 @@ class BasecommandObject(CommandObject):
     .. warning::
         Do not manually init this model.
 
-    :ivar has_subcommand: Indicates whether this base command has subcommands.
-    :ivar default_permissions: Indicates whether users should have permissions to run this command by default.
+    :ivar has_subcommands: Indicates whether this base command has subcommands.
```
quirky-bluejay: Is this another breaking change?

discord_slash/utils/manage_commands.py
```diff
@@ -316,6 +316,14 @@ def create_choice(value: str, name: str):
 
 
 def create_permission(id:int, id_type: int, permission: typing.Union[bool, SlashCommandPermissionsType]):
+    """
+    Create a single command permission.
+
+    :param id: Target id to apply the permission on.
+    :param id_type: Type of the id.
```
quirky-bluejay: This should have a link to the type enum.

docs/gettingstarted.rst
```diff
@@ -229,7 +229,81 @@ a string or integer. Below is an implementation of this design in the Python cod
   async def test(ctx, optone: str):
     await ctx.send(content=f"Wow, you actually chose {optone}? :(")
 
+Want to restrict access? Setup permissions!
+-------------------------------------------
+
+Slash commands also supports ability to set permissions to allow only certain roles and/or users 
+to run a slash command. Permissions can be applied to both global and guild based commands. They 
+are defined per guild, and each guild can have multiple permissions. Here table that shows the JSON 
+structure of how choices are represented for a single permission:
```
quirky-bluejay:
```suggestion
Slash commands also support the ability to set permissions to allow only certain roles and/or users 
to run a slash command. Permissions can be applied to both global and guild based commands. They 
are defined per guild, and each guild can have multiple permissions. Here is a table that shows the JSON 
structure of how choices are represented for a single permission:
```

docs/gettingstarted.rst
```diff
@@ -229,7 +229,81 @@ a string or integer. Below is an implementation of this design in the Python cod
   async def test(ctx, optone: str):
     await ctx.send(content=f"Wow, you actually chose {optone}? :(")
 
+Want to restrict access? Setup permissions!
+-------------------------------------------
+
+Slash commands also supports ability to set permissions to allow only certain roles and/or users 
+to run a slash command. Permissions can be applied to both global and guild based commands. They 
+are defined per guild, and each guild can have multiple permissions. Here table that shows the JSON
```
quirky-bluejay:
```suggestion
are defined per guild ,per top-level command (the base command for subcommands), and each guild can have multiple permissions. Here table that shows the JSON 
```
Not sure how to work this but I think it's important to point out.

docs/gettingstarted.rst
```diff
@@ -229,7 +229,81 @@ a string or integer. Below is an implementation of this design in the Python cod
...
+
+Additionally, we could also declare the id type of our permission through this method shown here:
+
+.. code-block:: python
+
+  from discord_slash.model import SubcommandApplicationPermissionType
+  
+  (...)
+  
+                  option_type=SubcommandApplicationPermissionType.ROLE
```
quirky-bluejay: How about just using the enum in the example, I think using the enums in files is always a good idea as it prevents confusion.

quirky-bluejay (changes requested):

docs/gettingstarted.rst
quirky-bluejay: Is choices the right word here?

quirky-bluejay:
> I also wondering, how can I programmatically add more permissions to commands or subcommand? Imagine I set `default_permission=False` to one command and then I want to add permissions from another slash command. I was thinking to add permissions to `slash.commands[COMMAND].permissions` and then force resync. But that looks like hack then a real solution.

The only reason I can see to do this is so that if people sync manually it'll update.
Permissions are only used to be set on the API, kind of like the description. 
Keep in mind that if the value is updated while the bot is running, it'll 'reset' whenever the bot process is reloaded. 
The only use case I can see for this is allowing users to add roles/user permissions without a restart, and to do that you shouldn't really be syncing the commands, as it's a lot of requests, instead just making a single request to update the permissions.

discord_slash/model.py
benwoo1110: It was `has_subcommands` in the current code, so I think the error without an `s` was typo caused by one of by commits.
https://github.com/eunwoo1104/discord-py-slash-command/blob/d5cba045a60aa5f9e3e16d16e7a4520fbdd31e3c/discord_slash/model.py#L126

docs/gettingstarted.rst
benwoo1110: sure

quirky-bluejay:
Using @everyone for permissions will be supported! https://github.com/discord/discord-api-docs/issues/2798#issuecomment-831672498
Thoughts on adding a user friendly way? Or do we just make a note about using the guild id

benwoo1110:
I dont think there is a need for any code change, since setting permissions for it is exactly the same as with any other roles. I think at most a short note in the documentation will do.

i0bs (approved):
LGTM, waiting for response from others.

ghost (approved):

discord_slash/client.py
```diff
@@ -300,14 +304,17 @@ async def sync_all_commands(self, delete_from_unused_guilds=False):
...
+        :param delete_perms_from_unused_guilds: If the bot should make a request to clear permissions for guilds that haven't got any permissions registered in :class:``SlashCommand``
         """
+        permissions_map = {}
         cmds = await self.to_dict()
         self.logger.info("Syncing commands...")
         cmds_formatted = {None: cmds['global']}
         for guild in cmds['guild']:
             cmds_formatted[guild] = cmds['guild'][guild]
-                
+
```
i0bs, line 325: Is this an intentional formatting change that is required?

discord_slash/utils/manage_commands.py
```diff
@@ -235,3 +313,62 @@ def create_choice(value: str, name: str):
...
     }
+
+
+def create_permission(id:int, id_type: int, permission: typing.Union[bool, SlashCommandPermissionsType]):
+    """
+    Create a single command permission.
+
+    :param id: Target id to apply the permission on.
+    :param id_type: Type of the id, :class:`..model.SlashCommandPermissionsType`.
+    :param permission: State of the permission. ``True`` to allow access, ``False to disallow access.
```
quirky-bluejay: Should this not be 
```suggestion
    :param permission: State of the permission. ``True`` to allow access, ``False`` to disallow access.
```

discord_slash/model.py
```diff
@@ -358,3 +379,64 @@ async def wrap():
...
+        else:
+            return False
+
+
+class GuildPermissionsData:
+    """
+    Slash permissions data for a guild.
+
+    :ivar id: Command id, provided by discord.
+    :ivar permissions: List of permissions dict.
```
quirky-bluejay: https://discord.com/developers/docs/interactions/slash-commands#guildapplicationcommandpermissions

```suggestion
    Slash permissions data for a command in a guild.

    :ivar id: Command id, provided by discord.
    :ivar guild_id: Guild id that the permissions are in. 
    :ivar permissions: List of permissions dict.
```

discord_slash/model.py
```diff
@@ -358,3 +379,64 @@ async def wrap():
...
+                and self.permissions == other.permissions
+            )
+        else:
+            return False
+
+
+class SlashCommandPermissionsType(IntEnum):
+    """
+    Equivalent of `ApplicationCommandPermissionType <https://discord.com/developers/docs/interactions/slash-commands#applicationcommandpermissiontype>`_  in the Discord API.
+    """
```
quirky-bluejay, line 438: Possibly add a note about `@everyone` permissions?

quirky-bluejay (changes requested):
What happens if you pass duplicate ids to the manage_commands functions?

discord_slash/client.py
benwoo1110: I think my ide removes trailing spaces, and also I believe this aligns with the check style you all use.

discord_slash/model.py
benwoo1110: yea sure

benwoo1110:
> What happens if you pass duplicate ids to the manage_commands functions?

which function you are referring to? For `generate_permissions` and `create_multi_ids_permission`, I can just cast the ids list to set to remove duplicates.

quirky-bluejay:
> > What happens if you pass duplicate ids to the manage_commands functions?
> 
> which function you are referring to? For `generate_permissions` and `create_multi_ids_permission`, I can just cast the ids list to set to remove duplicates.

Sorry, I meant if id's are both allowed and disabled for the same id
I think this would be `generate_permissions`, but also what would happen if a dict was passed with this?

benwoo1110:
I dont think there is a need to check for that in code, the false (i.e. disallow) permission just will take precedence. 

Also from my testing, this is the behaviour of permission. Let say we have 2 roles called `allowed` and `disallowed`.
* user doesn't have any role: command is fade out and no tab-complete. They actually still can run the command on discord desktop if they type out in full. `default_permission` needs to be set to False for that command to prevent this. (yea a bit weird by discord here)
* user has disallowed role: command is fade out and no tab-complete. The interaction will fail if they try to run on the discord desktop.
* user has allowed role: able to tab-complete and run the command.
* user has both allowed and disallowed role: able to tab-complete and run the command.

quirky-bluejay:
> [....]
> * user has both allowed and disallowed role: able to tab-complete and run the command.

I meant if they are the **same** id. Just wondering what will happen then :upside_down_face:

benwoo1110:
> the false (i.e. disallow) permission just will take precedence.

discord_slash/model.py
quirky-bluejay: @benwoo1110 I unresolved this before, see above ^^

benwoo1110: oh, I thought it was just a docs typo

benwoo1110: 89d38c0

eunwoo1104 (approved):
LGTM, there seems to be no more problems.

discord_slash/utils/manage_commands.py
```diff
@@ -235,3 +313,65 @@ def create_choice(value: str, name: str):
...
+    }
+
+
+def create_multi_ids_permission(ids: typing.List[int], id_type: int, permission: bool):
+    """
+    Creates a list of permissions from list of ids with common id_type and permission state.
+
+    :param id: List of target ids to apply the permission on.
+    :param id_type: Type of the id. 
+    :param permission: State of the permission. ``True`` to allow access, ``False to disallow access.
```
quirky-bluejay:
```suggestion
    :param permission: State of the permission. ``True`` to allow access, ``False`` to disallow access.
```
Sphinx was complaining :(
````
C:\Users\___\Documents\Coding\test py-slash\discord-py-slash-command\discord_slash\utils\manage_commands.py:docstring of discord_slash.utils.manage_commands.create_multi_ids_permission:5: WARNING: Inline literal start-string without end-string.
```

benwoo1110: what didn't I commit your suggested change previously lol

quirky-bluejay (changes requested):
The `@slash.permission` decorator doesn't seem to work.
I tested it (code [here](https://gist.github.com/AnotherCat/87c8f0bf2254c380f083a5aa68160d47)) and `map_permissions` in `sync_all_commands` logged as `{}`
Using it with the `permissions` parameter for `@slash.slash` seems to be working fine.

discord_slash/client.py
```diff
@@ -338,10 +346,55 @@ async def sync_all_commands(self, delete_from_unused_guilds=False):
...
+                        permissions_map[applicable_guild] = []
+                    permission = {
+                        "id": cmd_id,
+                        "guild_id": applicable_guild,
+                        "permissions": cmd_permissions[applicable_guild]
+                    }
+                    permissions_map[applicable_guild].append(permission)
+
+
+        self.logger.debug(permissions_map)
```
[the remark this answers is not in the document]
quirky-bluejay: Note that I'm only talking about the logs from `discord_slash`, not the ones from `discord.http`

benwoo1110:
> The `@slash.permission` decorator doesn't seem to work.
> I tested it (code [here](https://gist.github.com/AnotherCat/87c8f0bf2254c380f083a5aa68160d47)) and `map_permissions` in `sync_all_commands` logged as `{}`
> Using it with the `permissions` parameter for `@slash.slash` seems to be working fine.

Decorator order actually matters, the permission decorator must come after, If I am not wrong it's the same case for d.py checks decorators.

LordOfPolls (approved):

discord_slash/client.py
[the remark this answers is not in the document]
LordOfPolls: Can this be resolved so we can merge this?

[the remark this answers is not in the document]
benwoo1110: oh its was hidden GitHub website, too many commits and comments HAHA, should be resolved already.

eunwoo1104 (approved):
LGTM.

eunwoo1104:

PR #162, opened 2021-05-01
General cleanup and format code with black

LilSpazJoekp:
## About this pull request

This cleans up a few things in the codebase.

## Changes

- Moved `__version__` and `BASE` constants to `discord_slash/const.py` for easier access when they need changed
- Removed non-project patterns that should be placed in a [global ignore file](https://help.github.com/en/github/using-git/ignoring-files#create-a-global-gitignore) from `.gitignore`
- Removed unused imports
- setup.py
  - Optimized imports
  - Moved requirements into `setup.py` file since there are now extra requirements only needed when linting or building docs
  - Version info is now parsed from `const.py`
  - Modified how `README.md` and `const.py` are accessed to prevent issues in rare instances
  - Added more pypi classifiers
  - Added license parameter to `setup` call
- docs
  - conf.py
    - Added version and release info
    - Removed necessary exclude patterns
- Added `pre_push.py` that will format and lint code automatically for contributors
- Added GitHub Actions for enforcing code style
- Added step for running `pre_push.py` in pull request template
- Fixed two tiny errors that were uncovered from linting:
  - This needs to be a raw docstring: https://github.com/eunwoo1104/discord-py-slash-command/blob/d5cba045a60aa5f9e3e16d16e7a4520fbdd31e3c/discord_slash/http.py#L115-L125
  - This can be rewritten to use `isinstance` as suggested by flake8: https://github.com/eunwoo1104/discord-py-slash-command/blob/d5cba045a60aa5f9e3e16d16e7a4520fbdd31e3c/discord_slash/utils/manage_commands.py#L217
- Sorted imports with [isort](https://pycqa.github.io/isort/)
- Formats code with the tool [black](https://black.readthedocs.io/en/stable/)

## Checklist

    - Issue:


.gitignore
```diff
@@ -1,10 +1,8 @@
-.idea
```
benwoo1110, line 1: Is there a reason for `.idea` to be removed?

LilSpazJoekp: `.idea` is not a project specific folder. This pattern belongs in your global ignore file.

LordOfPolls: `.idea` should be in `gitignore`, its a product of most jetbrains IDEs, and has no place on git. Suggest resolving this review

LilSpazJoekp: > has no place on git. 

Correct. Which is why it belongs in the [global gitignore file](https://docs.github.com/en/github/getting-started-with-github/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer).

LordOfPolls: Apologies I was a bit too brief there (just got home). 

I agree with it being in global, however for the sake of making sure *everybody* who contributes to this not accidentally committing `.idea` it makes sense to have it in a project specific gitignore as well. There is no issue with duplicate entries in project/global afaik (please do correct if im wrong there)

Imo its best to keep the barrier of entry low, and not everyone knows about global config for git (let alone how to use `git config` 😆), hence why i think this should just be resolved rather than accepted

quirky-bluejay: I think this should be removed, it belongs in the global gitignore file, and if people don't know what that is and accidently commit that all prs should be checked anyways

quirky-bluejay:
@LilSpazJoekp new changes need to be merged 🙂 
Also what's the reasoning for ci linting on three os's?

LilSpazJoekp:
Will do. Since pre_push.py can be ran on those three OSs, it would catch any issues people would have running that.

.gitignore
LilSpazJoekp: I strongly feel that `.idea` should be left out as it's not hard to setup a global gitignore. Also as @AnotherCat has said, it would be caught in code review.

LilSpazJoekp:
This is ready now.

setup.py
```diff
@@ -1,20 +1,52 @@
...
+        "Development Status :: 4 - Beta",
+        "Intended Audience :: Developers",
+        "License :: OSI Approved :: MIT License",
+        "Natural Language :: English",
+        "Operating System :: OS Independent",
+        "Programming Language :: Python :: 3",
+        "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.7",
+        "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
```
quirky-bluejay, line 46:
```suggestion
```
d.py currently doesn't support 3.9

LilSpazJoekp: Python 3.x is compatible with Python 3.9.

quirky-bluejay: There's a reason they don't support it, i can't exactly remember but i think it was something to do with it not building well on 3.9? Not sure if it was resolved

LilSpazJoekp: I see, that doesn't make much sense though. By design, each version of Python 3.x must be compatible with each prior version preceding it.

quirky-bluejay: This reminds me of https://xkcd.com/1172/ :smile:
It seems like it was because dependencies didn't use to support 3.9, they've said d.py works on 3.9 now but don't explicitly include it in their setup.py
Not sure what to do honestly

LordOfPolls: Wasn't this issue with 3.9.x resolved a while ago? it was a dep issue that afaik is now fixed. Also, subjectively, all of my bots are running on 3.9.x without issue

LilSpazJoekp:
You will have to ignore CI errors because the doc strings in components isn't valid reStructuredText.

LordOfPolls:
[Merge #162](https://trello.com/c/j7tlB4Yb/14-merge-162)

PR #178, opened 2021-05-17
Fix kwargs being passed as dict in cog

BobDotCom:
## About this pull request

There is still an issue with commands outside of cogs, but this fixes for cogs.

## Changes

Updated the invoke function inside CogBaseCommandObject to be identical to its corresponding function in CogSubcommandObject.

## Checklist

    - Issue: Not an issue, rather fixes errors that I found (presumably created in https://github.com/eunwoo1104/discord-py-slash-command/commit/1629ff9c30d7621e2f9f91842c69539b7bb1a8ce)

### DIFF
--- a/discord_slash/model.py
+++ b/discord_slash/model.py
@@ -261,11 +261,31 @@ class CogBaseCommandObject(BaseCommandObject):
         :param args: Args for the command.
         :raises: .error.CheckFailure
         """
-        can_run = await self.can_run(args[0])
+        args = list(args)
+        ctx = args.pop(0)
+        can_run = await self.can_run(ctx)
         if not can_run:
             raise error.CheckFailure
 
-        return await self.func(self.cog, *args, **kwargs)
+        coro = None  # Get rid of annoying IDE complainings.
+
+        not_kwargs = False
+        if args and isinstance(args[0], dict):
+            kwargs = args[0]
+            ctx.kwargs = kwargs
+            ctx.args = list(kwargs.values())
+            try:
+                coro = self.func(self.cog, ctx, **kwargs)
+            except TypeError:
+                args = list(kwargs.values())
+                not_kwargs = True
+        else:
+            ctx.args = args
+            not_kwargs = True
+        if not_kwargs:
+            coro = self.func(self.cog, ctx, *args)
+
+        return await coro
 
 
 class CogSubcommandObject(SubcommandObject):


LordOfPolls:
Can you give an example of the behaviour this is trying to fix? 

Because, possibly because I've just woken up, I can't see why this is necessary, nor have I run into the issue you describe

BobDotCom:
> Can you give an example of the behaviour this is trying to fix?
> 
> Because, possibly because I've just woken up, I can't see why this is necessary, nor have I run into the issue you describe

When i added slash commands in my cogs, it was passing an empty dict into the command function if there were no options passed in the command. Also, if there were options passed it would pass them all inside one dict instead of expanding out into kwargs, which this fixes.

benwoo1110:
A code example of what you described would help.

BobDotCom:
> A code example of what you described would help.

Screen Shot 2021-05-17 at 8 51 12 PM
I also added options=[] which prevented the mystery_arg from being provided as an option on the command.

Kigstn:
Can confirm, functions without any arguments do not work anymore after [the latest commit](https://github.com/eunwoo1104/discord-py-slash-command/commit/1629ff9c30d7621e2f9f91842c69539b7bb1a8ce)

**Example:**
This used to work but doesn't anymore:
```python
@cog_ext.cog_slash(
    name="test",
    description="test",
)
async def _test(self, ctx: SlashContext):
    ctx.send("test")
```

This is how it needs to look after the latest commit:
```python
@cog_ext.cog_slash(
    name="test",
    description="test",
)
async def _test(self, ctx: SlashContext, dummy):
    ctx.send("test")
```

benwoo1110:
I don't really understand why dummy parameter is needed. You don't have any slash command options to pass any data to the dummy argument.

Not sure what issue with slash command this will fix as well.

LordOfPolls:
Tested, can confirm this is an issue

BobDotCom:
> I don't really understand why dummy parameter is needed. You don't have any slash command options to pass any data to the dummy argument.
> 
> Not sure what issue with slash command this will fix as well.

thats why its confusing, but i've tested and the code that i pushed fixes the said issue.

BobDotCom:
There still appears to be a bug for slash commands outside of cogs (`slash.slash` instead of `cog_ext.cog_slash`), so if someone wants to open an issue or pr and fix that as well, that needs to be done at some point.

LordOfPolls:
Yeah ive done some digging and your fix doesnt actually resolve the issue 

The actual issue is in `client.py` in `invoke_command`. So while yes this fixes *part* of the issue, it doesnt actually resolve it

LordOfPolls:
Check #lib-discussion in the discord

BobDotCom:
Closing in favor of #181

PR #181, opened 2021-05-17
Fix typeError properly, and resolve pr #178

LordOfPolls:
## About this pull request

Actually fixes the `typeError` bug, without losing functionality, and resolves the bug that was attempted to be fixed in #178 

## Changes

Reworked `invoke_command` to check the target function to see if it expects kwargs or args based on if the arguments match the name of the options

## Checklist

    - Issue:

### DIFF
--- a/discord_slash/client.py
+++ b/discord_slash/client.py
@@ -869,7 +869,22 @@ class SlashCommand:
         :param args: Args. Can be list or dict.
         """
         try:
-            await func.invoke(ctx, args)
+            # determine if args should be passed as kwargs or not
+            kwargs = False
+
+            if isinstance(args, dict):
+                for _arg_name in args.keys():
+                    if not hasattr(func, _arg_name):
+                        kwargs = False
+                        break
+                else:
+                    kwargs = True
+
+            if kwargs:
+                await func.invoke(ctx, **args)
+            else:
+                await func.invoke(ctx, *args)
+
         except Exception as ex:
             await self.on_slash_command_error(ctx, ex)
 


discord_slash/client.py
```diff
@@ -869,7 +869,22 @@ async def invoke_command(self, func, ctx, args):
...
         """
         try:
-            await func.invoke(ctx, args)
+            # determine if args should be passed as kwargs or not
+            kwargs = False
+
+            if isinstance(args, dict):
+                for _arg_name in args.keys():
+                    if not hasattr(func, _arg_name):
+                        kwargs = False
```
quirky-bluejay, line 878: isn't this line not needed?
```suggestion
```

discord_slash/client.py
```diff
@@ -869,7 +869,22 @@ async def invoke_command(self, func, ctx, args):
...
+                        break
+                else:
+                    kwargs = True
```
benwoo1110, line 881: Maybe a better solution is to assume `kwargs` is true first:
```py
kwargs = True

if isinstance(args, dict):
    for _arg_name in args.keys():
        if not hasattr(func, _arg_name):
            kwargs = False
            break
else:
    kwargs = False
```

quirky-bluejay: That would end up with `kwargs` always being `False`

benwoo1110: No? Why will it be the case?

quirky-bluejay: Don't mind me not noticing a change in indentation lol

i0bs: Either solution seems fine tbh.

PR #187, opened 2021-05-21
Update docs/quickstart.rst

Toricane:
Updated quickstart.rst to use commands.Bot instead of discord.Client.

## About this pull request

Updated docs/quickstart.rst to use commands.Bot instead of discord.Client.

## Changes

Some changes were made to the example code blocks in the quickstart page of the docs to use commands.Bot instead of discord.Client to be consistent with the migration page.

## Checklist

    - Issue:

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -28,7 +28,7 @@ For this example, ``main.py`` will be used.
     from discord.ext import commands
     from discord_slash import SlashCommand # Importing the newly installed library.
 
-    bot = commands.Bot(command_prefix="/", intents=discord.Intents.all())
+    bot = commands.Bot(intents=discord.Intents.all())
     slash = SlashCommand(bot, sync_commands=True) # Declares slash commands through the client.
 
     @bot.event
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -25,16 +25,17 @@ For this example, ``main.py`` will be used.
 .. code-block:: python
 
     import discord
+    from discord.ext import commands
     from discord_slash import SlashCommand # Importing the newly installed library.
 
-    client = discord.Client(intents=discord.Intents.all())
-    slash = SlashCommand(client, sync_commands=True) # Declares slash commands through the client.
+    bot = commands.Bot(command_prefix="/", intents=discord.Intents.all())
+    slash = SlashCommand(bot, sync_commands=True) # Declares slash commands through the client.
 
-    @client.event
+    @bot.event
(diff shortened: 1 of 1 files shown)


i0bs:
Is this really needed? The thing about using `commands.Bot()` is that it has a kwarg for `command_prefix` but because we are relying on interactions within the Discord API, this prefix is never inferred upon because of Slash Contextualization. The reason it was given `discord.Client()` instead was solely because of this.

**EDIT**: It's also worth noting that the `commands` path directory is explicitly made so that text-based commands would be processed. We're not dealing with more so text anymore but a UX that Discord provides. This would be really nothing more than a choice of the programmer and not an actual requirement whatsoever.

i0bs (commented):
Leaving this up to more feedback.

docs/quickstart.rst
```diff
@@ -25,16 +25,17 @@ For this example, ``main.py`` will be used.
 .. code-block:: python
 
     import discord
+    from discord.ext import commands
     from discord_slash import SlashCommand # Importing the newly installed library.
 
-    client = discord.Client(intents=discord.Intents.all())
-    slash = SlashCommand(client, sync_commands=True) # Declares slash commands through the client.
+    bot = commands.Bot(intents=discord.Intents.all())
```
benwoo1110: Very sure `command_prefix` is needed for commands.Bot to work.

benwoo1110 (commented):
I think the example in the readme should be the "most basic setup" needed for slash command to work. So `commands.Bot` don't look like something necessary to me.

Toricane:
> I think the example in the readme should be the "most basic setup" needed for slash command to work. So `commands.Bot` don't look like something necessary to me.

Okay then, we can keep discord.Client.

Toricane:
Do I close the PR?

benwoo1110:
That's just my opinion tho, the lead dev can think otherwise 🤷

i0bs:
If you're sure on keeping it, then you're able to close it yourself.

Toricane:
I have an idea; I will make another example showing how to use discord-py-slash-command with discord.py, would that be better?

Toricane:
Made another commit

quirky-bluejay:
I'm also thinking about making changes to how the events are listened for, which would remove the issues/annoyances when using `discord.Client`.
I agree that there is not much point with using `commands.Bot` if there are no text-based commands or cogs.

i0bs:
I am closing this because frankly I see no point or need in having this as a change for the Quick Start.

PR #197, opened 2021-05-31
Fix issue where default_permission doesn't get updated on sync.

### LINKED ISSUE #195: default_permission doesn't have any effect
Hi,

Thanks a lot for this project!

I am trying to restrict the use of a command to a specific role using the following options, though it does not seem to have any effect:
```
    "default_permission": False,
    "permissions": {
        GUILD_ID : [ create_permission(ROLE_ID, SlashCommandPermissionType.ROLE, True) ]
    },
```
Other options I've tried that don't work:
- Only setting one "positive" permission to the role
- Setting one "negative" permission to the role `@everyone` and one positive to one role (the negative permission has no effect)
- Setting default_permission=False without anything else also has no effect.

The only way I can disallow a role/user to use it is by explicit adding a negative permission for them, meaning that for admin commands I would need to add a negative permission to all the guild users (because some users don't have roles).

I believe that's not the intended behavior and is a bug right?

benwoo1110:
## About this pull request

Fix issue that CommandData is equal even though default_permission has be changed. 

## Changes

Added default_permission parameter to CommandData.

## Checklist

    - Issue: fixes #195

### DIFF
--- a/discord_slash/model.py
+++ b/discord_slash/model.py
@@ -70,16 +70,18 @@ class CommandData:
 
     :ivar name: Name of the command.
     :ivar description: Description of the command.
+    :ivar default_permission: Indicates whether users should have permissions to run this command by default.
     :ivar options: List of :class:`OptionData`.
     :ivar id: Command id, this is received from discord so may not be present
     :ivar application_id: The application id of the bot, required only when the application id and bot id are different. (old bots)
     """
 
     def __init__(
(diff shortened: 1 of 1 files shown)


quirky-bluejay (approved):
LGTM
I suggest this be a priority and released asap

eunwoo1104 (approved):

eunwoo1104:
Merging!

Issue #129, opened 2021-03-25
slash decorators should handle users using str for snowflakes


LordOfPolls:
Seeing a lot of users using `str` for their snowflakes. Specifically for `guild_ids`. The issue with this is that it doesn't cause any errors for the user, causing confusion, but obviously breaks their commands and is very hard to debug 

I think we should be either converting the snowflake to `int` on the library side, or throwing a warning on creation of the command

ghost:
Would be better to just do `int()`

ghost:
> Would be better to just do `int()`

Btw why is there still no PR for that? Or I just missed it was merged already

LordOfPolls:
Because I haven't made one :/

If you read this issue I'm giving 2 options. I made this issue to give people the opportunity to choose, as I don't really care either way

ghost:
> Because I haven't made one :/

This *repo* needs a new hero xD

quirky-bluejay:
What does d.py do if you pass a snowflake as a string?

gabrielcastelan:
Nothing.

boehs:
> Seeing a lot of users using `str` for their snowflakes. Specifically for `guild_ids`. The issue with this is that it doesn't cause any errors for the user, causing confusion, but obviously breaks their commands and is very hard to debug
> 
> I think we should be either converting the snowflake to `int` on the library side, or throwing a warning on creation of the command

I would be for option b, throwing a error, snowflakes are a int, and we should stick to best programming practices

benwoo1110:
Doing nothing is probably the easiest option since it's already documented that snowflakes are int and `'1' != 1` is the first few things I learnt in python as well. But definitely, an informative error could be useful for people new to discord bot API 😄.

ghost:
eh, didnt it now do warning in 2.0?

PR #209, opened 2021-06-12
String Snowflake ID in decorator args

FayeDel:
## About this pull request

This PR essentially warns the user that the command(s) that uses string snowflake IDs as a guild_ids arg does not work and therefore result in "This interaction failed".

## Changes

Added a `guilds_id` check if all of them are ints on both the regular slash client end and on the `cog_ext` end, and provided an Exception upon check failure.

## Checklist

- [X] This fixes something in [Issues](https://github.com/eunwoo1104/discord-py-slash-command/issues).
    - Issue: #129


i0bs (approved):

quirky-bluejay (changes requested):
This should raise a custom exception, for example like the others that are in `.error`.
Also not too sure about this one, but shouldn't all development be targeted at the `components` branch?
I wouldn't count this a breaking change as it doesn't 'work' when ids are strings anyways.

FayeDel:
To be completely honest, I wasn't too sure if it would be targeted towards component because I presumed that it was for components and its own spiel. That and, I had to second guess marking if it was a breaking change, but the tip helps in console so I marked it there.

LordOfPolls:
I'm happy with these changes, but can you please make them for the components branch, as per what cat said

LordOfPolls:
Is this complete?

FayeDel:
This is switched to components with its own respective changes, yes. Should have mentioned that earlier, my bad

LordOfPolls:
Please may you run `pre_push.py`, rn your pr is failing due to formatting issues

FayeDel:
Ran the PR and formatted (double checked even without the file and it seems to be fine).

LordOfPolls (approved):

LordOfPolls:
Merging, thanks for contributing, please dm me on discord for the role

Toricane (commented):
test

PR #212, opened 2021-06-17
add component tutorial

LordOfPolls:
## About this pull request

Add a guide on how to use components 

## Changes

Add components.rtf
Add components.rtf to index

## Checklist

    - Issue:

### DIFF
--- a/docs/components.rst
+++ b/docs/components.rst
@@ -1,0 +1,75 @@
+How to use components
+=====================
+
+
+So you've seen the new fancy buttons and want to give them a try, but dont know where to start. No problem!
+
+First lets cover the basics. Discord messages can have *components*, such as buttons, dropdowns etc. These components sit in whats called an "action row". An action row can hold 5 components at a time. Your message can have 5 action rows, so thats a total of 25 components!
+
+Sending some components
+_______________________
+
+First we need to create some buttons, lets put them in a list for now. We'll use :meth:`create_button() <discord_slash.utils.manage_components>` to create a green button
+
+.. code-block:: python
+
+    from discord_slash.utils import manage_components
+    from discord_slash.model import ButtonStyle
+
+    buttons = [
+                manage_components.create_button(
+                    style=ButtonStyle.green,
+                    label="A Green Button"
+                )
+              ]
+
+So we have a button, but where do we use it. Let's create an action row with :meth:`create_actionrow() <discord_slash.utils.manage_components>` and put our buttons in it
+
+.. code-block:: python
+
+    [...]
+    action_row = manage_components.create_actionrow(*buttons)
+
+
+Fantastic, we now have an action row with a green button in it, now lets get it sent in discord
+
+.. code-block:: python
+
+    await ctx.send("My Message", components=[action_row])
+
+Now if you've followed along, you have a green button in discord! But theres a problem, whenever you click it you see that the ``interaction failed``. Why is that?
+Well, in Discord, clicking buttons and using slash commands are called ``interactions``, and Discord doesn't know if we've received them or not unless we tell Discord. So how do we do that?
+
+Responding to interactions
+__________________________
+
+When responding to interactions, we have 2 choices, either to do it in the command itself, or a global event handler (like :meth:`on_slash_command_error`)
+
+Lets go through the most common method first, responding in the event itself. We simply need to :meth:`wait_for` the event, just like you do for reactions. For this we're going to use :meth:`wait_for_component() <discord_slash.utils.manage_components>`, and we're going to only wait for events from the action row we just .
+This method will return a :class:`ComponentContext <discord_slash.context.ComponentContext>` object that we can use to respond. For this example, we'll just edit the original message (:meth:`edit_origin() <discord_slash.context.ComponentContext.edit_origin>`)
+
+.. code-block:: python
+
+    await ctx.send("My Message", components=[action_row])
+    button_ctx: ComponentContext = await manage_components.wait_for_component(bot, action_row)
+    await button_ctx.edit_origin(content="You pressed a button!")
+
+.. note:: It's worth being aware that if you handle the event in the command itself, it will not persist reboots. As such when you restart the bot, the interaction will fail
+
+Next we'll go over the alternative, a global event handler. This works just the same as :meth:`on_slash_command_error` or `on_ready`.
+
+.. code-block:: python
+
+    @bot.event
+    async def on_component(ctx: ComponentContext):
+        await button_ctx.edit_origin(content="You pressed a button!")
+
+But [writer], I dont want to edit the message
+*********************************************
+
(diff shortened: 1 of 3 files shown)


i0bs (changes requested):
Make a reference link to this on the Getting Started page.

LordOfPolls:
Why? It's visible on all pages in the sidebar?

i0bs:
People are going to refer to it first if they're new, more of people not actually looking. It's sadly happened before

LordOfPolls:
Requesting review from @artem30801 and @hpenney2 as well

docs/components.rst
```diff
@@ -0,0 +1,76 @@
...
+    from discord_slash.model import ButtonStyle
+
+    buttons = [
+                manage_components.create_button(
+                    style=ButtonStyle.green,
+                    label="A Green Button"
+                )
+              ]
+
+So we have a button, but where do we use it. Let's create an action row with :meth:`create_actionrow() <discord_slash.utils.manage_components>` and put our buttons in it
```
artem30801: Shouldn't it be :func: instead of :meth:?
```suggestion
So we have a button, but where do we use it. Let's create an action row with :func:`create_actionrow() <discord_slash.utils.manage_components>` and put our buttons in it
```

docs/components.rst
```diff
@@ -0,0 +1,76 @@
...
+.. code-block:: python
+
+    from discord_slash.utils import manage_components
+    from discord_slash.model import ButtonStyle
+
+    buttons = [
+                manage_components.create_button(
+                    style=ButtonStyle.green,
+                    label="A Green Button"
+                )
```
artem30801: I suggest inserting `,` for easier copy-pasting
```suggestion
                ),
```

docs/components.rst
```diff
@@ -0,0 +1,76 @@
...
+
+    [...]
+    action_row = manage_components.create_actionrow(*buttons)
+
+
+Fantastic, we now have an action row with a green button in it, now lets get it sent in discord
+
+.. code-block:: python
+
+    await ctx.send("My Message", components=[action_row])
```
artem30801, line 38: I think it should be noted that sending buttons work both for slash (interction) context and for any d.py sending/editing/replying methods
Also, I would expand this snippet to be a very simple d.py or slash command so it's easier to immidiatly jump in and start playing with it

docs/components.rst
```diff
@@ -0,0 +1,76 @@
...
+
+Now if you've followed along, you have a green button in discord! But theres a problem, whenever you click it you see that the ``interaction failed``. Why is that?
+Well, in Discord, clicking buttons and using slash commands are called ``interactions``, and Discord doesn't know if we've received them or not unless we tell Discord. So how do we do that?
+
+Responding to interactions
+__________________________
+
+When responding to interactions, we have 2 choices, either to do it in the command itself, or a global event handler (like :meth:`on_slash_command_error`)
```
artem30801: I'm not sure about that wording, as it's choices of listening/waiting for interaction

docs/components.rst
```diff
@@ -0,0 +1,76 @@
...
+
+Lets go through the most common method first, responding in the event itself. We simply need to :meth:`wait_for` the event, just like you do for reactions. For this we're going to use :meth:`wait_for_component() <discord_slash.utils.manage_components>`, and we're going to only wait for events from the action row we just .
```
artem30801:
```suggestion
Lets go through the most common method first, responding in the event itself. We simply need to :meth:`wait_for` the event, just like you do for reactions. For this we're going to use :func:`wait_for_component() <discord_slash.utils.manage_components>`, and we're going to only wait for events from the action row we just .
```

docs/components.rst
```diff
@@ -0,0 +1,76 @@
...
+.. code-block:: python
+
+    @bot.event
+    async def on_component(ctx: ComponentContext):
+        await button_ctx.edit_origin(content="You pressed a button!")
+
+But [writer], I dont want to edit the message
+*********************************************
+
+Well lucky for you, you don't have to. You can either respond silently, with a thinking animation, or send a whole new message. Take a look here: :class:`ComponentContext <discord_slash.context.ComponentContext>`
```
artem30801, line 73: I think that you still have to respond either way. Defer will fail in 15 minutes, it's a bad practice to leave deferred responses hanging, IMO

docs/components.rst
```diff
@@ -0,0 +1,76 @@
...
+
+How do i know which button was pressed?
+_______________________________________
+
+Each button gets a ``custom_id``, this is a unique identifier of which button is being pressed. You can specify what the ID is when you define your button. When handling the event, simply check the custom_id, and handle accordingly.
```
artem30801: Worth mentioning that `create_button` generates random `custom_id` if it wasn't specified, and mention `get_components_ids` funtion to extract `custom_ids` from actionrow or list of actionrows

artem30801 (changes requested):

docs/components.rst
LordOfPolls: i intentionally didnt mention if it was slash context or dpy context so the reader would realise that both could be used. But yeah ill make it explicit

as for expanding this into a command, im opposed to that. expanding it into a command would mean the user would just copy and paste the entire function. It would also mean the snippet is no longer just showing what the above text is trying to teach

docs/components.rst
LordOfPolls: How about:

When responding, you have 2 choices in how you handle them. You can either wait for them in the command itself, or listen for them in a global event handler (like :meth:`on_slash_command_error`)

docs/components.rst
LordOfPolls: My understanding is that a type 6 response is considered acceptable, being that it ACKs the interaction, and gives the option to edit later. I dont see why not editing/sending would be bad practice, if anything it unnecessarily limits what you can do with them. If discord was opposed to hanging a type 6, surely an error would be shown, or a note in their docs?

Obviously, using a `thinking` animation defer would be undesirable however

docs/components.rst
LordOfPolls: I was trying to keep this as simple as possible, but i can add a note to explain that if you dont specify a `custom_id` a random one is given. I think mentioning `get_component_ids` is unnecessary in the scope of this guide however

docs/components.rst
artem30801: When responding, you have 2 choices in how you handle interactions. You can either wait for them in the command itself, or listen for them in a global event handler (similar to :meth:on_slash_command_error)

LordOfPolls: works for me, resolving

docs/components.rst
LordOfPolls: Ive added the note about `custom_id` random gen, not resolving due to the unresolved `get_components_ids` part

docs/components.rst
LordOfPolls: Added note about context, not resolving due to the unresolved discussion about expanding into a command

i0bs (approved):

PR #226, opened 2021-06-23
Fix error when guild id is None.

### LINKED ISSUE #225: Traceback from library on basic function
Traceback on Version 2.0.1. On 1.1.0 working as expected.
python 3.8.5

### Traceback
```
  File "bot.py", line 28, in <module>
    async def purge(ctx, number_of_messages : int):
  File "/home/solumath/.local/lib/python3.8/site-packages/discord_slash/client.py", line 760, in wrapper
    obj = self.add_slash_command(
  File "/home/solumath/.local/lib/python3.8/site-packages/discord_slash/client.py", line 538, in add_slash_command
    if not all(isinstance(item, int) for item in guild_ids):
TypeError: 'NoneType' object is not iterable
```

### Code
```py
@slash.slash(name="purge", description="delete number of messages")
@commands.has_permissions(manage_messages=True)
async def purge(ctx, number_of_messages : int):
    await ctx.channel.purge(limit=number_of_messages)
    await ctx.send("What is real? How do you define real?", delete_after=5)
```
Probably in library as discussed on discord server.

benwoo1110:
## About this pull request

I would say this is quite critical. Not sure why that checking was removed at this commit:
 https://github.com/discord-py-slash-commands/discord-py-interactions/commit/b32d0e2da956bbc31066d7bdbe4240c97258eca3#diff-37afa4fe0663ae6e7e30366179bcffecedbdba72972e291cb682fec2ac4f0a24L538-L539

## Changes

Add checking for when guild_id is None for command.

## Checklist

    - Issue: fixes #225

### DIFF
--- a/discord_slash/client.py
+++ b/discord_slash/client.py
@@ -535,7 +535,8 @@ class SlashCommand:
         """
         name = name or cmd.__name__
         name = name.lower()
-        if guild_ids and not all(isinstance(item, int) for item in guild_ids):
+        guild_ids = guild_ids if guild_ids else []
+        if not all(isinstance(item, int) for item in guild_ids):
             raise error.IncorrectGuildIDType(
                 f"The snowflake IDs {guild_ids} given are not a list of integers. Because of discord.py convention, please use integer IDs instead. Furthermore, the command '{name}' will be deactivated and broken until fixed."
             )
@@ -616,7 +617,8 @@ class SlashCommand:
         name = name or cmd.__name__
         name = name.lower()
         description = description or getdoc(cmd)
-        if guild_ids and not all(isinstance(item, int) for item in guild_ids):
+        guild_ids = guild_ids if guild_ids else []
+        if not all(isinstance(item, int) for item in guild_ids):
             raise error.IncorrectGuildIDType(
                 f"The snowflake IDs {guild_ids} given are not a list of integers. Because of discord.py convention, please use integer IDs instead. Furthermore, the command '{name}' will be deactivated and broken until fixed."
             )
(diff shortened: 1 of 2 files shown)


LilSpazJoekp:
That was my bad. I thought I added `if guild_ids` to both of those.

LilSpazJoekp (approved):
LGTM

I missed the other two when I changed that.

artem30801:
It seems to me that with such checks empty list of guild_ids would be a valid input? Is that actually desired? I'm not sure but it could cause some other bugs down the line

benwoo1110:
https://github.com/discord-py-slash-commands/discord-py-interactions/blob/a7f468b66137726a2535a9d62177f6e25a8f8d2f/discord_slash/model.py#L315

It will always be modelled to an empty list from what I can tell.

artem30801:
Then LGTM

benwoo1110:
Actually @artem30801 you are right. There are instances where it tried to .copy a None guild_ids, I think it best is to revert back to what was like before, at least for this hotfix. 

Then we see if we should consider later on whether removing that `guild_ids = guild_ids if guild_ids else []` has any benefit or not. Seen evidently more code change is needed to support it.

artem30801:
I think that
`guild_ids = guild_ids if guild_ids else []`
also could be changed to to
`guild_ids = guild_ids or []`

benwoo1110:
I think both have negligible difference, let keep with what proven works for now. We can do some optimisation on that after this hotfix.

LordOfPolls (approved):

PR #230, opened 2021-06-24
Make edit functions more like discord.py edit functions

### LINKED ISSUE #228: Cannot remove embeds nor components from a message
[`ComponentContext.edit_origin`](https://github.com/discord-py-slash-commands/discord-py-interactions/blob/83937fefc297c3062c018381ef5c7751225b5d9b/discord_slash/context.py#L369) and `SlashMessage.edit` ([`SlashMessage._slash_edit`](https://github.com/discord-py-slash-commands/discord-py-interactions/blob/a7f468b66137726a2535a9d62177f6e25a8f8d2f/discord_slash/model.py#L492)) do not support using an empty list `[]` for the `embeds` nor `components` keyword argument, which prevents removing them from a message.

AstreaTSS:
## About this pull request

In short, the current system for the edit function has a number of flaws, like not allowing empty lists to be passed to components and not allowing `None` values to be passed to `content`.

This PR attempts to fix that by making it more similar to discord.py's current system for edits, which can more accurately detect which type of changes someone may want.

This also fixes #228.

(NOTE: this is my first major PR. I might have messed up something. Also you might argue that this might be breaking, but not by much. Please do give feedback regardless.)

## Changes

Pretty much all changes were made exclusively to `ComponentContext.edit_origin` and `SlashMessage._slash_edit`. You can see the changes for yourself, but they've been changed to be handled more or less like how `discord.py` handles edits in 1.X (if it changed in 2.0 I wouldn't know). 

## Checklist

    - Issue: #228 

### DIFF
--- a/discord_slash/context.py
+++ b/discord_slash/context.py
@@ -377,29 +377,44 @@ class ComponentContext(InteractionContext):
         if content:
             _resp["content"] = str(content)
 
-        embed = fields.get("embed")
-        embeds = fields.get("embeds")
         file = fields.get("file")
         files = fields.get("files")
         components = fields.get("components")
 
-        if components:
+        if components is not None:
             _resp["components"] = components
 
-        if embed and embeds:
-            raise error.IncorrectFormat("You can't use both `embed` and `embeds`!")
-        if file and files:
+        if files is not None and file is not None:
             raise error.IncorrectFormat("You can't use both `file` and `files`!")
         if file:
             files = [file]
         if embed:
             embeds = [embed]
-        if embeds:
-            if not isinstance(embeds, list):
-                raise error.IncorrectFormat("Provide a list of embeds.")
-            elif len(embeds) > 10:
-                raise error.IncorrectFormat("Do not provide more than 10 embeds.")
-            _resp["embeds"] = [x.to_dict() for x in embeds]
+            
+            
+         # Check if the embeds interface is being used
+        try:
+            embeds = fields['embeds']
+        except KeyError:
+            # Nope
+            pass
+        else:
+            if embeds is None or len(embeds) > 10:
+                raise error.IncorrectFormat('embeds has a maximum of 10 elements')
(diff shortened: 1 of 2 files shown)


discord_slash/context.py
```diff
@@ -373,33 +373,57 @@ async def edit_origin(self, **fields):
...
+
+        try:
+            components = fields["components"]
+        except KeyError:
+            pass
+        else:
+            if components is None:
+                _resp["components"] = []
+            else:
+                _resp["components"] = components
```
LordOfPolls, line 393: Would it not make sense to duplicate this code block for `embeds`

This would avoid "Provide a list of embeds" errors when passing None in the embeds field, which would likely mean the user wants to remove the embeds

LordOfPolls:
Tested and this works, however i have one minor suggestion

discord_slash/context.py
AstreaTSS: Hmmm... probably? What's interesting, though, is that discord.py _doesn't_ allow for that type of behavior, as you can see [here](https://github.com/Rapptz/discord.py/blob/462ba84809f43349296c44cbe6468f631a00edab/discord/webhook.py#L1046) in the v1.x branch in `webhooks.py`.

I believe v2.0 also doesn't allow for this behavior? You can take a look [here](https://github.com/Rapptz/discord.py/blob/d30fea5b0dcba9cd130026b56ec01e78bd788aff/discord/webhook/async_.py#L440), but passing in a `None` value there would error out the code, I believe.

I'll let you decide, but honestly the best option might be to remove the `None` check if `embeds` is passed but still let it eventually error out when it tries to iterate over `None`.

LordOfPolls: I'm curious what their rationale behind that would have been. But oh well, its probably best if we mirror d.py's behaviour

LordOfPolls (approved):

LordOfPolls:
Merging, thanks for contributing

PR #232, opened 2021-06-25
Fix URL buttons context processing

artem30801:
## About this pull request

Context tries to grab component by id, but url buttons don't have one so an exception is thrown. That's a fix. Also fixes documentation formatting

## Changes

- Added a check to `ComponentMessage.get_component()` to ensure that component have custom_id in
- Fix doc formatting 

## Checklist

    - Issue:


LordOfPolls (approved):

benwoo1110:
Isn't there is a .get() method for python dictionary that prevents it? e.g. `component.get("custom_id")`

artem30801:
`component.get("custom_id")` would still raise KeyError, `component.get("custom_id", None)` though will return None if key is not found. So I just prefered more explicit check in this context. Maybe it can be compared in some benchmark
Edit: Okay I was a bit wrong here, `component.get("custom_id")` still returns None by default

LordOfPolls (approved):

benwoo1110:
`component.get("custom_id")` won't raise a key error btw, but anyways LordOfPolls says it doesn't matter, the bug fix is more important

benwoo1110 (approved):
👍

PR #239, opened 2021-06-30
Fix example in 'Component callbacks' section

Toricane:
## About this pull request

Fix example in 'Component callbacks' section

## Changes

The function in the example had a parameter called button_context, but ctx was used instead inside the defined function, so I replaced button_context with ctx.

## Checklist

    - Issue:


LordOfPolls:
Merging, thanks for contributing

PR #242, opened 2021-07-01
Add values to ComponentContext for select menu components

AstreaTSS:
## About this pull request

Select menus have finally came out. Select menus return a value for each item selected during an interaction, which is useful if you wish to do something when a specific value is selected. However, the current version of the library has no support for reading these values. This PR attempts to fix that.

## Changes

There were only two lines changed. In short, I added `values` to `ComponentContext`, which is an _optional_ list of, well, values that may come from the interaction. I also changed the docstrings to reflect this change (although I'll admit the line might need a bit of changing somehow).

## Checklist

    - Issue:


LordOfPolls:
I'm declining this pr simply because we've had a working build of selects for a while, just been waiting for release so I can push it

PR #246, opened 2021-07-01
How to send selects

Toricane:
## About this pull request

How to send selects

## Changes

The example of how to create and react to a select did not show how to send a select in the first place. I added an extra line showing how to send a select.

## Checklist

    - Issue:

### DIFF
--- a/docs/components.rst
+++ b/docs/components.rst
@@ -117,11 +117,11 @@ Each button gets a ``custom_id`` (which is always a string), this is a unique id
 What about selects / Dropdowns?
 _______________________________
 
-Yep we support those too. You use them much the same as buttons:
+Yep we support those too. You use them much the same as buttons, with 25 options per select, 1 select in each action row, 5 action rows in a message, totalling 125 options:
 
 .. code-block:: python
 
-    from discord_slash.utils.manage_components import create_select, create_select_option
+    from discord_slash.utils.manage_components import create_select, create_select_option, create_actionrow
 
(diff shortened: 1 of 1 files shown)


Toricane:
Made another change to this PR which fixed the example since you can't send a message without content, and explains the limits

docs/components.rst
```diff
@@ -117,11 +117,11 @@ Each button gets a ``custom_id`` (which is always a string), this is a unique id
 What about selects / Dropdowns?
 _______________________________
 
-Yep we support those too. You use them much the same as buttons:
+Yep we support those too. You use them much the same as buttons, with 25 options per select, 1 select in each action row, 5 action rows in a message, totalling 125 options:
```
LordOfPolls:
```suggestion
Yep we support those too. You use them much the same as buttons. You can only have 1 select per action row, but each select can have up to 25 options in it!
```

I really dont think we should be encouraging pushing selects to their limits with 125 options. That'd be incredibly bad UX

Toricane: Ok yeah that looks better

Toricane:
Um the check failed:

would reformat discord_slash/utils/manage_components.py
Oh no! 💥 💔 💥
1 file would be reformatted, 14 files would be left unchanged.
Error: Process completed with exit code 1.

There is something wrong with manage_components.py?

LordOfPolls:
Weird it didn't catch that when I merged last. I'll look into it

PR #253, opened 2021-07-03
Make `sync_all_commands` errors easier to understand

LordOfPolls:
## About this pull request

Convert the indexes representing commands into command names

## Changes

- Wrapped `sync_all_commands`'s push into a try-except 
- Parse the discord error message for command indexes
- Determine which command corresponds to which index

## Checklist

    - Issue:

### DIFF
--- a/discord_slash/client.py
+++ b/discord_slash/client.py
@@ -431,10 +431,8 @@ class SlashCommand:
                             )
 
                         ex.args = (error_string,)
-                        raise ex
 
-                    else:
-                        raise ex
+                    raise ex
             else:
                 self.logger.debug(
                     f"Detected no changes on {scope if scope is not None else 'global'}, skipping"
--- a/discord_slash/client.py
+++ b/discord_slash/client.py
@@ -1,5 +1,6 @@
 import copy
 import logging
+import re
 import typing
 from contextlib import suppress
 from inspect import getdoc, iscoroutinefunction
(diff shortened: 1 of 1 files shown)


discord_slash/client.py
```diff
@@ -410,9 +411,30 @@ async def sync_all_commands(
...
+                            error_string = error_string.replace(
+                                f"In {num}",
+                                f"'{error_command.get('name')}'",
+                            )
+
+                        ex.args = (error_string,)
+                        raise ex
+
+                    else:
+                        raise ex
```
benwoo1110:
```suggestion

                    raise ex
```

Going to raise an exception regardless, the else part seems a bit redundant?

discord_slash/client.py
```diff
@@ -410,9 +411,30 @@ async def sync_all_commands(
...
-                )
+                try:
+                    existing_cmds = await self.req.put_slash_commands(
+                        slash_commands=to_send, guild_id=scope
+                    )
+                except discord.HTTPException as ex:
+                    if ex.status == 400:
+                        # catch bad requests
+                        cmd_nums = set(
+                            re.findall(r"In\s(\d).", ex.args[0])
```
benwoo1110, line 422: What do you think about using `re.compile` it beforehand?

LordOfPolls: Due to how rarely this code should execute I didn't feel it was necessary.

No harm in adding it, but also very little benefit

discord_slash/client.py
LordOfPolls: Very true. I'll rectify that

discord_slash/client.py
benwoo1110: true, its your call lol

LordOfPolls: I'll add it. I guess theoretically someone could have hundreds of commands and that could take a while, relatively, to process

LordOfPolls: This code executes once per sync, on error. In which case it stops execution and raises an error. It only runs the regex once. So I won't add compile to it

(putting this here to explain my logic if it comes up later)

benwoo1110: Yea I get what you mean, the regex is only used if their program is bound to not work anyways.

benwoo1110 (approved):
looks good!

LordOfPolls:
Requesting one more review

FayeDel (approved):
Tested, LGTM.

LilSpazJoekp (approved):
LGTM

i0bs (approved):

PR #255, opened 2021-07-03
Fix KeyError in ComponentContext

TanguyMorelle:
## About this pull request

Fix a KeyError raised when deselecting all choices of a Select with multiple choices

## Changes

Set the `selected_options` attribute to and empty list when the `value` key is missing

## Checklist

    - Issue:

### DIFF
--- a/discord_slash/context.py
+++ b/discord_slash/context.py
@@ -324,9 +324,7 @@ class ComponentContext(InteractionContext):
         self.selected_options = None
 
         if self.component_type == 3:
-            self.selected_options = (
-                _json["data"]["values"] if "values" in _json["data"].keys() else []
-            )
+            self.selected_options = _json["data"].get("values", [])
 
     async def defer(self, hidden: bool = False, edit_origin: bool = False):
         """
--- a/discord_slash/context.py
+++ b/discord_slash/context.py
@@ -321,7 +321,12 @@ class ComponentContext(InteractionContext):
             )
             self.component = self.origin_message.get_component(self.custom_id)
 
-        self.selected_options = _json["data"]["values"] if self.component_type == 3 else None
+        self.selected_options = None
+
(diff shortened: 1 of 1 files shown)


LordOfPolls:
Hey, can you run `pre-push.py` as right now this code violates our coding style.

Besides that, good catch, waiting to merge

TanguyMorelle:
Should be good now !

LordOfPolls:
Why'd you force push?

TanguyMorelle:
Just to amend the commit, you'd rather I make a second one ? (I squashed the modification commit into the first one)

discord_slash/context.py
```diff
@@ -321,7 +321,12 @@ def __init__(
             )
             self.component = self.origin_message.get_component(self.custom_id)
 
-        self.selected_options = _json["data"]["values"] if self.component_type == 3 else None
+        self.selected_options = None
+
+        if self.component_type == 3:
+            self.selected_options = (
+                _json["data"]["values"] if "values" in _json["data"].keys() else []
```
benwoo1110: `.keys()` is not needed?

LilSpazJoekp: Correct. Doing `item in dictionary` same as `item in dictionary.keys()`.

discord_slash/context.py
```diff
@@ -321,7 +321,12 @@ def __init__(
...
+            )
```
LilSpazJoekp:
```suggestion
            self.selected_options = _json["data"].get("values", [])
```
This can be simplified.

LilSpazJoekp (changes requested):

discord_slash/context.py
LordOfPolls: This has been resolved

LordOfPolls:
Thanks for contributing, merging

PR #261, opened 2021-07-13
Fixed cmd_name parameter description.

Tripletri:
## About this pull request

Fixed some documentation issues and updated FAQ slightly.

## Changes

- Description of parameters.
- FAQ.

## Checklist

    - Issue:
- [X] This is not a code change. (README, docs, etc.)

### DIFF
--- a/discord_slash/http.py
+++ b/discord_slash/http.py
@@ -39,7 +39,7 @@ class SlashCommandRequest:
         """
         Sends a slash command delete request to Discord API.
 
-        :param guild_id: ID of the guild to add command. Pass `None` to add global command.
+        :param guild_id: ID of the guild to remove command. Pass `None` to remove global command.
         :param cmd_id: ID of the command.
         :return: Response code of the request.
         """
@@ -49,7 +49,7 @@ class SlashCommandRequest:
         """
         Sends a slash command get request to Discord API for all commands.
 
-        :param guild_id: ID of the guild to add command. Pass `None` to add global command.
+        :param guild_id: ID of the guild to get commands. Pass `None` to get global commands.
         :return: JSON Response of the request.
         """
         return self.command_request(method="GET", guild_id=guild_id)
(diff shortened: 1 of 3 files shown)


LordOfPolls:
This pr fails the lint checks, please resolve that before i can merge it

discord_slash/http.py
```diff
@@ -79,7 +79,7 @@ def add_slash_command(self, guild_id, cmd_name: str, description: str, options:
         Sends a slash command add request to Discord API.
 
         :param guild_id: ID of the guild to add command. Pass `None` to add global command.
-        :param cmd_name: Name of the command. Must be 3 or longer and 32 or shorter.
+        :param cmd_name: Name of the command. Must be match the regular expression ``^[a-z0-9_-]{1,32}$``.
```
LordOfPolls, line 82: You didn't need to completely change the line, you just needed to escape the `\w`
```suggestion
        :param cmd_name: Name of the command. Must be **lower-case** and match the regular expression ``^[\\w-]{1,32}$``.
```

LordOfPolls (changes requested):

discord_slash/http.py
Tripletri: `^[\\w-]{1,32}$` and `^[\w-]{1,32}$` are different regexes. It is not obvious that the slash is escaped. If you just copy it, the result will be incorrect. `\w` is equivalent to `[a-zA-Z0-9_]`, given that they must be in lowercase, we get `^[a-z0-9_-]{1,32}$`.

LordOfPolls: However for rendering in sphinx, `\\w` would render as `\w`, and the majority of the time docstrings will be viewed within sphinx, no?

Tripletri: If `^[\w-]{1,32}$` considering lowercase and `^[a-z0-9_-]{1,32}$` are the same, why don't we make it correct in both cases?

LordOfPolls: Reasonable, this isnt worth arguing over, at the end of the day its an exceptionally unimportant debate, as you say, the two result in the same regex. I was only being picky about it because obviously the former is actually the regex pattern used by Discord

LordOfPolls (approved):

LordOfPolls:
Merged, thanks for contributing

PR #262, opened 2021-07-13
Introduction to some d.py Aliases

FayeDel:
## About this pull request

This PR introduces d.py Aliases to resolve a 2.0.x issue (not created in Github), which pertains to the use of d.py's `commands.bot_has_guild_permissions()` check. The check references `ctx.me`, which doesn't exist in Slash invocation context.

## Changes

Adds `.me`, `.voice_client`, `.reply`, `.slash` and `.cog` to `InteractionContext` and `SlashContext`.

This remedies https://discord.com/channels/789032594456576001/801461008865951784/864602817394114610 and `reply` attr is based on https://discord.com/channels/789032594456576001/801461008865951784/865182138084491274

(Ignore the first 8 commits, Github is just being weird)

## Checklist

    - Issue:
- [X] (If required) Relevant documentation has been updated/added.


LordOfPolls:
Is this complete?

FayeDel:
Not sure what other Aliases needed to be added, so at the moment yeah it's finalized.

Tripletri:
Can we add  `self.slash: client.SlashCommand = _discord.slash` to `InteractionContext`, it will be helpfull, and we can get rid of ` # noinspection PyUnresolvedReferences`

FayeDel:
Not sure why self.slash should be added, rather I don't see how it can be helpful to use `ctx.slash` compared to `bot.slash`/`client.slash`, mind elaborating?

Tripletri:
`ctx.bot` returns the instance of `discord.Client`, which itself don't have `.slash`, its adding at runtime via `SlashCommand(bot)`. So it's not obvious for me and IDE that, the ctx.bot has `.slash`. And I every time must to write something like this, to get IDE auto completion.
```py
from discord_slash import SlashCommand
slash: SlashCommand = context.bot.slash
```
And I thought it might be helpful. At least it would make my life a little easier.

LordOfPolls (approved):

discord_slash/context.py
```diff
@@ -114,6 +114,22 @@ def channel(self) -> typing.Optional[typing.Union[discord.TextChannel, discord.D
...
+
+        """
+        return self.guild.voice_client if self.guild else None
+
+    @property
+    def me(self):
+        """
+        Similar to :attr:`.Guild.me` except that it may also return the bot user in DM context.
+        """
+        return self.guild.me if self.guild is not None else self.bot.user
```
i0bs: `is not` produces a raised `SyntaxWarning` which may not me mutable to some users when logged. Please consider using `!=` instead due to strict checking. The `is` operator checks for identity correlation which should be used in the event where your value can always be assured it has `None` as a value, but otherwise should be using the prior.

i0bs (changes requested):
meow

i0bs (approved):

i0bs:
Merging the changers, thank you for contributing.

PR #263, opened 2021-07-19
[Addition] Button based embed paginator

JUGADOR123:
## About this pull request

Adds a basic handler for interactive multiple page embeds

## Changes

What changes were made?
 Addition of Paginator.py under utils
## Checklist

    - Issue:

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,7 @@ embed3==discord.Embed(title="yet another title")
 pages=[embed1,embed2,embed3]
 
 await Paginator(content="A message",pages=pages)
+```
 
 
 ### Advanced
--- a/README.md
+++ b/README.md
@@ -117,6 +117,24 @@ action_row = create_actionrow(*buttons)
 await ctx.send(components=[action_row])
 ```
 
+## Paginator
+This simple example shows how to easily create interactive, multiple page embeds that annyone can interact with.
+```py
+import discord
+import discord_slash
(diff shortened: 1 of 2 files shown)


benwoo1110:
I feel like this should be an example in https://github.com/discord-py-slash-commands/slash-bot where ppl can test out on our server, refer to the code and adapt to their own bot. Shouldn't really be in this lib imo.

Catalyst4222:
There's a paginator for d.py (commands.Paginator). I think it's a good idea to have a button variant instead of the reaction based one, since buttons were partially made as a replacement for reaction menus

AstreaTSS:
> There's a paginator for d.py (commands.Paginator). I think it's a good idea to have a button variant instead of the reaction based one, since buttons were partially made as a replacement for reaction menus

I suppose you could say that. But there's a number of ways of writing a paginator like this - some would inherit from discord.py's pagniator, some would use different labels, etc. That makes establishing a standard much more difficult.

Also consider the fact that the default pagniator (and so, the default help command) does not use reactions at all (at least, I think so - doesn't look like there's any code to handle it). Most reaction-based pagniators are made by other people, and so button-based pagnitors, at least in my opinion, should too.

I think this would be great as a code example, like @benwoo1110 said. But it probably doesn't belong in the library itself.

LordOfPolls:
Not going to rehash what's already been said.

Closing this PR. Feel free to make an example for Slash-Bot

PR #268, opened 2021-07-26
Allow component response to be ignored

eunwoo1104:
## About this pull request

This allows user to ignore component's response and prevent `ConponentContext.send` to work as unintended.
By using `ignore=True`, user can optionally use component's send method without loading state or whatever errors.
Example:
```py
@client.event
async def on_message(message: discord.Message):
    select = manage_components.create_button(style=3, label="Select", emoji="✅", custom_id=f"select{message.id}")
    row = manage_components.create_actionrow(select)

    msg = await message.reply("Using no option for `defer`.", components=[row])
    ctx = await manage_components.wait_for_component(client, msg, row)
    await ctx.defer()  # Now since we didn't respond to this interaction, user will see loading state.

    msg = await message.reply("Using `edit_origin=True` for `defer`.", components=[row])
    ctx = await manage_components.wait_for_component(client, msg, row)
    await ctx.defer(edit_origin=True)
    await ctx.send("But I want this as separate message!", hidden=True)  # However, this will neither be hidden nor separate message.

    msg = await message.reply("Using `ignore=True` for `defer`.", components=[row])
    ctx = await manage_components.wait_for_component(client, msg, row)
    await ctx.defer(ignore=True)
    await ctx.send("Now this is separate message!", hidden=True)  # This finally works as intended, without any errors.
```

(Second case message is edited.)

## Changes

Added `ignore` parameter to `ComponentContext.defer`.

## Checklist

    - Issue:

### DIFF
--- a/discord_slash/context.py
+++ b/discord_slash/context.py
@@ -326,17 +326,18 @@ class ComponentContext(InteractionContext):
         if self.component_type == 3:
             self.selected_options = _json["data"].get("values", [])
 
-    async def defer(self, hidden: bool = False, edit_origin: bool = False):
+    async def defer(self, hidden: bool = False, edit_origin: bool = False, ignore: bool = False):
         """
         'Defers' the response, showing a loading state to the user
 
-        :param hidden: Whether the deferred response should be ephemeral . Default ``False``.
+        :param hidden: Whether the deferred response should be ephemeral. Default ``False``.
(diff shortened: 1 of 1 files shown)


LordOfPolls (approved):

FayeDel (approved):
Tested, LGTM 👍🏻

PR #288, opened 2021-08-12
[v4] Library Rewrite

i0bs:
## About this pull request

This merges all changes made from unstable into the master as the new v4.0 branch.

## Changes

A lot.

## Checklist

    - Issue:


i0bs:
#332

i0bs:
I'm moving this from a draft PR to an open one since commands are now functional and we can begin receiving information on issues.

interactions/api/http.py
```diff
@@ -0,0 +1,2285 @@
...
+
+    @property
+    def bucket(self) -> str:
+        """
+        Returns the route's bucket.
+
+        :return: The route bucket.
+        :rtype: str
+        """
+        return f"{self.channel_id}:{self.guild_id}:{self.path}"
```
quirky-bluejay, line 85: `webhook_id + webhook_token` should also be used for bucketing interaction requests (iirc blue requested this a while back)

interactions/api/http.py
```diff
@@ -0,0 +1,2285 @@
...
+    """
+
+    __slots__ = ("token", "loop", "ratelimits", "headers", "session", "lock")
+    token: str
+    loop: AbstractEventLoop
+    ratelimits: dict
+    headers: dict
+    session: ClientSession
+    lock: Event
+    limit: int
```
quirky-bluejay: This should be removed as ratelimits should not be hard coded

interactions/api/http.py
```diff
@@ -0,0 +1,2285 @@
...
+
+                try:
+                    reason = kwargs.get("reason")
+                except:  # noqa
+                    pass
+                else:
+                    if reason:
+                        kwargs["headers"]["X-Audit-Log-Reason"] = quote(reason, safe="/ ")
+
+                kwargs["headers"]["Content-Type"] = "application/json"
```
quirky-bluejay: Minor implementation detail, but what about doing this before the for loop?

interactions/api/http.py
```diff
@@ -0,0 +1,2285 @@
...
+                        log.info(
+                            "The application has reached the last remaining request. Cooling down to avoid rate limiting."
+                        )
+                        self.lock.set()
+                        await sleep(1)
+                        self.lock.clear()
+                    else:
+                        self.limit = response.headers["X-Ratelimit-Remaining"]
+                        data = await response.json(content_type=None)
+                        log.debug(data)
```
quirky-bluejay: Remove this, and replace it with logic that:

if `X-Ratelimit-Remaining` less than one
- If `X-Ratelimit-Global` or `X-Ratelimit-Scope` = 'global' (doesn't matter which)
-  - lock global lock and wait `X-RateLimit-Reset-After` milliseconds and unlock
- else `X-Ratelimit-Scope` = 'user'
- - lock per request ratelimit (this will be the `ratelimit` var and wait `X-RateLimit-Reset-After` milliseconds and unlock

Sorry about the *horrible* pseudocode, if you want i can put this into python but that'll take me a bit
here's the d.py implementation (yours should be a little different i think, just the way everything else has been setup) https://github.com/Rapptz/discord.py/blob/45d498c1b76deaf3b394d17ccf56112fa691d160/discord/http.py#L281

interactions/api/http.py
```diff
@@ -0,0 +1,2285 @@
...
+                        retry_after = data["retry_after"]
+
+                        if "X-Ratelimit-Global" in response.headers.keys():
+                            self.lock.set()
+                            log.warning("The HTTP request has encountered a global API ratelimit.")
+                            await sleep(retry_after)
+                            self.lock.clear()
+                        else:
+                            log.warning("A local ratelimit with the bucket has been encountered.")
+                            await sleep(retry_after)
```
quirky-bluejay: This should also lock the per route ratelimit

quirky-bluejay (commented):
This *is* based off discord.py's ratelimit implementation

I also may have missed something, not too sure

For a more 'correct' implementation `X-RateLimit-Bucket` should be also taken into consideration, for instances where multiple routes share the same bucket 
([here's hikari's implementation of this](https://github.com/hikari-py/hikari/blob/1552ace21c3c5c9449728563641c486a4566af99/hikari/impl/buckets.py#L32) )

> "Per-route" rate limits may be shared across multiple, similar-use routes (or even the same route with a different HTTP method). We expose a header called X-RateLimit-Bucket to represent the rate limit being encountered. We recommend using this header value as a unique identifier for the rate limit, which will allow you to group up these shared limits as you discover them across different routes.

i0bs:
For all intents (as of right now I'm working on other integrations) do you please mind writing the Python equivalent code for the new ratelimiting logic? It looks more correct from as far as I can tell but "priorities" have gotten the upper hand of me right now 😔 @AnotherCat

interactions/api/http.py
quirky-bluejay:
```py
remaining = response.headers["X-Ratelimit-Remaining"]

if remaining == '0' and response.status != 429:

    time_left = ## Somehow parse `X-Ratelimit-Reset-After`
    # Lock the `ratelimit` (never used asyncio.Lock before so have fun :) )
    await asyncio.sleep(time_left)
    # Unlock `ratelimit`

```
I *think* discord won't return headers to avoid the global ratelimit, this is something to double check

interactions/api/http.py
i0bs: This already should be doing that

VincentRPS:
a-lot if not most of the api docs files could just be put into 1 file instead of being so split

RPSMain:
a-lot if not most of the api docs files could just be put into 1 file instead of being so split

i0bs:
All current major bugs that have been found/sorted as critical problems to the library have all been addressed and fixed where needed. We will be preparing to merge this PR onto `stable` soon.

i0bs:
We will be finalizing the code review today and tomorrow, and pending the status, merge into `stable`.

FayeDel:
LGTM 👍🏻

FayeDel (approved):

PR #309, opened 2021-08-20
Fix sync_commands with options, choices = []

JellyWX:
## About this pull request

Fixes a bug where sync_commands would detect changes since one of the following comparisons would occur and fail:
* `options = []` on the local compared to `options = None` from remote
* `choices = []` on the local compared to `choices = None` from remote

## Changes

Instead of `self.choices = None`, if choices is unspecified, `self.choices = []` is used

## Checklist

### DIFF
--- a/discord_slash/client.py
+++ b/discord_slash/client.py
@@ -433,9 +433,8 @@ class SlashCommand:
         for guild in cmds["guild"]:
             cmds_formatted[guild] = cmds["guild"][guild]
 
-        for scope in cmds_formatted:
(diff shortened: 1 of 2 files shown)


i0bs:
Tbqh I don't have any plans to merge/commit because I will not be focusing on any future v3.0 release builds since v4.0 is actively under development, however, I will keep this as a considered design choice for the next release.

JellyWX:
Okay, sure. This resolves a bug though, so if you won't merge this, perhaps convert it to an issue so it doesn't crop up in version 4

-------- Original Message --------

PR #311, opened 2021-08-22
Aliases!

ActuallyFlamey:
## About this pull request

This pull request adds a Slash Command aliases argument to the decorators. Decided not to mess with subcommands (what would aliases for subcommands even be useful for?).

In the argument description, it warns the user that it will count up their command limit.

## Changes

- Added argument to `@slash.slash` and `@cog_ext.cog_slash` decorators for command aliases.

## Checklist

    - Issue:

### DIFF
--- a/discord_slash/client.py
+++ b/discord_slash/client.py
@@ -826,6 +826,7 @@ class SlashCommand:
         name: str = None,
         description: str = None,
         guild_ids: typing.List[int] = None,
+        aliases: typing.List[str] = None,
         options: typing.List[dict] = None,
         default_permission: bool = True,
         permissions: dict = None,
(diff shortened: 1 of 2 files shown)


LordOfPolls:
I don't understand why aliases are needed for Slash commands

ActuallyFlamey:
> I don't understand why aliases are needed for Slash commands

Me neither, but it will definitely satisfy the guys who keep asking for them.
Now they can get a different answer other than "fuk u dumbass API doesn't support", lol.

JUGADOR123:
i dont think that satisfying people that dont understand how slash commands work would bring anything positive, at an extreme it will bring people that have used all 50 commands even though they only have 20 "real" (non alias) commands

ActuallyFlamey:
> i dont think that satisfying people that dont understand how slash commands work would bring anything positive, at an extreme it will bring people that have used all 50 commands even though they only have 20 "real" (non alias) commands

The argument description explains it, so if they don't read it, we can tell them, when they come crying for help, that it clearly says that it counts up to the limits, so they should remove some aliases.

i0bs:
I'm gonna be that guy who will say "API doesn't support." I see many issues with aliases being established in this fashion:

- You quickly hit the global/guild command limit
- You **very** quickly hit the interaction creation limit (so now you can't make commands for another 24 hours)
- This is a very spammy solution that hacks away at the current codebase
- You risk running into double commands

If people want aliases, they're going to need to wait for the API to support them.

benwoo1110:
yea, literally no point in aliases, you can tab-complete slash commands.

i0bs:
I'm going to be closing this PR because I do not think it would be a wise decision implementing it at this time when there is a lack of API support for it as well as the problems aforementioned.

PR #319, opened 2021-08-29
emoji_to_dict fix

flfy:
## About this pull request

fixes issue where emojis passed as strings aren't returned as a properly formatted dict

## Changes

see above

## Checklist

    - Issue:

### DIFF
--- a/discord_slash/utils/manage_components.py
+++ b/discord_slash/utils/manage_components.py
@@ -1,6 +1,7 @@
 import logging
 import typing
 import uuid
+import re
 
 import discord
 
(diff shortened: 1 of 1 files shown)


i0bs:
Couple of things:

1. That check is regex-bound based off of certain conditions of the emoji structure, which I want to really avoid unless absolutely required by the API.
2. Is the `animated` field a necessity?

The reason why the dict appears as it is -- is because of how we want to define the data model for emojis to be interpreted in a button/select menu, so we're working with the bare minimums. You can also always just use `discord.Emoji` to get that as well.

PR #328, opened 2021-09-05
Weakly typed CommandData and PermissionData comparison.

MathiasSven:
## About this pull request

Fixes a bug where sync_commands would detect changes in both permissions and commands due to inconsistent types across the library.

## Changes

Changes were made as to `CommadData.__eq__` and `PermissionData.__eq__` method, making it more weakly typed, eg: str guild_id will be equal to int guild_id. These changes are **nonbreaking**.

Another change was that even though context commands do not have a description, poping them will result in having the description set to None. However, Discord API returns the description as an empty string, hence setting it to an empty string as a solution over poping it/setting it to `NoneType`.

There was also a typo, (I assume) in `PermissionData.__eq__` where the library tries to compare a permission type to permission id.

Lastly, I also slid in the readme update for the png.

## Checklist

    - Issue:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 <div align="center">
     <a href="https://pypi.org/project/discord-py-interactions/">
-        <img src="https://raw.githubusercontent.com/muqshots/discord-py-interactions/master/.github/banner_transparent.png" alt="discord-py-interactions" height="128">
+        <img src="https://raw.githubusercontent.com/goverfl0w/discord-interactions/stable/.github/banner_transparent.png" alt="discord-py-interactions" height="128">
     </a>
     <h2>Your ultimate Discord interactions library for <a href="https://github.com/Rapptz/discord.py">discord.py</a>.</h2>
 </div>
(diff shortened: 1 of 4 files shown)


i0bs:
This design choice is practical because of the way identity checks work in Python. You could also easily fix this with an absolution check `===` and see if they're the same data being passed to one another. We kept it as a loosely-checked identity comparison because of the way permission IDs are handled and declared in our library. JSON will always be a serialized string format when an HTTP/1.1 request is parsed (since this is what is currently being used with the Discord API) but our library has them written as integers. So it naturally makes sense that permissions are naturally handled with synchronous events of commands since they are supposed to fall into that as well.

MathiasSven:
I am not entirely sure what is meant by your comment. The PR is meant to fix bugs that come from trying to compare these data types in the least disruptive way possible.

i0bs:
My comment means that this was an intention design choice, it's meant to be doing an identity check rather than a strict type check.

PR #286, opened 2021-08-11
debug_guild option for testing global commands

JellyWX:
## About this pull request

Whilst using this module, I wanted to easily switch my global commands to a guild since they update instantly on guilds

## Changes

A new keyword parameter `debug_guild` is added, which if set will redirect global commands to the guild when commands are synced (and if unset or set as `None`, will revert to setting global commands)

## Checklist

- [ y ] I've run the `pre_push.py` script to format and lint code.
- [ y ] I've checked this pull request runs on `Python 3.6.X`.
- [ n ] This fixes something in [Issues](https://github.com/eunwoo1104/discord-py-slash-command/issues).
- [ y ] This adds something new.
- [ n ] There is/are breaking change(s).
- [ y ] (If required) Relevant documentation has been updated/added.
- [ n ] This is not a code change. (README, docs, etc.)

PR #329, opened 2021-09-05
Added option to disable permission syncing

MathiasSven:
## About this pull request

Adds the option to disable permission syncing when sync_commands. Why? It allows, if the user chooses, to implement his own permission system with discord's API while still benefiting from command_syncing. This feature is nonbreaking and if ignored changes nothing by default. That is if sync_commands is set to True, and sync_permission isn't passed as either True or False, It will default to True.

## Changes

Prevents `SlashCommand.sync_all_commands` from syncing permissions if `sync_premission` set to False. The option is passed through SlashCommand(args) the same way `sync_commands` is.

## Checklist

    - Issue:

### DIFF
--- a/discord_slash/client.py
+++ b/discord_slash/client.py
@@ -29,6 +29,8 @@ class SlashCommand:
     :type client: Union[discord.Client, discord.ext.commands.Bot]
     :param sync_commands: Whether to sync commands automatically. Default `False`.
     :type sync_commands: bool
+    :param sync_permissions: Whether to also sync permissions automatically when sync_commands set True. Default `True`.
+    :type sync_permissions: bool
     :param debug_guild: Guild ID of guild to use for testing commands. Prevents setting global commands in favor of guild commands, which update instantly
     :type debug_guild: int
     :param delete_from_unused_guilds: If the bot should make a request to set no commands for guilds that haven't got any commands registered in :class:``SlashCommand``. Default `False`.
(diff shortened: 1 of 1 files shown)


i0bs:
The code while being idiomatic doesn't seem to intuitively provide anything else than something already considered a synchronously changed factor of command models. Permissions as per the API are still appended to the Application Command not as a separate table. I have no criterion for things like this but this PR seems rather inclusive to helping you with your case and most likely (in my eyes) won't be used by lots of people.

Why?
A simple workaround to this is only syncing your commands **when needed.** You should not be leaving your `sync_commands` keyword-argument to always stay true, because that eventually leads to a potential disconnect between the caching state of commands from the API to the client, which causes the notorious "Invalid application command ID" we all love and know.

I'll keep this PR open for others to contribute to but I see no reason to implement this at all.

MathiasSven:
While this problem did come from a personal need, I PRd because I don't think this really is an edge case. Given how restrictive having hardcoded permissions into the bot is, I believe many will opt for a dynamic system, where a few checked commands are used to update Discord's API permission cache to grant/revoke permissions for other commands.

Currently, if someone decides to do such a system they might stumble on the fact that by setting no permissions on a server with id XYZ they can implement this system (The bot won't erase permissions from the server), but as soon as just one command permission for that server is set up locally, all the currently cached permissions on discord are erased (if sync commands are turned on, or if they eventually are).

Also, in my mind, syncing when needed via a parameter on bot initialization makes it less of a "set and forget" and more of a CLI to interact with discord's side. With that said if syncing is not a pillar of the library, and is itself also a workaround, I understand the decision.

i0bs:
We already provide warnings for bot developers to not leave this on as a "set and forget" functionality. The practical use of syncing is only to be one when needed as I mentioned, because of this it is not really our fault if there are others out there inappropriately using it to gage updated and elevated changes with the amount of HTTP requests sent to the Discord API.

This PR is a lot like #286 in the premise that it allows further versatility with debugging and testing your commands, with the only exception here being that permissions are affected because of the base command code changing. Since the API registers permissions as an integral aspect of the application command itself, there's personally no need for this to be a thing.

PR #330, opened 2021-09-06
[v4] Errors and JSON Responses

FayeDel:
## About this pull request

This PR fixes a bug with the current error system with a key lookup that was caused by an override of the base error class's dict and implements JSON error codes. This PR also repoints exceptions from `http.py` to the new Error class and is tested with the `users/@me` endpoint for testing error 403.

## Changes

(already explained)

## Checklist

    - Issue:


i0bs:
thank you delta for making my life easier <3

PR #333, opened 2021-09-06
Fix invalid code in documentation for cog support

lexicalunit:
## About this pull request
There is some invalid code in the `README.md` file. For example the `Client` object does not have a `load_extension()` method on it. I believe this was intended to be a `Bot` instead.

## Changes
Simply updating the example code so that it can actually run.

## Checklist


i0bs:
Haha *whoops*. Good catch on that, thank you for spotting this for me.

PR #343, opened 2021-09-25
Set `origin_message` for component interactions on ephemeral messages

MichaelJHarvey:
## About this pull request

Discord made an API change some time ago to send ephemeral message payloads with interactions. [Link to the announcement in the Discord Developers server.](https://discord.com/channels/613425648685547541/697138785317814292/875828152763547649) The library currently checks whether a component interaction's message is ephemeral, and skips setting the `origin_message` of the interaction if it is. I have removed this check so that the `origin_message` will be set.

## Changes

In `context.py` line 457: removed the flag check that skipped setting the `origin_message` if the message is ephemeral.

## Checklist

    - Issue:
- [ ] (If required) Relevant documentation has been updated/added. *(Docstrings have been updated, but I could not get Sphinx to run successfully)*

### DIFF
--- a/discord_slash/client.py
+++ b/discord_slash/client.py
@@ -482,8 +482,7 @@ class SlashCommand:
                         for num in cmd_nums:
                             error_command = to_send[int(num)]
                             error_string = error_string.replace(
-                                f"In {num}",
-                                f"'{error_command.get('name')}'",
+                                f"In {num}", f"'{error_command.get('name')}'"
                             )
 
                         ex.args = (error_string,)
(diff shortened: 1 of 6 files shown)


i0bs:
This is going to be labeled as a `wontfix` for 2 reasons:

1. The code currently standing already works in v3's implementation of ephemerals supporting components. It might not follow closely with the schema anymore, but in terms of functionality, it still remains present.
2. This is already something that will be getting addressed in v4 with a closer approach to following object data classes.

PR #344, opened 2021-09-27
[v4] Recursive attrs, mixins, and repair payload handling.

FayeDel:
## About this pull request

This PR fixes recursive attributes (i.e. `message.author.id` works!) and fixes decorator payloads.

This PR also marks command decorator usage (hackishly) working with differentiable results :)

## Changes

- Introduces a mixin for inheritance
- Application Command payload from decorator omits unnecessary kwargs.
- Switch `data=` to `json=` (oops)

## Checklist

    - Issue:


interactions/client.py
```diff
@@ -130,13 +134,11 @@ def decorator(coro: Coroutine) -> Any:
                 # so that it gives back what is/isn't.
 
                 _payload = {
-                    "type": type,
+                    "type": _type if not isinstance(_type, ApplicationCommandType) else _type.value,
```
i0bs: Please change to `_type.value if isinstance(_type, ApplicationCommandType) else _type`. This isn't incorrect code, it's just that my anxiety goes through the roof reading an inversion and it bothers me.

interactions/client.py
```diff
@@ -148,18 +150,17 @@ def decorator(coro: Coroutine) -> Any:
...
-                    self.me._json["id"], data=_payload, guild_id=guild
+                    self.me.id, data=_payload, guild_id=guild
                 )
 
                 _att = self.loop.run_until_complete(request)
-                print(f"request returns...: {_att}")
 
-                if _att["code"] == 50035:
-                    raise JSONException(50035)  # todo: work on this pls
+                if _att.get("code"):
```
i0bs: Have you tested this in the event that it produces back a 2XX event that is not `200 OK`? We want to comfortably show exception codes that only apply outside of that ruleset.

interactions/api/http.py
```diff
@@ -357,7 +358,7 @@ async def create_dm(self, recipient_id: int) -> dict:
         # only named recipient_id because of api mirroring
 
         return await self._req.request(
-            Route("POST", "/users/@me/channels"), data=dict(recipient_id=recipient_id)
+            Route("POST", "/users/@me/channels"), json=dict(recipient_id=recipient_id)
```
i0bs: Why are you using a `dict()` literal in this case instead of `{}` dictionary structuring everywhere else in your other class methods? Would help if the codebase is consistent in formatting.

interactions/client.py
FayeDel: When it succeeded, it didn't return a code, just the object in question, which was weird.

interactions/client.py
```diff
@@ -148,18 +150,17 @@ def decorator(coro: Coroutine) -> Any:
                     self.me = User(**data)
 
                 request = self.http.create_application_command(
-                    self.me._json["id"], data=_payload, guild_id=guild
+                    self.me.id, data=_payload, guild_id=guild
                 )
 
                 _att = self.loop.run_until_complete(request)
```
i0bs: Not against the existence of this variable but what is `_att` representing as the name? I might smooth brain 8 months later revisiting this code and wonder what the hell that means even though I understand the event loop request following it.

i0bs (changes requested):
Some general changes.

interactions/client.py
i0bs: This is mildly concerning. What if we throw a warning/info log on 2XX? That might be nice for some bot devs.

interactions/client.py
FayeDel: Fixed as of 725b9cd.

interactions/api/http.py
FayeDel: Fixed as of c64c4c4.

interactions/client.py
FayeDel: `_att` (now `req_data` as of 04237d0) is the object generated from the request, if it fails, it has the code key that the lib parses.

interactions/client.py
i0bs: We'll make this another thing sep. to this PR for commit.

i0bs (approved):

PR #349, opened 2021-10-03
fixed commands in the debug_guild not registering

rouven0:
## About this pull request

Global commands weren't registering at all when a `debug_guild` was specified and you had commands having the same guild id in their `guild_ids` attribute because the list of global commands re-routed to this guild was overridden by the normal guild commands you specified for that guild.

## Changes

I changed the code in `client.py` to check whether there are commands in the list coming from the debug_guild re-route and append the normal guild commands to them.

## Checklist

    - Issue:


benwoo1110:
Is there really a use for it to be a list tho? It’s suppose to make it such that it’s fast for development. Adding multiple guilds will make unneeded api calls to discord.

PR #358, opened 2021-10-19
Fix global commands getting overridden.

benwoo1110:
## About this pull request

Fix an edge case where some commands can specifically only be run on one guild, and also happens to be the debug_guild. This resulted in the global commands being overridden.

## Changes

Bugfix to sync method.

## Checklist

    - Issue:


i0bs:
No plans to update v3 at this time.

PR #376, opened 2021-12-13
Unstable

MeidoNoHitsuji:
- generates an appropriate name for the autocomplete event
- add argument "value" of the command in the autocomplete event
- don't be afraid to cancel if you find any error :D

PR #382, opened 2021-12-15
Fix not working selects and weird choice behavior; add defer for components and fix responding to components

sillyElvy:
## About this Pull Request

- Fixes Error when creating selectoptions
- Seperates Button and Select when sending
- Converts options to dict before sending

- when syncing options are called twice, while choices are a dict the second time not having the `._json` attribute
grafik

- Also, I added `defer` to components and fixed the problem that always the original message was updated instead of sending a new

## Checklist

- [x] I've run the `pre_push.py` script to format and lint the change(s) made.
- [x] I've checked to make sure this pull request runs on Python `3.6.X`.
- [ ] This fixes/solves something from the project's [Issues](https://github.com/goverfl0w/discord-interactions/issues).
    - Issue (if referenceable):
    - [ ] This is not a code change.
    - [ ] There is/are (a) breaking change(s).


interactions/models/command.py
```diff
@@ -110,7 +110,9 @@ def __init__(self, **kwargs) -> None:
         if self._json.get("options"):
             self._json["options"] = [option._json for option in self.options]
         if self._json.get("choices"):
-            self._json["choices"] = [choice._json for choice in self.choices]
+            self._json["choices"] = [
```
i0bs:
```
            if isinstance(self._json.get("choices"), dict):
                self._json["choices"] = [choice for choice in self.choices]
            else:
                self._json["choices"] = [choice._json for chocie in self.choices]
```

We can build a double check instead of a ternary operation that you have here.

i0bs (changes requested):
Cleaner approach than using a direct `type(value)` check. We generally use `isinstance()` for most type comparisons because we want to apply the actual instance property to a known typing alias, which works well for when we use the `typing` module.

i0bs:
Additionally, resolve merge conflicts with `gateway.py` being updated (from #376) so once changes are made, the PR may be merged.

sillyElvy:
should be all done now

i0bs (approved):
LGTM.

i0bs:
Merging this pull request, thank you for your contribution! Welcome to the contributors table. 😃

PR #386, opened 2021-12-16
new functions for models:

sillyElvy:
New functions:
`Channel`:
- Channel.send()
- Channel.delete()
- Channel.modify()

`Guild`:
- Guild.ban()
- Guild.remove_ban() 
- Guild.kick() 
- Guild.add_member_role()
- Guild.remove_member_role() 
- Guild.create_role() 
- Guild.get_member() 
- Guild.delete_channel()
- Guild.delete_role()
- Guild.modify_role() 
- Guild.create_channel()
- Guild.modify_channel() 

`Member`:
- Member.ban()
- Member.kick()
- Member.add_role() 
- Member.remove_role()
- Member.send()

`Message`:
- Message.delete()
- Message.edit() 
- Message.reply()
- Message.get_channel()
- Message.get_guild()

`Role`:
- Role.delete()
- Role.modify()

## About this Pull Request

This PR adds new functions to the existing data models (like `Message`, `Channel`, `Member`, etc).

## Checklist

- [x] I've run the `pre_push.py` script to format and lint the change(s) made.
- [x] I've checked to make sure this pull request runs on Python `3.6.X`.
- [ ] This fixes/solves something from the project's [Issues](https://github.com/goverfl0w/discord-interactions/issues).
    - Issue (if referenceable):
    - [ ] This is not a code change.
    - [ ] There is/are (a) breaking change(s).


interactions/context.py
```diff
@@ -280,7 +280,10 @@ async def edit(
 
         if isinstance(components, ActionRow):
             _components[0]["components"] = [component._json for component in components.components]
-        elif isinstance(components, (Button, SelectMenu)):
```
i0bs: What is the relation of this code here relevant to new functions for models?

interactions/api/models/message.py
```diff
@@ -277,6 +280,78 @@ def __init__(self, **kwargs):
...
+        :param allowed_mentions?: The message interactions/mention limits that the message can refer to.
+        :type allowed_mentions: Optional[MessageInteraction]
+        :param components?: A component, or list of components for the message.
+        :type components: Optional[Union[Component, List[Component]]]
+        :return: The edited message as an object.
+        :rtype: Message
+        """
+
+        _content: str = "" if content is None else content
+        _tts: bool = False if tts is None else tts
```
i0bs:
```suggestion
        _tts: bool = True if bool(tts) else tts
```

interactions/api/models/message.py
```diff
@@ -277,6 +280,78 @@ def __init__(self, **kwargs):
...
+        """
+        await self.client.delete_message(
+            message_id=int(self.id), channel_id=int(self.channel_id), reason=reason
+        )
+
+    async def edit(
+        self,
+        content: Optional[str] = None,
+        *,
+        tts: Optional[bool] = None,
```
i0bs:
```suggestion
        tts: Optional[bool] = False,
```

i0bs (changes requested):
`_tts` and `tts` fields respectively were forgotten to get updated in the context `edit` method, which I see you've copied off of and made adjustments were needed to make things work. This isn't your fault and will be addressed, but that's why those code suggestions were made.

Also, curious about the reason for `context.py` receiving a change to JSON collection values with the relevance of what this PR is for.

interactions/context.py
sillyElvy: No relation, I encountered it when making the edit, I'll put it into the other PR

interactions/api/models/guild.py
```diff
@@ -236,6 +239,144 @@ def __init__(self, **kwargs):
             else None
         )
 
+    async def ban(
```
i0bs, line 255: This is awesome, but if the ultimate goal of the helper methods here is to help add recursion to what d.py had in a similar fashion, this would make better sense to include in the `Member` object as well. having `ctx.guild.ban()` and `ctx.guild.members[member].ban()` will both be very convenient.

interactions/api/gateway.py
```diff
@@ -297,6 +297,14 @@ def check_sub_auto(option) -> tuple:
...
                         _name,
                     )
+                    if event.lower() in [
+                        "message_create",
+                        "channel_create",
+                        "guild_member_add",
+                        "guild_create",
+                        "guild_role_create",
+                    ]:
+                        data["client"] = self.http
```
i0bs: Why are you building a special condition for the event to allow an `HTTPClient`? A better approach would be doing `if event.lower() in ("_create", "_add")` to check for generic events that allow subsequent actioning.

interactions/api/models/channel.py
```diff
@@ -144,6 +145,7 @@ class Channel(DictSerializerMixin):
         "member",
         "default_auto_archive_duration",
         "permissions",
+        "client",
```
```suggestion
        "_client",
```

interactions/api/gateway.py
i0bs:
```suggestion
                        data["_client"] = self.http
```

interactions/api/models/channel.pyi
```diff
@@ -26,6 +30,7 @@ class ThreadMember(DictSerializerMixin):
 
 class Channel(DictSerializerMixin):
     _json: dict
+    client: HTTPClient
```
i0bs:
```suggestion
    _client: HTTPClient
```

interactions/api/models/guild.py
```diff
@@ -141,6 +143,7 @@ class Guild(DictSerializerMixin):
     __slots__ = (
         "_json",
         "id",
+        "client",
```
i0bs:
```suggestion
        "_client",
```

interactions/api/models/channel.py
```diff
@@ -164,3 +166,75 @@ def __init__(self, **kwargs):
...
+            content=_content,
+            tts=_tts,
+            # file=file,
+            # attachments=_attachments,
+            embeds=_embeds,
+            allowed_mentions=_allowed_mentions,
+            components=_components,
+        )
+
+        res = await self.client.create_message(channel_id=int(self.id), payload=payload._json)
```
```suggestion
        res = await self._client.create_message(channel_id=int(self.id), payload=payload._json)
```

interactions/api/models/channel.py
```diff
@@ -164,3 +166,75 @@ def __init__(self, **kwargs):
...
+        message = Message(**res, client=self.client)
+        return message
+
+    async def delete(self) -> None:
+        """
+        Deletes the channel.
+        """
+        await self.client.delete_channel(channel_id=int(self.id))
```
```suggestion
        await self._client.delete_channel(channel_id=int(self.id))
```

interactions/api/models/guild.py
```diff
@@ -236,6 +239,144 @@ def __init__(self, **kwargs):
...
+        """
+        Bans a member from the guild
+        :param member_id: The id of the member to ban
+        :type member_id: int
+        :param reason?: The reason of the ban
+        :type reason: Optional[str]
+        :param delete_message_days?: Number of days to delete messages, from 0 to 7. Defaults to 0
+        :type delete_message_days: Optional[int]
+        """
+        await self.client.create_guild_ban(
```
i0bs:
```suggestion
        await self._client.create_guild_ban(
```

interactions/api/models/guild.py
```diff
@@ -236,6 +239,144 @@ def __init__(self, **kwargs):
...
+        reason: Optional[str] = None,
+    ) -> None:
+        """
+        Removes the ban of a user
+        :param user_id: The id of the user to remove the ban from
+        :type user_id: int
+        :param reason?: The reason for the removal of the ban
+        :type reason: Optional[str]
+        """
+        await self.client.remove_guild_ban(
```
i0bs:
```suggestion
        await self._client.remove_guild_ban(
```

interactions/api/models/guild.py
```diff
@@ -236,6 +239,144 @@ def __init__(self, **kwargs):
...
+        reason: Optional[str] = None,
+    ) -> None:
+        """
+        Kicks a member from the guild
+        :param member_id: The id of the member to kick
+        :type member_id: int
+        :param reason?: The reason for the kick
+        :type reason: Optional[str]
+        """
+        await self.client.create_guild_kick(
```
i0bs:
```suggestion
        await self._client.create_guild_kick(
```

interactions/api/models/guild.py
```diff
@@ -236,6 +239,144 @@ def __init__(self, **kwargs):
...
+        :type roles: Union[List[Union[Role, int]], Role, int]
+        :param member_id: The id of the member to add the roles to
+        :type member_id: int
+        :param reason: The reason why the roles are added
+        :type reason: Optional[str]
+        """
+        if isinstance(roles, list):
+            roles = [int(role.id) if isinstance(role, Role) else role for role in roles]
+            for role in roles:
+                await self.client.add_member_role(
```
i0bs:
```suggestion
                await self._client.add_member_role(
```

interactions/api/models/guild.py
```diff
@@ -236,6 +239,144 @@ def __init__(self, **kwargs):
...
+                    guild_id=int(self.id),
+                    user_id=member_id,
+                    role_id=role,
+                    reason=reason,
+                )
+        elif isinstance(roles, Role):
+            await self.client.add_member_role(
```
i0bs:
```suggestion
            await self._client.add_member_role(
```

interactions/api/models/guild.py
```diff
@@ -236,6 +239,144 @@ def __init__(self, **kwargs):
...
+        """
+        await self.client.create_guild_kick(
+            guild_id=int(self.id),
+            user_id=member_id,
+            reason=reason,
+        )
+
+    async def add_member_roles(
+        self,
+        roles: Union[List[Union[Role, int]], Role, int],
```
i0bs: This type hint doesn't respect the current pattern that we have for singulars and multiples. Please consider `Union[List[Role], Role, int]` instead.

interactions/api/models/guild.py
```diff
@@ -236,6 +239,144 @@ def __init__(self, **kwargs):
...
+                )
+        elif isinstance(roles, Role):
+            await self.client.add_member_role(
+                guild_id=int(self.id),
+                user_id=member_id,
+                role_id=int(roles.id),
+                reason=reason,
+            )
+        else:
+            await self.client.add_member_role(
```
i0bs:
```suggestion
            await self._client.add_member_role(
```

interactions/api/models/guild.py
```diff
@@ -236,6 +239,144 @@ def __init__(self, **kwargs):
...
+                guild_id=int(self.id),
+                user_id=member_id,
+                role_id=roles,
+                reason=reason,
+            )
+
+    async def remove_member_roles(
+        self,
+        roles: Union[List[Union[Role, int]], Role, int],
```
i0bs:
```suggestion
        roles: Union[List[Role], Role, int],
```

interactions/api/models/guild.py
```diff
@@ -236,6 +239,144 @@ def __init__(self, **kwargs):
...
+        :type roles: Union[List[Union[Role, int]], Role, int]
+        :param member_id: The id of the member to remove the roles from
+        :type member_id: int
+        :param reason: The reason why the roles are removed
+        :type reason: Optional[str]
+        """
+        if isinstance(roles, list):
+            roles = [int(role.id) if isinstance(role, Role) else role for role in roles]
+            for role in roles:
+                await self.client.remove_member_role(
```
i0bs:
```suggestion
                await self._client.remove_member_role(
```

interactions/api/models/guild.py
```diff
@@ -236,6 +239,144 @@ def __init__(self, **kwargs):
...
+                    guild_id=int(self.id),
+                    user_id=member_id,
+                    role_id=role,
+                    reason=reason,
+                )
+        elif isinstance(roles, Role):
+            await self.client.remove_member_role(
```
i0bs:
```suggestion
            await self._client.remove_member_role(
```

interactions/api/models/guild.py
```diff
@@ -236,6 +239,144 @@ def __init__(self, **kwargs):
...
+                guild_id=int(self.id),
+                user_id=member_id,
+                role_id=int(roles.id),
+                reason=reason,
+            )
+        else:
+            await self.client.remove_member_role(
```
i0bs:
```suggestion
            await self._client.remove_member_role(
```

interactions/api/models/guild.pyi
```diff
@@ -36,6 +37,7 @@ class StageInstance(DictSerializerMixin):
 
 class Guild(DictSerializerMixin):
     _json: dict
+    client: HTTPClient
```
i0bs:
```suggestion
    _client: HTTPClient
```

interactions/api/models/member.py
```diff
@@ -41,6 +41,7 @@ class Member(DictSerializerMixin):
         "permissions",
         "communication_disabled_until",
         "hoisted_role",
+        "client",
```
i0bs:
```suggestion
        "_client",
```

interactions/api/models/member.pyi
```diff
@@ -4,10 +4,12 @@ from typing import Any, List, Optional
 from .misc import DictSerializerMixin
 from .role import Role
 from .user import User
+from ..http import HTTPClient
 
 class Member(DictSerializerMixin):
 
     _json: dict
+    client: HTTPClient
```
```suggestion
    _client: HTTPClient
```

interactions/api/models/message.py
```diff
@@ -226,6 +227,7 @@ class Message(DictSerializerMixin):
         "components",
         "sticker_items",
         "stickers",
+        "client",
```
i0bs:
```suggestion
        "_client",
```

interactions/api/models/message.py
```diff
@@ -278,6 +280,136 @@ def __init__(self, **kwargs):
         )
         self.thread = Channel(**self.thread) if self._json.get("thread") else None
 
+    async def delete(self, reason: Optional[str] = None) -> None:
+        """
+        Deletes the message.
+        :param reason: Optional reason to show up in the audit log. Defaults to `None`.
+        :type reason: Optional[str]
+        """
+        await self.client.delete_message(
```
i0bs:
```suggestion
        await self._client.delete_message(
```

interactions/api/models/message.py
```diff
@@ -278,6 +280,136 @@ def __init__(self, **kwargs):
...
+        payload: Message = Message(
+            content=_content,
+            tts=_tts,
+            # file=file,
+            embeds=_embeds,
+            allowed_mentions=_allowed_mentions,
+            message_reference=_message_reference,
+        )
+
+        await self.client.edit_message(
```
i0bs:
```suggestion
        await self._client.edit_message(
```

interactions/api/models/message.py
```diff
@@ -278,6 +280,136 @@ def __init__(self, **kwargs):
...
+            tts=_tts,
+            # file=file,
+            # attachments=_attachments,
+            embeds=_embeds,
+            message_reference=_message_reference,
+            allowed_mentions=_allowed_mentions,
+            components=_components,
+        )
+
+        res = await self.client.create_message(
```
i0bs:
```suggestion
        res = await self._client.create_message(
```

interactions/api/models/message.pyi
```diff
@@ -51,6 +54,8 @@ class ChannelMention(DictSerializerMixin):
     def __init__(self, **kwargs): ...
 
 class Message(DictSerializerMixin):
+    client: Optional[HTTPClient]
+    http: HTTPClient
```
i0bs: Why do you have two attributes of the same thing with one being optional?

interactions/api/models/role.py
```diff
@@ -53,6 +53,7 @@ class Role(DictSerializerMixin):
         "mentionable",
         "tags",
         "permissions",
+        "client",
```
i0bs:
```suggestion
        "_client",
```

interactions/api/models/role.pyi
```diff
@@ -11,6 +12,7 @@ class RoleTags(DictSerializerMixin):
 
 class Role(DictSerializerMixin):
     _json: dict
+    client: HTTPClient
```
i0bs:
```suggestion
    _client: HTTPClient
```

i0bs (changes requested):
We want to rename the reference of client being passed into our models here as `_client` because this is an internal attribute that does not need to be accessed. We generally apply the underscore character on attributes when we want to distinguish it as a special use case that doesn't necessarily apply to crucial data of the object itself.

There are also some areas where I'm confused about what you're doing.

interactions/api/models/message.pyi
sillyElvy: I was testing something and forgot to undo it, will change

interactions/api/models/guild.py
sillyElvy: `ctx,guild.members`  is only sent in the `GUILD_CREATE` event https://discord.com/developers/docs/resources/guild#guild-object , so I don't see the sense to create a method for it. Also I wanted to implement it directly into the `Member` object, but since it has no `guild_id` attribute I couldn't

sillyElvy:
> We want to rename the reference of client being passed into our models here as `_client` because this is an internal attribute that does not need to be accessed. We generally apply the underscore character on attributes when we want to distinguish it as a special use case that doesn't necessarily apply to crucial data of the object itself.
> 
> There are also some areas where I'm confused about what you're doing.

I agree with the name of `_client`. At first I was thinking about naming it that way, but I followed `context.py` where it was named without underscore.

So why don't you have it as `_client` in `context`?

interactions/api/models/guild.py
i0bs: That doesn't mean it shouldn't be added in only because 1 gateway event has it. During the startup of the bot, any `GUILD_CREATE` events dispatched by the gateway before the client is internally ready, they're cached into the global `Cache` as a K:V in `self_guilds`. This also generally goes against using this paradigm logic:

```py
for member in ctx.guild.members:
    await member.ban()
```

interactions/api/models/guild.py
```diff
@@ -236,6 +241,422 @@ def __init__(self, **kwargs):
...
+        :param reason?: The reason for the kick
+        :type reason: Optional[str]
+        """
+        await self._client.create_guild_kick(
+            guild_id=int(self.id),
+            user_id=member_id,
+            reason=reason,
+        )
+
+    async def add_member_roles(
```
i0bs: You should have this as `add_member_role`, and the `roles` argument in the signature should be `Union[Role, int]`. This does not respect the actual use of the official API endpoint: https://discord.com/developers/docs/resources/guild#remove-guild-member-role

interactions/api/models/guild.py
```diff
@@ -236,6 +241,422 @@ def __init__(self, **kwargs):
...
+            )
+        else:
+            await self._client.add_member_role(
+                guild_id=int(self.id),
+                user_id=member_id,
+                role_id=roles,
+                reason=reason,
+            )
+
+    async def remove_member_roles(
```
i0bs: Same issue as with `add_member_roles`.

interactions/api/models/guild.py
```diff
@@ -236,6 +241,422 @@ def __init__(self, **kwargs):
...
+        )
+        res = await self._client.create_guild_role(
+            guild_id=int(self.id),
+            reason=reason,
+            data=payload._json,
+        )
+        role = Role(**res, _client=self._client)
+        return role
+
+    async def get_guild_member(
```
i0bs: A problem I see with this for consistency purposes is having some methods with `get_guild_X` and others as `get_X` when all endpoints for `Guild` are always in reference to the guild object according to the Discord Developers documentation. This use case is `ctx.guild.get_guild_member()` which I think is unnecessary because calling this off of a guild object is inclusive enough.

I would recommend calling this `get_member` instead.

interactions/api/models/guild.py
```diff
@@ -236,6 +241,422 @@ def __init__(self, **kwargs):
...
+        :type member_id: int
+        """
+        res = await self._client.get_member(
+            guild_id=int(self.id),
+            member_id=member_id,
+        )
+        member = Member(**res, _client=self._client)
+        return member
+
+    async def delete_guild_channel(
```
i0bs: In reference to the debacle of `get_guild_member`, please use `delete_channel`.

interactions/api/models/guild.py
```diff
@@ -236,6 +241,422 @@ def __init__(self, **kwargs):
...
+
+        res = await self._client.modify_guild_role(
+            guild_id=int(self.id),
+            role_id=role_id,
+            data=payload._json,
+            reason=reason,
+        )
+        return Role(**res, _client=self._client)
+
+    async def create_guild_channel(
```
i0bs: Please use `create_channel`.

interactions/api/models/guild.py
```diff
@@ -236,6 +241,422 @@ def __init__(self, **kwargs):
...
+        res = await self._client.create_channel(
+            guild_id=int(self.id),
+            reason=reason,
+            payload=payload._json,
+        )
+
+        channel = Channel(**res, _client=self._client)
+        return channel
+
+    async def modify_guild_channel(
```
i0bs: Please use `modify_channel`.

interactions/api/models/member.py
```diff
@@ -56,3 +59,174 @@ def __init__(self, **kwargs):
...
+            )
+        else:
+            await self._client.add_member_role(
+                guild_id=guild_id,
+                user_id=int(self.user.id),
+                role_id=roles,
+                reason=reason,
+            )
+
+    async def remove_roles(
```
i0bs: In reference to `add_roles`, please use the following suggested changes.

i0bs (changes requested):

FayeDel (approved):

interactions/api/models/channel.py
```diff
@@ -164,3 +166,139 @@ def __init__(self, **kwargs):
...
+        :rtype: Message
+        """
+        from ...models.component import ActionRow, Button, SelectMenu
+        from .message import Message
+
+        _content: str = "" if content is None else content
+        _tts: bool = False if tts is None else tts
+        # _file = None if file is None else file
+        # _attachments = [] if attachments else None
+        _embeds: list = (
```
i0bs: This `_embeds` value declaration falls out of line for most of the conditionals we create for type-based values.

```py
_embeds: list = []
ifisinstance(embeds, list):
    _embeds = [embed._json for embed in embeds]
else:
    _embeds = [embeds._json]

i0bs (changes requested):

i0bs (approved):

PR #390, opened 2021-12-18
(f) fixed the logic of the sub-function for defining parameters for t…

MeidoNoHitsuji:
…he command, as well as the logic of passing parameters in the sub-function

## About

Due to the fact that it is not clear what exactly is in the option, an error occurs. I eliminated it. There was also an error in logic, when options parameters were not entered and this parameter was simply not there.

## Checklist

  - (If existent):
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix

### DIFF
--- a/interactions/api/gateway.py
+++ b/interactions/api/gateway.py
@@ -261,7 +261,7 @@ class WebSocket:
                     for sub_option in option["options"]:
                         kwargs[sub_option["name"]] = sub_option["value"]
             else:
-                kwargs[option['name']] = option['value']
+                kwargs[option["name"]] = option["value"]
 
             return kwargs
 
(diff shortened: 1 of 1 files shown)


i0bs:
Last time I tested your last commit, I took it for granted that it would work but it had an error between dict and object. Does *this one* actually handle that correctly?

MeidoNoHitsuji:
@goverfl0w 
The fact of the matter is that everything worked. But after the changes and other pr'ov, confusion began in the types in options. It did not start because of me, I don’t know why this is happening. Now I have just fixed this problem that has arisen.
Well, yes, I slightly corrected the logic that I forgot about last time (now my current bot is working on the current logic, so everything should be fine).

PR #400, opened 2021-12-22
docs(components.py/Button): label typo

tagptroll1:
## About

This pull request is about a typo, which is fixed.

## Checklist

- [ ] I've ran `pre-commit` to format and lint the change(s) made.
- [X] I've checked to make sure the change(s) work on `3.8.6` and higher.
- [X] This fixes/solves an [Issue](https://github.com/goverfl0w/discord-interactions/issues/399).
  - (If existent): [Issue 399](https://github.com/goverfl0w/discord-interactions/issues/399)
- [ ] I've made this pull request for/as: (check all that apply)
  - [X] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [ ] Bugfix

PR #401, opened 2021-12-23
added new functions, fixed typos

sillyElvy:
## About

This PR adds new 2 new functions:
- Guild.modify_member()
- Member.modify()
- Member.add_to_thread()
- Channel.add_member

Also, it corrects some typos (inspired by #400 but corrected a bit more)
I also already included the correction of #400 so there should be no conflicts.

## Checklist

  - (If existent):
  - [x] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [x] Bugfix

### DIFF
--- a/interactions/api/http.py
+++ b/interactions/api/http.py
@@ -106,7 +106,7 @@ class Padlock:
         self.keep_open = True
 
     def click(self) -> None:
-        """Re-closes the lock after the instiantiation and invocation ends."""
+        """Re-closes the lock after the instantiation and invocation ends."""
         self.keep_open = False
 
     def __enter__(self) -> Any:
@@ -609,7 +609,7 @@ class HTTPClient:
 
     async def get_guild_widget_image(self, guild_id: int, style: Optional[str] = None) -> str:
         """
-        Get a url representing a png image widget for the guild.
+        Get an url representing a png image widget for the guild.
         ..note::
             See _<https://discord.com/developers/docs/resources/guild#get-guild-widget-image> for list of styles.
 
@@ -752,7 +752,7 @@ class HTTPClient:
         self, template_code: str, name: str, icon: Optional[str] = None
     ) -> Guild:
         """
-        Create a a new guild based on a template.
+        Create a new guild based on a template.
 
         ..note::
             This endpoint can only be used by bots in less than 10 guilds.
@@ -1118,7 +1118,7 @@ class HTTPClient:
 
     async def search_guild_members(self, guild_id: int, query: str, limit: int = 1) -> List[Member]:
         """
-        Search a guild for members who's username or nickname starts with provided string.
+        Search a guild for members whose username or nickname starts with provided string.
 
         :param guild_id: Guild ID snowflake.
         :param query: The string to search for
@@ -2333,7 +2333,7 @@ class HTTPClient:
             Route("DELETE", f"/webhooks/{webhook_id}/{webhook_token}/messages/@original")
         )
 
-    # Emoji endpoints, a subset of guild but it should get it's own thing...
+    # Emoji endpoints, a subset of guild but it should get its own thing...
 
     async def get_all_emoji(self, guild_id: int) -> List[Emoji]:
         """
--- a/interactions/api/models/channel.py
+++ b/interactions/api/models/channel.py
@@ -327,4 +327,3 @@ class Thread(Channel):
     """
 
     ...
-
--- a/interactions/api/models/channel.py
+++ b/interactions/api/models/channel.py
@@ -328,7 +328,9 @@ class Channel(DictSerializerMixin):
         :type member_id: int
         """
         if not self.thread_metadata:
-            raise TypeError("The Channel you specified is not a thread!")  # TODO: Move to new error formatter.
+            raise TypeError(
+                "The Channel you specified is not a thread!"
+            )  # TODO: Move to new error formatter.
         await self._client.add_member_to_thread(thread_id=int(self.id), user_id=member_id)
 
 
--- a/interactions/api/models/channel.py
+++ b/interactions/api/models/channel.py
@@ -316,3 +316,17 @@ class Channel(DictSerializerMixin):
             data=payload._json,
         )
         return Channel(**res, _client=self._client)
+
+    async def add_member(
+        self,
+        member_id: int,
+    ) -> None:
+        """
+        This adds a member to the channel, if the channel is a thread
+
+        :param member_id: The id of the member to add to the channel
+        :type member_id: int
+        """
+        if not self.thread_metadata:
+            raise TypeError("The Channel you specified is not a thread!")
+        await self._client.add_member_to_thread(thread_id=int(self.id), user_id=member_id)
--- a/interactions/api/models/channel.py
+++ b/interactions/api/models/channel.py
@@ -317,6 +317,20 @@ class Channel(DictSerializerMixin):
         )
         return Channel(**res, _client=self._client)
 
+    async def add_member(
+        self,
+        member_id: int,
+    ) -> None:
+        """
+        This adds a member to the channel, if the channel is a thread
+
+        :param member_id: The id of the member to add to the channel
+        :type member_id: int
+        """
+        if not self.thread_metadata:
+            raise TypeError("The Channel you specified is not a thread!")
+        await self._client.add_member_to_thread(thread_id=int(self.id), user_id=member_id)
+
 
 class Thread(Channel):
     """An object representing a thread.
--- a/interactions/api/models/channel.py
+++ b/interactions/api/models/channel.py
@@ -328,7 +328,7 @@ class Channel(DictSerializerMixin):
         :type member_id: int
         """
         if not self.thread_metadata:
-            raise TypeError("The Channel you specified is not a thread!")
+            raise TypeError("The Channel you specified is not a thread!")  # TODO: Move to new error formatter.
         await self._client.add_member_to_thread(thread_id=int(self.id), user_id=member_id)
 
 
--- a/interactions/api/models/guild.py
+++ b/interactions/api/models/guild.py
@@ -672,18 +672,19 @@ class Guild(DictSerializerMixin):
 
         :param member_id: The id of the member to modify
         :type member_id: int
-        :param nick: The nickname of the member
+        :param nick?: The nickname of the member `None` removes the nickname
         :type nick: Optional[str]
-        :param roles: A list of all role ids the member has
+        :param roles?: A list of all role ids the member has
         :type roles: Optional[List[int]]
-        :param mute: whether the user is muted in voice channels
+        :param mute?: whether the user is muted in voice channels
         :type mute: Optional[bool]
-        :param deaf: whether the user is deafened in voice channels
-        :param channel_id: id of channel to move user to (if they are connected to voice)
+        :param deaf?: whether the user is deafened in voice channels
+        :type deaf: Optional[bool]
+        :param channel_id?: id of channel to move user to (if they are connected to voice). `None` to disconnect the user from its channel
         :type channel_id: Optional[int]
-        :param communication_disabled_until: when the user's timeout will expire and the user will be able to communicate in the guild again (up to 28 days in the future)
+        :param communication_disabled_until?: when the user's timeout will expire and the user will be able to communicate in the guild again (up to 28 days in the future). `None` removes the timeout
         :type communication_disabled_until: Optional[datetime.isoformat]
-        :param reason: The reason of the modifying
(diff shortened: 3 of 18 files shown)


interactions/api/http.py
```diff
@@ -609,7 +609,7 @@ async def get_guild_widget_settings(self, guild_id: int) -> dict:
 
     async def get_guild_widget_image(self, guild_id: int, style: Optional[str] = None) -> str:
         """
-        Get a url representing a png image widget for the guild.
+        Get an url representing a png image widget for the guild.
```
Toricane, line 612: `a url` is correct

interactions/api/models/misc.py
```diff
@@ -203,7 +203,7 @@ class Format:
     def stylize(cls, format: str, **kwargs) -> str:
         r"""
         This takes a format style from the object and
-        converts it into a useable string for ease.
+        converts it into an useable string for ease.
```
Toricane: `a usable`
Useable *is* correct, but usable is used more, in both American and British English.

Toricane (changes requested):
Here's some requested changes

interactions/api/http.py
sillyElvy: Could you explain me why?

Toricane: It depends on the sound, not vowel, for whether you use an
URL is pronounced yoo-arr-ell, and the y sound makes it a url, not an url

https://www.latimes.com/socal/glendale-news-press/opinion/tn-gnp-a-word-please-theres-a-sound-reason-to-use-a-or-an-20150313-story.html 
here's an article I found that explains it

sillyElvy: I knew it depends on the sound but I didn't know that it's pronounced as "yoo"

Thanks!

interactions/api/models/channel.py
```diff
@@ -317,6 +317,20 @@ async def modify(
...
+        member_id: int,
+    ) -> None:
+        """
+        This adds a member to the channel, if the channel is a thread
+
+        :param member_id: The id of the member to add to the channel
+        :type member_id: int
+        """
+        if not self.thread_metadata:
+            raise TypeError("The Channel you specified is not a thread!")
```
```suggestion
            raise TypeError("The Channel you specified is not a thread!")  # TODO: Move to new error formatter.
```

Adds an internal note for the developers to flag for 2022 roadmap work.

interactions/api/models/guild.py
```diff
@@ -655,6 +655,73 @@ async def modify_channel(
...
+        :type deaf: Optional[bool]
+        :param channel_id?: id of channel to move user to (if they are connected to voice). `None` to disconnect the user from its channel
+        :type channel_id: Optional[int]
+        :param communication_disabled_until?: when the user's timeout will expire and the user will be able to communicate in the guild again (up to 28 days in the future). `None` removes the timeout
+        :type communication_disabled_until: Optional[datetime.isoformat]
+        :param reason?: The reason of the modifying
+        :type reason: Optional[str]
+        """
+        member = Member(**await self._client.get_member(guild_id=int(self.id), member_id=member_id))
+        _nick = member.nick if nick == " " else nick
```
i0bs:
```suggestion
        _nick = member.nick if nick is None else nick
```

Good python practice would be doing `nick=""` in the method when written which strictly tells for the nickname to be removed. `nick=None` is a bad practice.

interactions/api/models/guild.py
```diff
@@ -655,6 +655,73 @@ async def modify_channel(
...
+        _roles = member.roles if not roles else roles
+        _communication_disabled_until = (
+            member.communication_disabled_until
+            if communication_disabled_until == 0
```
i0bs:
```suggestion
            if communication_disabled_until is None
```

The same applies to how `nick` can be handled. Additionally, `communication_disabled_until` is an ISO8601 range and not an integer. Having a optional argument mutability as something not as `NoneType` or a sentinel-like value would be more problematic. We regulate `None` for this.

interactions/api/models/guild.py
```diff
@@ -655,6 +655,73 @@ async def modify_channel(
         )
         return Channel(**res, _client=self._client)
 
+    async def modify_member(
+        self,
+        member_id: int,
+        nick: Optional[str] = " ",  # can't be None because None would remove the nickname
```
i0bs:
```suggestion
        nick: Optional[str] = None,
```

interactions/api/models/guild.py
```diff
@@ -655,6 +655,73 @@ async def modify_channel(
...
+        roles: Optional[List[int]] = None,
+        mute: Optional[bool] = None,
+        deaf: Optional[bool] = None,
+        channel_id: Optional[
+            int
+        ] = 0,  # can't be None because None would kick the member from the channel
```
i0bs:
```suggestion
        ] = None
```

interactions/api/models/guild.py
```diff
@@ -655,6 +655,73 @@ async def modify_channel(
...
+        communication_disabled_until: Optional[
+            datetime.isoformat
+        ] = 0,  # can't be None because None would remove the timeout
```
i0bs:
```suggestion
        ] = None,
```

interactions/api/models/guild.py
```diff
@@ -655,6 +655,73 @@ async def modify_channel(
...
+            else communication_disabled_until
+        )
+
+        payload = Member(
+            nick=_nick,
+            roles=_roles,
+            communication_disabled_until=_communication_disabled_until,
+        )
+
+        if channel_id != 0:
```
i0bs:
```suggestion
        if channel_id:
```

This is a simpler approach to checking if a value is supplied from the argument to add to the payload.

interactions/api/models/member.py
```diff
@@ -212,3 +212,84 @@ async def send(
 
         message = Message(**res, _client=self._client)
         return message
+
+    async def modify(
+        self,
+        guild_id: int,
+        nick: Optional[str] = " ",  # can't be None because None would remove the nickname
```
i0bs:
```suggestion
        nick: Optional[str] = None,
```

interactions/api/models/member.py
```diff
@@ -212,3 +212,84 @@ async def send(
...
+        roles: Optional[List[int]] = None,
+        mute: Optional[bool] = None,
+        deaf: Optional[bool] = None,
+        channel_id: Optional[
+            int
+        ] = 0,  # can't be None because None would kick the member from the channel
```
i0bs:
```suggestion
        ] = None,
```

interactions/api/models/member.py
```diff
@@ -212,3 +212,84 @@ async def send(
...
+        communication_disabled_until: Optional[
+            datetime.isoformat
+        ] = 0,  # can't be None because None would remove the timeout
```
i0bs:
```suggestion
        ] = None,
```

interactions/api/models/member.py
```diff
@@ -212,3 +212,84 @@ async def send(
...
+        reason: Optional[str] = None,
+    ) -> "Member":
```
i0bs, line 226: You can additionally use `self` if you want, but this is fine for now.

interactions/api/models/member.py
```diff
@@ -212,3 +212,84 @@ async def send(
...
+        :param deaf?: whether the user is deafened in voice channels
+        :type deaf: Optional[bool]
+        :param channel_id?: id of channel to move user to (if they are connected to voice). `None` to disconnect the user from its channel
+        :type channel_id: Optional[int]
+        :param communication_disabled_until?: when the user's timeout will expire and the user will be able to communicate in the guild again (up to 28 days in the future). `None` removes the timeout
+        :type communication_disabled_until: Optional[datetime.isoformat]
+        :param reason?: The reason of the modifying
+        :type reason: Optional[str]
+        """
+        _nick = self.nick if nick == " " else nick
```
i0bs:
```suggestion
        _nick = self.nick if nick is None else nick
```

interactions/api/models/member.py
```diff
@@ -212,3 +212,84 @@ async def send(
...
+        _roles = self.roles if not roles else roles
+        _communication_disabled_until = (
+            self.communication_disabled_until
+            if communication_disabled_until == 0
```
i0bs:
```suggestion
            if communication_disabled_until is None
```

interactions/api/models/member.py
```diff
@@ -212,3 +212,84 @@ async def send(
...
+            else communication_disabled_until
+        )
+
+        payload = Member(
+            nick=_nick,
+            roles=_roles,
+            communication_disabled_until=_communication_disabled_until,
+        )
+
+        if channel_id != 0:
```
i0bs:
```suggestion
        if channel_id:
```

interactions/api/models/member.pyi
```diff
@@ -57,3 +57,23 @@ class Member(DictSerializerMixin):
         embeds=None,
         allowed_mentions=None,
     ) -> Message: ...
+
+    async def modify(
+        self,
+        guild_id: int,
+        nick: Optional[str] = " ",
```
```suggestion
        nick: Optional[str] = None,
```

interactions/api/models/member.pyi
```diff
@@ -57,3 +57,23 @@ class Member(DictSerializerMixin):
...
+        roles: Optional[List[int]] = None,
+        mute: Optional[bool] = None,
+        deaf: Optional[bool] = None,
+        channel_id: Optional[
+            int
+        ] = 0,
```
```suggestion
        ] = None,
```

interactions/api/models/member.pyi
```diff
@@ -57,3 +57,23 @@ class Member(DictSerializerMixin):
...
+        communication_disabled_until: Optional[
+            datetime.isoformat
+        ] = 0,
```
```suggestion
        ] = None,
```

interactions/api/models/message.py
```diff
@@ -189,7 +189,7 @@ class Message(DictSerializerMixin):
     :ivar Optional[MessageInteraction] interaction?: Message interaction object, if the message is sent by an interaction.
     :ivar Optional[Channel] thread:? The thread that started from this message, if any, with a thread member object embedded.
     :ivar Optional[Union[Component, List[Component]]] components?: Components associated with this message, if any.
-    :ivar Optional[List[PartialSticker"]] sticker_items?: An array of message sticker item objects, if sent with them.
+    :ivar Optional[List["PartialSticker"]] sticker_items?: An array of message sticker item objects, if sent with them.
```
i0bs:
```suggestion
    :ivar Optional[List[PartialSticker]] sticker_items?: An array of message sticker item objects, if sent with them.
```

This was probably a mistake when I was doing the documentation. I took a lazy approach to copy-pasting the interfacing which introduced the quotations.

interactions/api/models/message.py
```diff
@@ -189,7 +189,7 @@ class Message(DictSerializerMixin):
     :ivar Optional[MessageInteraction] interaction?: Message interaction object, if the message is sent by an interaction.
     :ivar Optional[Channel] thread:? The thread that started from this message, if any, with a thread member object embedded.
```
i0bs:
```suggestion
    :ivar Optional[Channel] thread?: The thread that started from this message, if any, with a thread member object embedded.
```

interactions/api/models/misc.py
```diff
@@ -153,7 +153,7 @@ def epoch(self) -> float:
     @property
     def timestamp(self) -> datetime.datetime:
         """
-        The Datetime object variation of the the "Timestamp" field of the snowflake.
+        The Datetime object variation of the "Timestamp" field of the snowflake.
```
i0bs:
```suggestion
        The Datetime object variation of the Timestamp field of the snowflake.
```

i0bs (changes requested):
From what I've seen in other servers, there's a little bit of a misconception to how `None` checking optional arguments works. We always want to avoid using mutable default types where possible when it can stay complacent with how values are entered through conditionals. The none return has been used as a way to determine the equivalent of a "missing" argument.

We like to avoid doing `= " "`, `= 0` and alike where we're working directly to the API and isn't a client-facing aspect where we expect a core integral part to *need* catering. Our decorators somewhat break this but not really at the same time because we still use conditionals to check. This post goes into better detail on why we don't do this: https://nikos7am.com/posts/mutable-default-arguments/

Some of our code does stuff like this instead:

```py
# meth 1
def foo(bar: typing.Optional[str] = None):
    _bar = "" if bar is None else bar

# meth 2
def foo(bar: typing.Optional[str] = ""):
    _bar = bar if not bool(bar) else bar
```

For consistency reasons, we preferably want method 1.

i0bs:
There's still a merge conflict

sillyElvy:
> There's still a merge conflict

Fixed

i0bs (approved):

PR #418, opened 2022-01-05
feat: implement new functions

sillyElvy:
## About

This PR:
- Adds new functions:
  - Channel:
    - pin_message()
    - unpin_message()
    - publish_message()
    - get_pinned_messages()
  - Message:
    - pin()
    - unpin()
    - publish()
  - Guild:
    - get_all_channels()
    - get_all_roles()
    - modify_role_position()
    - get_bans()
  - Role:
    - modify_position()

- Adds missing `return` and `rtype` to docstrings 

- Allows component sending in `Member.send()`

- fixes component sending

## Checklist

  - (If existent):
  - [x] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [x] Bugfix


interactions/api/models/channel.py
```diff
@@ -332,6 +332,61 @@ async def add_member(
...
+
+    async def publish_message(
+        self,
+        message_id: int,
+    ):
+        """Publishes (API calls it crossposts) a message in the channel to any that is followed by.
+
+        :param message_id: The id of the message to publish
+        :type message_id: int
+        :return: message object
```
```suggestion
        :return: The message published
```

interactions/api/models/channel.py
```diff
@@ -332,6 +332,61 @@ async def add_member(
...
+        :rtype: Message
+        """
+        from .message import Message
+
+        res = await self._client.publish_message(
+            channel_id=int(self.id), message_id=int(message_id)
+        )
+        return Message(**res, _client=self._client)
+
+    async def get_pinned_messages(self):
```
i0bs, line 471:
```suggestion
    async def get_pinned_messages(self) -> List[Message]:
```

interactions/api/models/channel.py
```diff
@@ -332,6 +332,61 @@ async def add_member(
...
+        """
+        Get all pinned messages from the channel.
+        :return: A list of pinned message objects.
```
i0bs, line 474:
```suggestion
        :return: A list of pinned message objects.
        :rtype: List[Message]
```

interactions/api/models/guild.py
```diff
@@ -1031,6 +1042,65 @@ async def delete_scheduled_event(self, event_id: int) -> None:
...
+        """
+        res = self._client.get_all_channels(int(self.id))
+        channels = [Channel(**channel, _client=self._client) for channel in res]
+        return channels
+
+    async def get_all_roles(self) -> List[Role]:
+        """
+        Gets all roles of the guild as list
+
+        :rtype: List[Role]
```
i0bs, line 1061:
```suggestion
        :return: The roles of the guild.
        :rtype: List[Role]
```

interactions/api/models/guild.py
```diff
@@ -1031,6 +1042,65 @@ async def delete_scheduled_event(self, event_id: int) -> None:
             guild_scheduled_event_id=Snowflake(event_id),
         )
 
+    async def get_all_channels(self) -> List[Channel]:
+        """
+        Gets all channels of the guild as list
+
+        :rtype: List[Channel]
```
i0bs, line 1050:
```suggestion
        :return: The channels of the guild.
        :rtype: List[Channel]
```

interactions/api/models/guild.py
```diff
@@ -1031,6 +1042,65 @@ async def delete_scheduled_event(self, event_id: int) -> None:
...
+        Gets a list of banned users
+
+        :return: List of banned users with reasons
+        :rtype: List[dict]
+        """
+
+        res = await self._client.get_guild_bans(int(self.id))
+        for ban in res:
+            ban["user"] = User(**ban["user"])
+        return res
```
i0bs, line 1104:
```suggestion
        return ban["user"]
```

This should be returning them in the converted object.

interactions/api/models/guild.py
```diff
@@ -1031,6 +1042,65 @@ async def delete_scheduled_event(self, event_id: int) -> None:
...
+        )
+        roles = [Role(**role, _client=self._client) for role in res]
+        return roles
+
+    async def get_bans(self) -> List[dict]:
+        """
+        Gets a list of banned users
+
+        :return: List of banned users with reasons
+        :rtype: List[dict]
```
i0bs, line 1098:
```suggestion
        :rtype: List[User]
```

interactions/api/models/guild.py
```diff
@@ -1031,6 +1042,65 @@ async def delete_scheduled_event(self, event_id: int) -> None:
...
+        """
+
+        _role_id = role_id.id if isinstance(role_id, Role) else role_id
+        res = await self._client.modify_guild_role_position(
+            guild_id=int(self.id), position=position, role_id=_role_id, reason=reason
+        )
+        roles = [Role(**role, _client=self._client) for role in res]
+        return roles
+
+    async def get_bans(self) -> List[dict]:
```
i0bs, line 1093:
```suggestion
    async def get_bans(self) -> List[User]:
```

interactions/api/models/guild.pyi
```diff
@@ -265,6 +265,15 @@ class Guild(DictSerializerMixin):
...
     ) -> None: ...
+    async def get_all_channels(self) -> List[Channel]: ...
+    async def get_all_roles(self) -> List[Role]: ...
+    async def modify_role_position(
+        self,
+        role_id: Union[Role, int],
+        position: int,
+        reason: Optional[str] = None,
+    ) -> List[Role]: ...
+    async def get_bans(self) -> List[dict]: ...
```
i0bs, line 276:
```suggestion
    async def get_bans(self) -> List[User]: ...
```

interactions/api/models/message.py
```diff
@@ -423,6 +427,26 @@ async def reply(
...
         return Message(**res, _client=self._client)
 
+    async def pin(self) -> None:
+        """Pins the message to its channel"""
+
+        await self._client.pin_message(channel_id=int(self.channel_id), message_id=int(self.id))
+
+    async def unpin(self) -> None:
+        """Unpins the message from its channel"""
+        await self._client.unpin_message(channel_id=int(self.channel_id), message_id=int(self.id))
```
i0bs, line 438:
```suggestion

        await self._client.unpin_message(channel_id=int(self.channel_id), message_id=int(self.id))
```

i0bs (changes requested):
I also see that you're importing specific models per iteration in a method, which can affect performance. (although it should be a very tiny amount because it'll recall upon loaded memory) Is there any reason why they're being done in the methods instead of an import in the file itself?

interactions/api/models/guild.py
sillyElvy: I don't understand why I should change that

sillyElvy:
> I also see that you're importing specific models per iteration in a method, which can affect performance. (although it should be a very tiny amount because it'll recall upon loaded memory) Is there any reason why they're being done in the methods instead of an import in the file itself?

Circular import

interactions/api/models/guild.py
i0bs: Every other helper method you have returns a converted model object, but you don't do it for this one. It's silly to keep it inconsistent.

sillyElvy: it's a 
`{reason: user}`
dict
so why should I not include the reasons for the ban?

i0bs: Why not return a map then?

sillyElvy: how would I do that? I never worked with map

i0bs: Map is just another way to say dictionary, it's a different term for other programming languages. Just return a dictionary where the key and value are mapped respectively with a User object to reason relation.

sillyElvy: ok, how would I typehint that?

sillyElvy: I just checked again
It's
`[{"user": user, "reason": reason}]`
actually

FayeDel (approved):
Tested selects on this particular PR, and got no errors on the boot of this instance, and of sending selects. Looks good to me :)

PR #422, opened 2022-01-11
Fix/autocomplete doc

pwuertz:
## About

Small fix for the code example in `Client.autocomplete`.

The current example doesn't really work, since the decorated handler is called with the current user input string as first argument if the input is not empty. Not accepting this positional arg causes some call mismatch in the event handling stack.

## Checklist

  - (If existent):
  - [x] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [ ] Bugfix

### DIFF
--- a/interactions/client.py
+++ b/interactions/client.py
@@ -414,7 +414,7 @@ class Client:
         .. code-block:: python
 
             @autocomplete("option_name")
-            async def autocomplete_choice_list(ctx):
+            async def autocomplete_choice_list(ctx, user_input: str = ""):
                 await ctx.populate([...])
 
         :param name: The name of the option to autocomplete.


i0bs:
@DeltaXWizard look at pr pls 👀

FayeDel (approved):
Hiya!

This looks good to me, thanks for contributing :)

pwuertz:
I just realized that in the `unstable` branch `@autocomplete("option_name")` isn't sufficient anymore. An `ApplicationCommand` parameter is required now. You happen to know where to get one of these from? I tried plugging in the command created with `@bot.command`, but that doesn't fulfill the type.

sillyElvy:
I believe it's cached

i0bs:
`ApplicationCommand` per known command from the API (and decorators) are cached under the `Client().cache.interactions` approach. You can use `.view()` in order to see what commands are there, the K:V relation is name to the command object in question.

Additionally, you can supplement `command` as an `int` type for the command ID instead, you're not forced to use the object.

pwuertz:
Right, but if ApplicationCommand (and its ID) is a run-time variable now, the compile-time decorator approach doesn't seem to fit anymore, does it? You'd have to wait for the client to be connected, having synced the application command, and only then you'd be able to give an ID to `@autocomplete`. Also, I don't know the details here, but won't you get multiple IDs when registering a single command at guild scope for each guild?

I assume the general idea of the new command parameter is to tie an `@autocomplete` handler to a `@bot.command` definition, is it not?

i0bs:
The commands are synced before the internal ready state of the client activates the websocket connection for login. `@autocomplete` is a separate decorator to the one for commands

pwuertz:
Wait, you mean `@bot.command` actually executes async network activity at decoration-time?

i0bs:
Yes, the decorator not only declares a sync check with the API, it acts as a callback method during dispatch

pwuertz:
Any reason for not using `autocomplete_{command_name}_{parameter_name}` for routing autocomplete requests instead of the runtime registration id?

This would allow binding autocomplete handlers to command handlers and remove the dependency on runtime information:
```python
@bot.command(...)
async def my_cmd(...):
   ...

@my_cmd.options["my_opt"].autocomplete
async def _(...):
   ...
```

i0bs:
If you want the honest truth: no clue. Our decorators work as a constructor and a callback when working with the Gateway and Web API. We don't use them as `to_dict`'s as other libraries do, so the design varies somewhat.

PR #423, opened 2022-01-11
fix(gateway,component): SelectMenu parse per SelectOption

MeidoNoHitsuji:
## About

fix SelectMenu
add values in message_component

  - (If existent):
  - [ ] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [x] Bugfix


i0bs:
@DeltaXWizard look at pr pls 👀

interactions/api/gateway.py
```diff
@@ -320,6 +320,8 @@ def check_sub_auto(option: dict) -> tuple:
                                     )
                     elif data["type"] == InteractionType.MESSAGE_COMPONENT:
                         _name = f"component_{context.data.custom_id}"
+                        if context.data._json.get("values"):
+                            _args.append(context.data.values)
```
FayeDel, line 318: The PR looks good to me,
but I want to know the reason behind the `add values in message component` detail, over simply just using context object as how it was in v3's callbacks

FayeDel (changes requested):

interactions/api/gateway.py
MeidoNoHitsuji: Because in the callback function it will be much easier to immediately receive the selected value than to request it from nowhere (I, for example, did not immediately understand that the value can be obtained from the context).

sillyElvy (changes requested):
You got a merge conflict

PR #428, opened 2022-01-13
feat: create basic functionality for Extensions

Catalyst4222:
Expands Extension.__new__ and adds new functions to mimic client decorators

## About

This pr adds in basic functionality for the Extension class, along with decorators for commands/components/listeners
Marked as a draft because some things are missing, like properly removing commands on removal of the Extension
Any suggestions on improvements are welcome

## Checklist

  - (If existent): [[MISC] Revamping the cogs/extension framework](https://github.com/goverfl0w/discord-interactions/issues/425)
  - [ ] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix


Catalyst4222:
I would like someone to double check that everything works, and I'm not just hallucinating a lack of errors. It would probably be a good idea to simplify some of the code, I got it working but it doesn't look the nicest

i0bs:
Delta and I will test this code when this PR is marked ready for reviewing

interactions/client.py
```diff
@@ -1,11 +1,13 @@
+import functools
```
i0bs:
```suggestion
from functools import wraps
```

interactions/client.py
```diff
@@ -722,25 +763,117 @@ def __init__(self, client):
...
+                self.client.loop.create_task(
+                    self.client.http.delete_application_command(
+                        cmd["application_id"], cmd["id"], cmd["guild_id"]
+                    )
+                )
+                for cmd in cmds
+            ]
+
+
+@functools.wraps(command)
```
i0bs:
```suggestion
@wraps(command)
```

interactions/client.py
```diff
@@ -722,25 +763,117 @@ def __init__(self, client):
...
+def extension_listener(name=None):
+    def decorator(func):
+        func.__listener_name__ = name or func.__name__
+
+        return func
+
+    return decorator
+
+
+# @functools.wraps(Client.component)
```
i0bs:
```suggestion
@wraps(Client.component)
```

Please either uncomment this or remove it if you do not plan on using it.

interactions/client.py
```diff
@@ -1,11 +1,13 @@
+import functools
+import inspect
```
i0bs:
```suggestion
from inspect import getmembers
```

interactions/client.py
```diff
@@ -617,16 +639,34 @@ def remove(self, name: str, package: Optional[str] = None) -> None:
...
             log.error(f"Extension {name} has not been loaded before. Skipping.")
+            return
+
+        try:
+            extension.teardown()  # made for Extension, usable by others
+        except AttributeError:
+            pass
+
+        if isinstance(extension, types.ModuleType):  # loaded as a module
+            for ext_name, ext in inspect.getmembers(
```
i0bs:
```suggestion
            for ext_name, ext in getmembers(
```

interactions/client.py
```diff
@@ -722,25 +763,117 @@ def __init__(self, client):
...
+        self = super().__new__(cls)
+
+        self.client = client
+        self._commands = {}
+        self._listeners = {}
+        self._components = {}
+
+        # This gets every coroutine in a way that we can easily change them
+        # cls
+        for name, func in inspect.getmembers(self, predicate=iscoroutinefunction):
```
i0bs:
```suggestion
        for name, func in getmembers(self, predicate=iscoroutinefunction):
```

interactions/client.py
```diff
@@ -1,11 +1,13 @@
+import functools
+import inspect
 import sys
-from asyncio import get_event_loop
+import types
```
i0bs:
```suggestion
from types import ModuleType
```

interactions/client.py
```diff
@@ -617,16 +639,34 @@ def remove(self, name: str, package: Optional[str] = None) -> None:
...
+        if _name not in self.extensions:
             log.error(f"Extension {name} has not been loaded before. Skipping.")
+            return
+
+        try:
+            extension.teardown()  # made for Extension, usable by others
+        except AttributeError:
+            pass
+
+        if isinstance(extension, types.ModuleType):  # loaded as a module
```
i0bs:
```suggestion
        if isinstance(extension, ModuleType):
```

interactions/client.pyi
```diff
@@ -61,7 +61,24 @@ class Client:
     async def raw_channel_create(self, message) -> dict: ...
     async def raw_message_create(self, message) -> dict: ...
     async def raw_guild_create(self, guild) -> dict: ...
+    def add_extension(self, extension: "Extension"): ...
```
i0bs:
```suggestion
```
This isn't used at all anywhere else in the script (from where I can tell)

interactions/client.py
```diff
@@ -722,25 +763,117 @@ def __init__(self, client):
...
+
+        for component, funcs in self._components.items():
+            for func in funcs:
+                self.client.websocket.dispatch.events[component].remove(func)
+
+        for cmd, funcs in self._commands.items():
+            for func in funcs:
+                self.client.websocket.dispatch.events[cmd].remove(func)
+
+        clean_cmd_names = ["_".join(cmd.split("_")[1:]) for cmd in self._commands.keys()]
```
i0bs: `["_".join(cmd.split("_")[1:])`

This is a messy operation that I highly dislike. Why is this being done? Let's say we have a command named as `hello-world`, this won't matter any. But if it uses snake_case like `long_function_name` then you're running the same thing as without the wildcard. You could just do `cmd[7:]` since you're encompassing the length of `command_` here.

interactions/client.py
```diff
@@ -722,25 +763,117 @@ def __init__(self, client):
...
+        cmds = filter(
+            lambda x: x["name"] in clean_cmd_names,
```
i0bs:
```suggestion
            lambda cmd: cmd["name"] in clean_cmd_names,
```

Try and step away from alphabet soup abstraction, `x` may make sense from a maintainer's perspective but anyone else reading this can easily get confused by common iterator naming conventions

i0bs (changes requested):

FayeDel:
There's support for `@client.component`, `@Client.command`, etc using these, but what about `@Client.autocomplete`?

i0bs:
No support for modals either

interactions/client.py
```diff
@@ -617,34 +649,61 @@ def remove(self, name: str, package: Optional[str] = None) -> None:
...
-        if module is None:
+        if extension is None:
             log.warning(f"Extension {name} could not be reloaded because it was never loaded.")
-            self.extend(name, package)
+            self.load(name, package)
+            return
 
         self.remove(name, package)
-        self.load(name, package)
+        return self.load(name, package)
```
mAxYoLo01:
```suggestion
        return self.load(name, package, *args, **kwargs)
```

interactions/client.py
```diff
@@ -617,34 +649,61 @@ def remove(self, name: str, package: Optional[str] = None) -> None:
...
         :param name: The name of the extension.
         :type name: str
         :param package?: The package of the extension.
         :type package: Optional[str]
+        :param \*args?: Optional arguments to pass to the extension
+        :type \**args: tuple
+        :param \**kwargs?: Optional keyword-only arguments to pass to the extension.
+        :type \**kwargs: dict
+        :return: The reloaded extension.
+        :rtype: Optional[interactions.client.Extension]
```
i0bs:
```suggestion
        :rtype: Optional[Extension]
```

interactions/client.py
```diff
@@ -581,32 +601,44 @@ def decorator(coro: Coroutine) -> Any:
...
         :param name: The name of the extension.
         :type name: str
         :param package?: The package of the extension.
         :type package: Optional[str]
+        :param \*args?: Optional arguments to pass to the extension
+        :type \**args: tuple
+        :param \**kwargs?: Optional keyword-only arguments to pass to the extension.
+        :type \**kwargs: dict
+        :return: The loaded extension.
+        :rtype: Optional[interactions.client.Extension]
```
i0bs:
```suggestion
        :rtype: Optional[Extension]
```

i0bs (changes requested):

i0bs (approved):
LGTM. Please pre-commit though.

interactions/client.py
```diff
@@ -736,25 +798,178 @@ def __init__(self, client):
...
+                func = client.modal(*args, **kwargs)(func)
+
+                modal = kwargs.get("modal") or args[0]
+                modal_name = f"modal_{modal.custom_id}"
+
+                listeners = self._listeners.get(modal_name, [])
+                listeners.append(func)
+                self._listeners[modal_name] = listeners
+
+        client.extensions[cls.__name__] = self
```
mAxYoLo01:
```suggestion
        client._extensions[cls.__name__] = self
```
With this small fix I was able to create a working basic cog command

i0bs:
This is LGTM, I think we're finally ready to go ahead and merge this. I'll be leaving the last approval on this for @DeltaXWizard , but all of these changes look good on my end.

FayeDel:
You may need to run pre-commit again after new pending changes, I'll test afterwards though :)
(Code looks good, I have no objections, I just need to test the thing before I approve)

interactions/client.py
```diff
@@ -595,32 +618,44 @@ def decorator(coro: Coroutine) -> Any:
...
+        :type \**args: tuple
+        :param \**kwargs?: Optional keyword-only arguments to pass to the extension.
+        :type \**kwargs: dict
+        :return: The loaded extension.
+        :rtype: Optional[Extension]
         """
         _name: str = resolve_name(name, package)
 
         if _name in self._extensions:
             log.error(f"Extension {name} has already been loaded. Skipping.")
```
mAxYoLo01, line 654:
```suggestion
            log.error(f"Extension {name} has already been loaded. Skipping.")
            return
```

mAxYoLo01:
After some more tests, I figured out that there still are a few problems to address:
1. When loading any extension **after** `bot.start()`, all commands that are initialized outside of a cog (for example initialized in the main launch file) will all disappear from the Slash Commands Menu. Commands of the loaded cog will not be removed from the Menu but will not respond either.
2. After unloading an extension, the commands from this extension are not removed, and can still be accessed from the Slash Commands Menu.

Feel free to ask me if you need some example code to recreate these issues.

Toricane:
I don't think you are supposed to load extensions after `bot.start()`
But it does seem like an issue, shouldn't be wiping the ones in the main file

Catalyst4222:
I want to test again, just to make sure I fixed the breaking changes from the last patch

mAxYoLo01:
> I don't think you are supposed to load extensions after `bot.start()` But it does seem like an issue, shouldn't be wiping the ones in the main file

Didn't know it might be a possibility that it shouldn't be loaded afterwards, in fact I was creating a Slash Command that allows the user to un/re/load cogs directly via Discord. I've previously done it with d.py, so I thought that it would be a feature that i.py should also have.

sillyElvy:
Afaik fl0w wanted something like hot reload for cogs at first so it should be possible tho shouldn't it?

Toricane:
> Afaik fl0w wanted something like hot reload for cogs at first so it should be possible tho shouldn't it?

Oh ok never mind then

interactions/client.py
```diff
@@ -631,34 +666,61 @@ def remove(self, name: str, package: Optional[str] = None) -> None:
...
+        if isinstance(extension, ModuleType):  # loaded as a module
+            for ext_name, ext in getmembers(
+                extension, lambda x: isinstance(x, type) and issubclass(x, Extension)
+            ):
+                self.remove(ext_name)
+
+            del sys.modules[_name]
 
-        log.debug(f"Removed extension {name}.")
-        del sys.modules[_name]
```
i0bs, line 656: For what it's worth, apparently this is no longer deleting the actual module from the system. So, this is why your extensions are still working after they've been removed. Deleting it from the list of extensions stored in the Client is not a viable thing.

interactions/client.py
```diff
@@ -631,34 +666,61 @@ def remove(self, name: str, package: Optional[str] = None) -> None:
...
         del self._extensions[_name]
```
i0bs, line 706:
```suggestion
        del sys.modules[_name]
        del self._extensions[_name]
```

i0bs (changes requested):

interactions/client.py
Catalyst4222: That causes an error when unloading something that isn't a module, e.x. an Extension

Catalyst4222:
I don't know how to do pre-commit
I do `git add <changes>`, `black.exe .`, `isort.exe .`, `flake8.exe .` and do the changes, but it *still* failes

Catalyst4222:
image

mAxYoLo01:
Alright there is still the problem that "global" commands (not initialized in cogs) disappear after loading any cog on my side. The other issues have been fixed though.

Catalyst4222:
> "global" commands (not initialized in cogs) disappear after loading any cog on my side

I'm unable to reproduce this. All commands outside of the cog remain usable, and any inside get added/removed as the cog gets loaded/unloaded. The only reason I can think of for it happening would be global commands, as I've only been testing in the guild scope. I'll try global commands now, but I don't expect a difference

mAxYoLo01:
> The only reason I can think of for it happening would be global commands, as I've only been testing in the guild scope. 

Oh my bad on this one, I formulated badly my previous message. In fact, I also have tried it only with a guild scope. 
Weird that we don't have the same behaviour, on my hand whenever I load/reload a cog, all the commands outside of this cog disappear.

Catalyst4222:
I've recreated the missing commands issue, and it seems to only happen when a cog is unloaded then reloaded in quick succession. Any ideas as to why or where this is happening is welcome

i0bs:
I'll be moving through with merging this to `unstable`, I'm looking to hit a feature lock on getting extensions in rather soon and we can always make revisions with the pull request being brought in. For any relating concerns to this, I advise others to create an [Issue](https://github.com/interactions-py/library/issues) or make a mention of it in the #unstable-dev channel in our support server.

Issue #435, opened 2022-01-15
[BUG] 404 when using `ComponentContext.edit()`


Toricane:
**Describe the bug**
I get `interactions.api.error.HTTPException: Resource does not exist. Error code: 404` whenever I try `await ctx.edit(...)` with `ComponentContext`.
The interaction loads for 3 seconds, and then fails and errors at the same time.

However, this seems to only affect `ComponentContext`; `CommandContext` works fine and is able to edit the message properly without any errors.

**Reproducing steps**
These are the steps I took in order to produce this bug, which should be able to be reproduced for everyone else as well.

0. Use `unstable` branch
1. Use this code:
```py
@bot.component("test")
async def a_test_component(ctx: ComponentContext):
    await ctx.edit(content="lol")
```
2. Make a command that sends a button with `custom_id="test"` or whatever you want as long as the one in the component callback decorator is the same
3. Run the bot
4. Use the aforementioned command
5. Click on the button that the bot sent
6. Wait for 3 seconds
7. The interaction fails, and the error shows up in the console

**What's normally expected**
When I edit the message using `await ctx.edit(...)`, the message should edit properly without any issues or errors

**What actually happened**
Instead, I received this traceback error given from my Python terminal:

```py
Task exception was never retrieved
future: <Task finished name='Task-14' coro=<a_test_component() done, defined at c:\Users\hi\Desktop\Coding\Python\paginator\dinteractions-Paginator\bot.py:47> exception=HTTPException('Resource does not exist. Error code: 404')>
Traceback (most recent call last):
  File "c:\Users\hi\Desktop\Coding\Python\paginator\dinteractions-Paginator\bot.py", line 49, in a_test_component
    await ctx.edit(content="lol")
  File "C:\Users\hi\AppData\Local\Programs\Python\Python39\lib\site-packages\interactions\context.py", line 552, in edit
    await func()
  File "C:\Users\hi\AppData\Local\Programs\Python\Python39\lib\site-packages\interactions\context.py", line 539, in func
    res = await self.client.edit_interaction_response(
  File "C:\Users\hi\AppData\Local\Programs\Python\Python39\lib\site-packages\interactions\api\http.py", line 2104, in edit_interaction_response
    return await self._req.request(
  File "C:\Users\hi\AppData\Local\Programs\Python\Python39\lib\site-packages\interactions\api\http.py", line 214, in request
    raise HTTPException(response.status)
interactions.api.error.HTTPException: Resource does not exist. Error code: 404
[...]

PR #436, opened 2022-01-15
fix!: allow ctx.edit to edit component responses when not deferred

sillyElvy:
This PR allows messages to be edited as a component response using ctx.edit
It also prevents message contents from being removed when editing the message
## Checklist

  - (If existent): #435 
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix

### DIFF
--- a/interactions/context.py
+++ b/interactions/context.py
@@ -184,7 +184,7 @@ class CommandContext(Context):
         """
         if content is None and self.message and self.callback == InteractionCallbackType.DEFERRED_UPDATE_MESSAGE:
             _content = self.message.content
-		else:
+        else:
             _content: str = "" if content is None else content
         _tts: bool = False if tts is None else tts
         # _file = None if file is None else file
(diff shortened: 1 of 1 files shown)


interactions/context.py
```diff
@@ -182,15 +182,24 @@ async def send(
         :return: The sent message as an object.
         :rtype: Message
         """
+        if content is None and self.message:
+            _content = self.message.content
```
i0bs, line 190: Why do you have a separate conditional here on top of the ternary operation?

sillyElvy: Because (iirc) I made it so that can use `send` when you defer components to edit the original message

i0bs: `_content` will be overwritten using this logic. Assuming that `message` is an attribute with a value, the second declaration of the variable in this method will simply reset the value of it. The ternary fails to consider the latter here, turning this into a lossy solution.

i0bs:
Due to write access inheritance on forks of this project as the owner/maintainer, I was able to make commits fixing this. I will be merging this now.

PR #450, opened 2022-01-24
fix: create new event loop if none exists on Python 3.10

Jalancar:
## About

This pull request is about a bug in http.py, which does not create an event loop in Python 3.10 if one isn't running.

## Checklist

  - (If existent):
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix

### DIFF
--- a/interactions/api/http.py
+++ b/interactions/api/http.py
@@ -1,6 +1,6 @@
 import asyncio
 import traceback
-from asyncio import AbstractEventLoop, Lock, get_event_loop, get_running_loop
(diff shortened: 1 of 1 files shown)


Jalancar:
This works with disable_sync=True, but I just re-enabled sync and the bot just hangs at some point after creating the loop. Investigating

i0bs:
FYI, these changes were introduced to the `unstable` branch, hence the closing status.

PR #451, opened 2022-01-25
refactor: make the Gateway optimised in connection and dispatch.

### LINKED ISSUE #452: [REQUEST] A better Gateway connection.
**What is the feature you're proposing?**
Due to issues tying in with #430 , it's known that a better implementation of a connection with the Gateway has been long due. We are currently working on refactoring the main `WebSocket` class which will consolidate `Heartbeat` and the former into `WebSocketClient`. This will act as a user-facing solution to the main websocket connection to the Gateway, as well as internalising the heartbeater.

These are the current goals we're hoping to achieve with the refactoring:

- A daemonic heartbeater process that respects the event loop.
- Safeguards against general zombified states and dropped connections.
- Gracefully connecting and disconnecting the client from the Gateway.
- Race conditions that do not cause sporadic disconnection or indefinite looping of the heartbeat.

**Additional Information**
- [ ] My feature request is related to an existing Issue.
    - Issue (if referenceable):

i0bs:
## About

This pull request is about refactoring the main gateway script, which does all of the heavy lifting for connecting to the gateway and establishing a websocket connection as the client. This refactor will be able to help us make reconnection logic much more simple as the main prominent issue is between the heartbeater being a non-daemonic process. (`threading.Thread` vs `asyncio.Event` here)

## Checklist

- [X] This fixes/solves an [Issue](https://github.com/goverfl0w/discord-interactions/issues).
  - (If existent): #452 
- [X] I've made this pull request for/as: (check all that apply)
  - [X] Documentation
  - [x] Breaking change
  - [X] New feature/enhancement
  - [ ] Bugfix


dxf:
Commenting for pings - happy to beta test this when it's usable, as I've been waiting for this to save my bot's connection problems :)

CBradell:
I can test as well; I'd be happy to pull down an unstable version once it's been merged.

CBradell:
Downloaded the most recent push and am testing. Will let you know!

i0bs:
FYI, this is still very unstable and some things might not work as expected. @DeltaXWizard 's made great progress on the reconnection logic atm (kudos to them) but I'm not sure whether or not it's completely stable for the most part.

FayeDel:
Just reiterating from the commit change, don't expect it to work (It's "less unstable" in terms of when it stops as it stops a few steps later than the current stable, but it is still very unstable, it just can't restart the session but everything else in methodology should be proper?)

CBradell:
Gotcha, that makes sense. It looks like the heartbeat fails sooner now. I'll wait for the next update to try again. Thanks for working on this!

image

interactions/api/gateway.py
```diff
@@ -1,430 +1,321 @@
...
 
-        :param event: The name of the event.
-        :type event: str
-        :param data: The data of the event.
-        :type data: dict
+        :return: The packet stream.
+        :rtype: Optional[Dict[str, Any]]
         """
+        packet: WSMessage = await self._client.receive()
+        return loads(packet.data) if packet and isinstance(packet.data, str) else None
```
Bluenix2, line 451: By doing this you are throwing away every other useful message you can get from the WebSocket connection. More importantly, you are throwing away `CLOSED`, `CLOSING` and `CLOSE` messages.

interactions/api/gateway.py
```diff
@@ -1,430 +1,321 @@
...
-                    raise GatewayException(code)
-
-                await self.handle_connection(stream, shard, presence)
-
-    async def handle_connection(
-        self, stream: dict, shard: Optional[List[int]] = None, presence: Optional[Presence] = None
+                if self._client.close_code in range(4010, 4014) or self._client.close_code == 4004:
+                    raise GatewayException(self._client.close_code)
+                elif self._client.close_code is not None:
+                    await self._establish_connection()
```
Bluenix2, line 188: Is this recursive?

interactions/api/gateway.py
```diff
@@ -1,430 +1,321 @@
...
+        self.session_id = None if session_id is MISSING else session_id
+        self.sequence = None if sequence is MISSING else sequence
+        self.ready = None
+
+    @property
+    async def __heartbeat_manager(self) -> None:
+        """Manages the heartbeat loop."""
+        while True:
+            if self.__heartbeater.event.is_set():
+                await self.__heartbeat_packet
```
Bluenix2: What happens if you try to send a heartbeat packet when the WebSocket is reconnecting?

interactions/api/gateway.py
```diff
@@ -1,430 +1,321 @@
...
+                self.__heartbeater.event.clear()
+                await sleep(self.__heartbeater.delay / 1000)
```
Bluenix2, line 145: You should use `asyncio.wait` so that you can wait for the first to complete (event or sleep).

interactions/api/gateway.py
```diff
@@ -1,430 +1,321 @@
...
+            else:
+                log.debug("Heartbeat ACK not recieved, reconnecting to Gateway...")
+                await self._client.close()
+                await self._establish_connection()
```
Bluenix2: What happens to the already existing `_establish_connection()` call?

interactions/api/gateway.py
```diff
@@ -1,430 +1,321 @@
...
+
+                await self._handle_connection(stream, shard, presence)
+
+    async def __close(self):
+        """Closes the client's connection and heartbeat with the Gateway."""
+        if self.__task:
+            self.__task: Task()
+            self.__task.cancel()
+        await self._client.close()
+        await self._establish_connection()
```
Bluenix2: Isn't the point for this to close the gateway? Why is it reconnecting?

interactions/api/gateway.py
i0bs: The idea is to be recursive, yeah. Simply re-calling the state acting like a "non-1000 closing code" as gateway docs suggest starting a new websocket connection for reconnecting

interactions/api/gateway.py
i0bs: Will have to check, but it's likely nothing. When a new connection is established, we force the heartbeat delay back to 0 again to avoid an unsynced heartbeater

interactions/api/gateway.py
i0bs: Theory is to essentially reset it and start fresh as written in one of the comments above

interactions/api/gateway.py
i0bs: We run it again since the closing code check is being ran in it, and if it's captured a non-reconnectable code, it'll simply raise an exception. We can move the closing code condition out of the connection establisher, but the sequence of operations would likely remain to be the same as written

interactions/api/gateway.py
Bluenix2: Yes, you should create another WebSocket connection when reconnecting (as the previous one is disconnected) but I would not recommend that it's done recursively.

interactions/api/gateway.py
Bluenix2: I am saying that this line of code can execute when the WebSocket is in the process of reconnecting. Either you accidentally try to send a packet to a closed WebSocket, one that's in-fact None (`self._client` is None), or one that has not yet received the HELLO event / finished connecting.

interactions/api/gateway.py
Bluenix2: Ah update on this, I was a bit confused by `self.__heartbeater.event`. I assumed it was meant to close the heartbeater, so this change is probably not what is appropriate.

interactions/api/gateway.py
i0bs: The manager only happens after we've declared `_client` as a `ClientWebSocketResponse` in the with loop, so I believe it's impossible to have a `TypeError` race condition.

We can add an extra condition that only triggers this when we've ensured the Gateway have sent us a `HELLO` payload. This ties in pretty well to our establisher, however, as any issue whatsoever outside of this should raise an exception when a closing code is found.

interactions/api/gateway.py
i0bs: We refer to `__task` as our attempt to start a new task in closing the event fyi

interactions/api/gateway.py
i0bs: What's the correct type? I extracted `WSMessage` from the type hint on `ClientWebSocketResponse`'s `receive()` method. Or are you referring to `packet.data` being the only return?

i0bs:
Pending tasks we have to make grace on merging:

- Refactor `_establish_connection()` and `_handle_connection()` methods:
    - Launch the heartbeater once `HEARTBEAT` is requested from the Gateway.
    - Read for `CLOSING` and `CLOSED` events from the `ClientWebSocketResponse`.
- Lock the `_client` session once a closing code is detected by awaiting to ensure it's canceled.
- Rework `__receive_packet_stream` to not return `None` when the serialisation operation fails.
    - Maybe pass `MISSING` for clarity?
- Reset the heartbeater upon detecting we're starting a new connection from zombification or the Gateway dispatch.

FayeDel (approved):

sillyElvy (approved):
hello, better gateway 👋

PR #455, opened 2022-01-25
refactor: make methods use the `MISSING` value

sillyElvy:
## About

This pull request refactors methods to use the `MISSING` value
## Checklist

  - (If existent):
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [ ] Bugfix


interactions/api/models/channel.py
```diff
@@ -183,13 +183,13 @@ def __init__(self, **kwargs):
...
-        content: Optional[str] = None,
+        content: Optional[str] = MISSING,
         *,
-        tts: Optional[bool] = False,
+        tts: Optional[bool] = MISSING,
         # attachments: Optional[List[Any]] = None,  # TODO: post-v4: Replace with own file type.
-        embeds=None,
-        allowed_mentions=None,
-        components=None,
+        embeds=MISSING,
```
i0bs: These are missing the `Optional` typing annotation. Please add these to the argument signature

interactions/api/models/member.py
```diff
@@ -182,13 +182,13 @@ async def remove_role(
...
+        content: Optional[str] = MISSING,
         *,
-        components=None,
-        tts: Optional[bool] = False,
+        components=MISSING,
+        tts: Optional[bool] = MISSING,
         # attachments: Optional[List[Any]] = None,  # TODO: post-v4: Replace with own file type.
-        embeds=None,
-        allowed_mentions=None,
+        embeds=MISSING,
```
i0bs: Missing `Optional` typing annotation.

interactions/api/models/member.pyi
```diff
@@ -53,23 +53,25 @@ class Member(DictSerializerMixin):
...
-        components: Optional[Union[ActionRow, Button, SelectMenu, List[Union[ActionRow, Button, SelectMenu]]]] = None,
-        tts: Optional[bool] = False,
+        components: Optional[
+            Union[ActionRow, Button, SelectMenu, List[Union[ActionRow, Button, SelectMenu]]]
+        ] = MISSING,
+        tts: Optional[bool] = MISSING,
         # attachments: Optional[List[Any]] = None,  # TODO: post-v4: Replace with own file type.
-        embeds=None,
-        allowed_mentions=None,
+        embeds=MISSING,
```
i0bs: Missing `Optional` typing annotation.

i0bs (changes requested):

i0bs (changes requested):
I'm not going to comment on every single file needing the `Optional`, but they should target the object that they need. For example, `embeds` should be `Optional[Union[Embed, List[Embed]]` and vice versa.

sillyElvy:
But i can't import it due to circular import

And if I'd put it there without having it imported I'd not pass the pre-commit

i0bs:
How are you hitting circular imports?
Additionally, if there is no way for you to realistically import it, set the type annotation with a string. `"Embed"` can also suffice.

sillyElvy:
> How are you hitting circular imports? Additionally, if there is no way for you to realistically import it, set the type annotation with a string. `"Embed"` can also suffice.

Message is importing Channel and member, so I can't import Embed from Message 

ok, I'll do that

interactions/api/models/channel.py
```diff
@@ -183,13 +183,20 @@ def __init__(self, **kwargs):
...
-        allowed_mentions=None,
-        components=None,
+        embeds: Optional[Union["Embed", List["Embed"]]] = MISSING,  # noqa
+        allowed_mentions: Optional["MessageInteraction"] = MISSING,  # noqa
+        components: Optional[
+            Union[
+                "ActionRow",  # noqa
+                "Button",  # noqa
+                "SelectMenu",  # noqa
+                List[Union["ActionRow", "Button", "SelectMenu"]],  # noqa
```
i0bs: Not requesting, but, is `List[Union[...]]` proper? My understanding of `List` was that you could only apply one type in it. 

image

sillyElvy: I don't know, I'm not familiar enough with it, in my opinion `List[Union[int, str]] ` is the same as `List[int], List[str]`

i0bs: I like the idea of explicitly mentioning what can be a list by singular types. Doing `List[Union[...]]` can seem a little confusing to some because it's a repeat of the base `Union`. This also doesn't seem to follow how `GenericAlias` works in the typing module upon closer investigation, so if we were to ever implement reading from the argument signatures' types in the future, it wouldn't work correctly.

My recommendation is switching over to `List[X], List[Y], ...` in the base `Union` structure.

sillyElvy: Could you request it as a change then? It's easier to just accept it from mobile than to change it

sillyElvy: Hm, I'm gonna change it, let's see if I'm good enough for pre-commit

sillyElvy: OK, I'm done now with that issue

And I managed to make the checks pass

i0bs (approved):

PR #459, opened 2022-01-29
Add BitField objects for User & Application flags

panley01:
Also adds Application flag checks to the client code, checking priv. intents against Application flags to raise exceptions on unpermitted intents before initiating a connection with Discord.

### DIFF
--- a/interactions/client.py
+++ b/interactions/client.py
@@ -11,7 +11,7 @@ from .api.dispatch import Listener
 from .api.error import InteractionException
 from .api.gateway import WebSocket
 from .api.http import HTTPClient
-from .api.models.flags import Intents
+from .api.models.flags import Intents, AppFlags
 from .api.models.guild import Guild
 from .api.models.misc import Snowflake
 from .api.models.team import Application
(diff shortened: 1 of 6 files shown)


i0bs (approved):
You have pleased me.

panley01:
Amended code to account for `_intent` as opposed to `intent` & priv intent flags for unverified apps (_LIMITED suffix). Also moved intent checks before command checks for that sweet nanosecond optimizaton on knowing you fucked up

interactions/client.py
```diff
@@ -256,6 +257,12 @@ async def _ready(self) -> None:
         ready: bool = False
 
         try:
+            if self._intents.GUILD_PRESENCES and not (self.me.flags.GATEWAY_PRESENCE or self.me.flags.GATEWAY_PRESENC_LIMITED):
```
Damego:
```suggestion
            if self._intents.GUILD_PRESENCES and not (self.me.flags.GATEWAY_PRESENCE or self.me.flags.GATEWAY_PRESENCE_LIMITED):
```

i0bs:
Covering over Delta's review part, but the changes look proper and I haven't hit any issues in testing. LGTM. Thank you for contributing, gamer. 🤓

PR #463, opened 2022-02-03
refactor: Minor parameter catching

V3ntus:
## About

- 001c23a897a496764f8f471701a3f25a1b8a4b80 : raise InteractionException if required parameter "description" is missing in Option's
- 423542e : raise InteractionException if parameter "name" contains uppercase characters
- 33f05c6 : raise InteractionException if 
	1. command name is longer than 32 characters
	2. command description is longer than 100 characters
	3. more than 25 options in command
	4. autocomplete is set to true, but choices are present in command decorator
	5. command option name is longer than 32 characters
	6. command option description is longer than 100 characters

## Checklist

  - (If existent): If users do not supply required parameters in command creation, no error handling would take place, and (example) a `TypeError` will occur:
  ```py
  ...
  File "/usr/local/lib/python3.8/dist-packages/interactions/client.py", line 229, in _synchronize
    await self.__create_sync(payload)
  File "/usr/local/lib/python3.8/dist-packages/interactions/client.py", line 144, in __create_sync
    command: ApplicationCommand = ApplicationCommand(
  TypeError: type object argument after ** must be a mapping, not NoneType
  ```
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix


interactions/client.py
```diff
@@ -378,10 +380,13 @@ def decorator(coro: Coroutine) -> Callable[..., Any]:
                     )
                 _option: Option
                 for _option in options:
-                    if not _option._json.get("description"):
+                    if not _option._json.get("description") and _option not in (
+                        OptionType.SUB_COMMAND,
+                        OptionType.SUB_COMMAND_GROUP,
+                    ):
```
sillyElvy: shouldn't this be `_option._json.get("type")` or am I thinking too complicated?

V3ntus: Oh oops. I pressed undo too many times. Will fix

V3ntus (commented):
Should an exception be raised for uppercase names, or should we convert to lower case and pass?

interactions/client.py
```diff
@@ -361,6 +361,17 @@ def decorator(coro: Coroutine) -> Callable[..., Any]:
...
+            for _ in name:
+                if _.isupper():
+                    __name_is_upper = True
+            
+            if __name_is_upper:
+                raise InteractionException(
+                    11,
+                    message="Your command name must not contain uppercase characters (Discord limitation)"
+                )
+
```
sillyElvy:
```suggestion
            for _ in name:
                if _.isupper():
                  raise InteractionException(
                      11,
                      message="Your command name must not contain uppercase characters (Discord limitation)"
                  )

```

V3ntus: For some reason, I thought this would raise an exception for every uppercase char, but no, that's not how exceptions work. Thanks 👍🏾

sillyElvy:
> Should an exception be raised for uppercase names, or should we convert to lower case and pass?

I would raise an exception because it is a bad practice when you are doing it the wrong way and we just pass it the way that it should be

sillyElvy:
Also, please run pre-commit

V3ntus:
Pre-commit struggling server-side today
[10dbeb2](https://github.com/interactions-py/library/pull/463/commits/10dbeb2d100d21253d2ed7813ca212602e005ff7) passes pre-commit local-side

sillyElvy:
> Pre-commit struggling server-side today [10dbeb2](https://github.com/interactions-py/library/pull/463/commits/10dbeb2d100d21253d2ed7813ca212602e005ff7) passes pre-commit local-side

the ubuntu check passes so it should be fine

sillyElvy (approved):
LGTM

interactions/client.py
```diff
@@ -380,6 +380,22 @@ def decorator(coro: Coroutine) -> Callable[..., Any]:
             if name is MISSING:
                 raise InteractionException(11, message="Your command must have a name.")
 
+            elif len(name) > 32:
+                raise InteractionException(
+                    11, message="Command name must not be more than 32 characters."
```
sillyElvy: A: names*
B: use "less than" as below

interactions/client.py
```diff
@@ -380,6 +380,22 @@ def decorator(coro: Coroutine) -> Callable[..., Any]:
...
+                )
+            elif len(description) > 100:
+                raise InteractionException(
+                    11, message="Command description must be less than 100 characters"
```
sillyElvy: descriptions*

interactions/client.py
```diff
@@ -389,11 +405,44 @@ def decorator(coro: Coroutine) -> Callable[..., Any]:
...
+                    if _option.type not in (
+                        OptionType.SUB_COMMAND,
+                        OptionType.SUB_COMMAND_GROUP,
+                    ):
+                        if getattr(_option, "autocomplete", False) and getattr(
+                            _option, "choices", False
+                        ):
+                            raise InteractionException(
+                                11,
+                                message="Autocomplete may not be set to true if choices are present",
```
sillyElvy: This sounds like you raise an exception as an information, I'd do an `log.warning` instead if there's no need to cancel the process.

interactions/client.py
```diff
@@ -389,11 +405,44 @@ def decorator(coro: Coroutine) -> Callable[..., Any]:
...
+                            )
+                        if len(_option.description) > 100:
+                            raise InteractionException(
+                                11,
+                                message="Command option descriptions must be less than 100 characters",
+                            )
+
+                    if len(_option.name) > 32:
+                        raise InteractionException(
+                            11, message="Command option name must be less than 32 characters"
```
sillyElvy: Again: names*

sillyElvy (changes requested):

interactions/client.py
V3ntus: When you say less than, you mean follow L409 here? :
```py
if len(coro.__code__.co_varnames) + 1 < len(options):
```

sillyElvy: I mean the request below this one and the last request, you're writing everywhere "less than" except here

interactions/client.py
V3ntus: Yes, but according to the API, this sounds illegal. If I just throw a warning and let this pass to API requests, will this cause errors down the line?

interactions/client.py
V3ntus: Ahh grammar, not actual code. Gotcha, sorry lol
Will fix, thanks

interactions/client.py
sillyElvy: I don't know, but the usage of "may" implies to me, that it *could* be the case but it *does not have to* what, imo, should not raise an exception then.

V3ntus: Alright, that's fair. Will change

sillyElvy (approved):

i0bs (approved):
Loving the appropriated use of the exception here. :)

Please run `pre-commit` one last time.

V3ntus:
`pre-commit` returned all green, nothing to commit

PR #465, opened 2022-02-04
feat: implement Member.id

Toricane:
## About

This pull request is about implementing `interactions.Member.id`, which lets you use `member.id` instead of `member.user.id`.
Another example is `ctx.author.id` instead of `ctx.author.user.id`.

## Checklist

  - (If existent):
  - [ ] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix

### DIFF
--- a/interactions/api/models/member.py
+++ b/interactions/api/models/member.py
@@ -3,7 +3,7 @@ from typing import List, Optional, Union
 
 from .channel import Channel
 from .flags import Permissions
-from .misc import MISSING, DictSerializerMixin
+from .misc import MISSING, DictSerializerMixin, Snowflake
 from .role import Role
 from .user import User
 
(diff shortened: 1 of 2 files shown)


sillyElvy (changes requested):
This does not reflect the current api scheme, which is not including the ID into the [member object](https://discord.com/developers/docs/resources/guild#guild-member-object)

Because of this, I am against an implementation of the ID as a members attribute.

In my opinion you could add it as a property to the member class. But it's up to @goverfl0w whether he wants that property there or not.

Toricane:
Okay, I'll wait for fl0w

i0bs:
I agree with @EdVraz , I'd personally like to see that as a property method instead. The attributes per model are meant to really respect the API schema.

Toricane:
Ok I will change it to a property later today

sillyElvy (approved):
LGTM

PR #466, opened 2022-02-04
feat: fill Context's Guild and Channel when possible, or replace with MISSING

Catalyst4222:
## About

This pull request checks the cache if the guild or channel is absent from the Context class. If there is no data, and the id (if it exists) is not found in cache, then the value is replaced with MISSING to signal that the data wasn't found

## Checklist

  - (If existent):
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix

### DIFF
--- a/interactions/context.py
+++ b/interactions/context.py
@@ -27,8 +27,8 @@ class Context(DictSerializerMixin):
     :ivar Optional[Message] message?: The message data model.
     :ivar Member author: The member data model.
     :ivar User user: The user data model.
-    :ivar Union[Channel, MISSING] channel: The channel data model.
-    :ivar Union[Guild, None, MISSING] guild: The guild data model.
+    :ivar Optional[Channel] channel: The channel data model.
+    :ivar Optional[Guild] guild: The guild data model.
     """
 
     __slots__ = ("message", "member", "author", "user", "channel", "guild", "client")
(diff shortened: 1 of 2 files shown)


FayeDel (approved):
LGTM 👍🏻

interactions/context.py
```diff
@@ -27,8 +27,8 @@ class Context(DictSerializerMixin):
     :ivar Optional[Message] message?: The message data model.
     :ivar Member author: The member data model.
     :ivar User user: The user data model.
-    :ivar Channel channel: The channel data model.
-    :ivar Guild guild: The guild data model.
+    :ivar Union[Channel, MISSING] channel: The channel data model.
+    :ivar Union[Guild, None, MISSING] guild: The guild data model.
```
i0bs: Docstring on a sentinel value and a None return should always respect `Optional[]` typing schema.

i0bs (changes requested):

i0bs (approved):

i0bs:
LGTM, in the future I might refactor this within the mixin proto impl.

I would run `pre-commit` one last time.

Catalyst4222:
About that, it's all green here
image

FayeDel:
Thanks for contributing! Will merge :)

PR #477, opened 2022-02-05
fix: passing command names in autocomplete decorator

V3ntus:
## About
This fixes autocomplete when passing command names in the decorator
- Adheres to scope specific commands
- Properly handles incorrect choice population

---

## Usage:
```py
@bot.command(
	...
    options=[interactions.Option(
        type=interactions.OptionType.STRING,
        name="topic",
        description="Test autocomplete",
        autocomplete=True  # define a command option used for autocomplete
    )]
)
async def test_command(ctx, topic: str = ""):
	...
	
# define an autocomplete callback
@bot.autocomplete(command="test", name="topic")
async def do_autocomplete(ctx, topic: str = ""):
    await ctx.populate([
        interactions.Choice(
            name="One",
            value="one"
        ),
        interactions.Choice(
            name="Two",
            value="two"
        )
    ])
```

## Checklist

  - (If existent):
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix


V3ntus:
~~Todo: figure out why the return from `get_application_commands` is inconsistent with defined commands, and how to get a consistent response from the Discord API~~

Passing a guild ID in `get_application_commands` will fix. Without passing an ID, this defaults to global commands in the API when I was wanting commands in a scope, thus the inconsistency.

mAxYoLo01:
Ok so this has fixed some previous issues, however when disable_sync = True, the autocomplete fails:
```python
  File "*\launch.py", line 57, in <module>
    @bot.autocomplete("cog", "load")
  File "*\interactions\client.py", line 616, in autocomplete
    application_id=self.me.id, guild_id=_command_obj.guild_id
AttributeError: 'NoneType' object has no attribute 'guild_id'
```

Would switching the `name` and `command` arguments of the autocomplete function make more sense? To go from general -> option

V3ntus:
> Ok so this has fixed some previous issues, however when disable_sync = True, the autocomplete fails:

Ah I had a hunch this would break. Will look into it

> Would switching the `name` and `command` arguments of the autocomplete function make more sense? To go from general -> option

Oh yes, I agree. I'll switch them over

mAxYoLo01:
Can confirm everything works on my side, **LGTM!**

interactions/client.py
```diff
@@ -579,8 +586,34 @@ def decorator(coro: Coroutine) -> Any:
...
+    @staticmethod
+    def _find_command(commands: List, command: str) -> ApplicationCommand:
+        """
+        Iterates over commands and returns an ApplicationCommand if it matches the name from command
+
+        :ivar commands: The list of dicts to iterate through
+        :type commands: List[ApplicationCommand]
+        :ivar command: The name of the command to match:
+        :type command: str
+        """
```
FayeDel: Add a `:return` and `:rtype` to documentation that states that this will return an `ApplicationCommand`

interactions/client.py
```diff
@@ -579,8 +586,34 @@ def decorator(coro: Coroutine) -> Any:
 
         return decorator
 
+    @staticmethod
+    def _find_command(commands: List, command: str) -> ApplicationCommand:
```
FayeDel: Rename the typehint from `List` to `List[Dict]`
(to match consistency with the associated `.pyi` file.

interactions/client.py
```diff
@@ -579,8 +586,34 @@ def decorator(coro: Coroutine) -> Any:
...
+        """
+        Iterates over commands and returns an ApplicationCommand if it matches the name from command
+
+        :ivar commands: The list of dicts to iterate through
+        :type commands: List[ApplicationCommand]
```
FayeDel: Documentation mismatch with input args, same change as the typehint change above ^

i0bs: Additionally, please use either \`\` or :class:\`\` formatting on L592.

interactions/client.py
```diff
@@ -589,7 +589,7 @@ def decorator(coro: Coroutine) -> Any:
     @staticmethod
     def _find_command(commands: List[Dict], command: str) -> ApplicationCommand:
         """
-        Iterates over commands and returns an ApplicationCommand if it matches the name from command
+        Iterates over `commands` and returns an :class: ``ApplicationCommand`` if it matches the name from `command`
```
V3ntus: Is this appropriate RST usage?

V3ntus:
`pre-commit` returning green here too, nothing to commit

i0bs:
I'll move forward with merging this, `pre-commit` passed green on my end and the CI job for windows is specifically broken for the time being.

interactions/client.py
```diff
@@ -627,8 +634,36 @@ def decorator(coro: Coroutine) -> Any:
 
         return decorator
 
+    @staticmethod
+    def _find_command(commands: List[Dict], command: str) -> ApplicationCommand:
+        """
+        Iterates over `commands` and returns an :class: ``ApplicationCommand`` if it matches the name from `command`
```
i0bs:
```suggestion
        Iterates over `commands` and returns an :class:`ApplicationCommand` if it matches the name from `command`
```

i0bs (changes requested):

i0bs (approved):

PR #479, opened 2022-02-06
feat: Abstract `modify` method in data models for simplified usage.

### LINKED ISSUE #472: [REQUEST] expand helper methods
I am thinking of "expanding" existing helper methods like `member.edit()` into, for example,`member.set_timeout` etc
```py
async def set_timeout(timeout):
  return await self.edit(timeout=timeout)
```

what do you think about this?

sillyElvy:
Allows the usage of 
```py
await channel.set_name("newname")
```

instead 
```py
await channel.modify(name="newname")
```

## Checklist

  - (If existent): #472
  - [ ] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix

### DIFF
--- a/interactions/api/models/channel.py
+++ b/interactions/api/models/channel.py
@@ -339,7 +339,12 @@ class Channel(DictSerializerMixin):
         )
         return Channel(**res, _client=self._client)
 
-    async def set_name(self, name: str, reason: Optional[str] = None) -> "Channel":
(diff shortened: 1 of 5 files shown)


sillyElvy:
i won't include member and role because those actually require a guild_id

i0bs (approved):

PR #494, opened 2022-02-09
refactor: create a ContextMixin class, and extract methods into there

### LINKED ISSUE #487: [REQUEST] Mixin implementations
**What is the feature you're proposing?**
With the ever-growing complexity of context functionality being used in our commands and components, we're in a situation where we need to abstract and break apart the code written for it in more simplified methods that don't involve dunder abstractions. Because of this, we're looking into metaclassing other classes that purely add methods as "mixins."

- `EventMixin`: a mixin that can build event data, this will theoretically run on top of `DictMixin` but for Gateway-unique events that do not work off of objects.
- `ContextMixin`: a mixin that will serve the purpose of simple `super().method(**kwargs)` method implementations in various context managers, e.g. send and edit methods.
- `ErrorMixin`: a mixin for base support of the `ErrorFormatter` class that allow custom `message`, `id` and `extended_message` attributes. `extended_message` should be designed to represent the nested information of a dictionary to a `key -> value` arrow tree in the form of a string.

**Additional Information**
- [ ] My feature request is related to an existing Issue.
    - Issue (if referenceable):

Catalyst4222:
## About

This pull request adds in a ContextMixin class, inheriting from DictSerializerMixin and designed to be used in the creation of new context classes

## Checklist

  - (If existent): Adds in one mixin from [487](https://github.com/interactions-py/library/issues/487)
  - [ ] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix


interactions/context.py
```diff
@@ -574,80 +452,211 @@ async def edit(
...
+        _ephemeral: int = (1 << 6) if ephemeral else 0
+        self.callback = InteractionCallbackType.DEFERRED_CHANNEL_MESSAGE_WITH_SOURCE
+
+        await self.client.create_interaction_response(
+            token=self.token,
+            application_id=int(self.id),
+            data={"type": self.callback.value, "data": {"flags": _ephemeral}},
+        )
+
+    async def send(self, content: str = MISSING, **kwargs):
```
sillyElvy: Please add optional to content

sillyElvy (changes requested):

interactions/context.py
```diff
@@ -574,80 +452,211 @@ async def edit(
...
+        payload = await super().send(content, **kwargs)
+
+        if not self.deferred:
+            self.callback = InteractionCallbackType.CHANNEL_MESSAGE_WITH_SOURCE
+
+        _payload: dict = {"type": self.callback.value, "data": payload._json}
```
sillyElvy, line 642: If I am seeing this correctly, payload would contain the `_client` attribute what would make the request invalid?

interactions/context.py
```diff
@@ -574,80 +452,211 @@ async def edit(
...
+
+        msg = None
+        if self.responded or self.deferred:
+            if self.deferred:
+                res = await self.client.edit_interaction_response(
+                    data=payload._json,
```
sillyElvy, line 648: When you create the `_payload` why don't you use it here? (applies to every occurance)

interactions/context.py
```diff
@@ -708,57 +717,8 @@ async def func():
...
-        :rtype: Guild
-        """
-
-        res = await self.client.get_guild(int(self.guild_id))
-        self.guild = Guild(**res, _client=self.client)
-        return self.guild
-
 
-class ComponentContext(CommandContext):
+class ComponentContext(ContextMixin):
```
sillyElvy: Personally I would make it inherit from `CommandContext`, this would have the benefit of not having the `send` and `edit` methods twice and it would still inherit from the Mixin because `CommandContext` does that.

sillyElvy (changes requested):

interactions/context.py
Catalyst4222: The mixin was made specifically to avoid one context subclassing the other. The sending is partially split between the mixin and the classes because each context has a different way of responding

interactions/context.py
Catalyst4222: It seems to work, and current unstable uses `payload._json` after setting `_client`

sillyElvy: no it does not

```py
        self.message = payload
        self.message._client = self.client
        _payload: dict = {"type": self.callback.value, "data": payload._json}
```

interactions/context.py
Catalyst4222: It is used in both send function, in the same way that current unstable does

interactions/context.py
Catalyst4222: Fixed, thanks for telling me

interactions/context.py
```diff
@@ -17,7 +17,7 @@
 log: Logger = get_logger("context")
 
 
-class Context(DictSerializerMixin):
+class ContextMixin(DictSerializerMixin):
```
i0bs:
```suggestion
class _Context(DictSerializerMixin):
```

The naming for mixins will be changing since we are considering dunder to be applied to our attributes, instances and methods as of the 2022 roadmap.

interactions/context.py
```diff
@@ -587,79 +463,216 @@ async def edit(
...
+            token=self.token,
+            application_id=int(self.id),
+            data=payload,
+        )
+        self.responded = True
+
+        return payload
+
+
+class CommandContext(ContextMixin):
```
i0bs:
```suggestion
class CommandContext(_Context):
```

interactions/context.pyi
```diff
@@ -1,77 +1,132 @@
...
+from .enums import InteractionType as InteractionType
+from .models.command import Choice as Choice
+from .models.component import ActionRow as ActionRow
+from .models.component import Button as Button
+from .models.component import Modal as Modal
+from .models.component import SelectMenu as SelectMenu
+from .models.misc import InteractionData as InteractionData
 
-class Context(DictSerializerMixin):
+class ContextMixin(DictSerializerMixin):
```
i0bs:
```suggestion
class _Context(DictSerializerMixin):
```

interactions/context.pyi
```diff
@@ -1,77 +1,132 @@
...
+        embeds: Optional[Union[Embed, List[Embed]]] = MISSING,
+        allowed_mentions: Optional[MessageInteraction] = MISSING,
+        message_reference: Optional[MessageReference] = MISSING,
+        components: Optional[
+            Union[ActionRow, Button, SelectMenu, List[ActionRow], List[Button], List[SelectMenu]]
+        ] = MISSING
+    ) -> Message: ...
+    async def popup(self, modal: Modal) -> None: ...
+
+class CommandContext(ContextMixin):
```
i0bs:
```suggestion
class CommandContext(_Context):
```

interactions/context.pyi
```diff
@@ -1,77 +1,132 @@
...
+    ) -> Message: ...
+    async def defer(self, ephemeral: Optional[bool] = False) -> None: ...
     async def delete(self) -> None: ...
-    async def popup(self, modal: Modal): ...
     async def populate(self, choices: Union[Choice, List[Choice]]) -> List[Choice]: ...
-    async def get_channel(self) -> Channel: ...
-    async def get_guild(self) -> Guild: ...
 
-class ComponentContext(CommandContext):
+class ComponentContext(ContextMixin):
```
i0bs:
```suggestion
class ComponentContext(_Context):
```

i0bs (changes requested):

i0bs (approved):

sillyElvy (approved):
I read it like 10 times, I hope the code is fine

Issue #512, opened 2022-02-12
[REQUEST] Message.mentions should be a list of Users


mAxYoLo01:
**What is the feature you're proposing?**
Message.mentions currently is a `dict`, while it says in its documentation it should be a `List[User]`.

**Additional Information**
- [ ] My feature request is related to an existing Issue.
    - Issue (if referenceable):

i0bs:
Closing in favor of #513

PR #513, opened 2022-02-12
Message.mentions as a list of users

mAxYoLo01:
## About

Transform Message.mentions as a list of Users instead of a dict

## Checklist

  - (If existent): #512 
  - [ ] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix

### DIFF
--- a/interactions/api/models/message.py
+++ b/interactions/api/models/message.py
@@ -237,7 +237,7 @@ class Message(DictSerializerMixin):
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
-        
+
         self.id = Snowflake(self.id) if self._json.get("id") else None
         self.channel_id = Snowflake(self.channel_id) if self._json.get("channel_id") else None
         self.guild_id = Snowflake(self.guild_id) if self._json.get("guild_id") else None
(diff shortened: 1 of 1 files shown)


sillyElvy:
Could you quickly send that dictionary it returns, I'd like to know if it returns a user or a member

mAxYoLo01:
`'mentions': [{'id': '247040742021791746', 'username': 'mAxYoLo01', 'avatar': 'a_56b8ad46d1aea2de169a09559a55811a', 'discriminator': '4491', 'public_flags': 64}, {'id': '730465536525467819', 'username': 'BTE FR Test Bot', 'avatar': None, 'discriminator': '3508', 'public_flags': 0, 'bot': True}]`
That's what I tried it with. I saw on Discord's Docs that it should return a partial member field (?) but I don't see what this could be.

sillyElvy:
> That's what I tried it with. I saw on Discord's Docs that it should return a partial member field (?) but I don't see what this could be.

Same, that's the reason I asked for the dictionary. But it looks pretty much like a normal user, so it looks good to me

sillyElvy (approved):

i0bs:
Labeling this as a wontfix, this violates the current API schema for the `mentions` field in a `Message` object. The Gateway can currently dispatch `MESSAGE_CREATE` and `MESSAGE_UPDATE` events that include partial `Member` objects, however, we're pretty much throwing that data away. A new PR should be created conforming to this if this were to continue.

Issue #596, opened 2022-03-01
[BUG] TypeError 'NoneType' in get_member() if user absent from guild


Donbur4156:
### Describe the bug.

If I try to get an Member Object with get_member() for an absent User it caused a TypeError.
I think it can´t resolve the roles correctly.

Foxtrots hint: 
self.roles had to check `if _json.get("roles")`

### List the steps.

Try to create a Member Object:
`member = interactions.Member(**await bot._http.get_member(member_id=dc_id, guild_id=serverid), _client=bot._http)`
I generalize this, so the Traceback have other variables

### What you expected.

I expected a Member Object if User is in Guild.
If not I appreciate a None Object to check `if self.member`

### What you saw.

  File "\bot\objects.py", line 79, in get_member_obj
    self.member = di.Member(**await self.bot._http.get_member(member_id=self.dc_id, guild_id=c.serverid), _client=self.bot._http)
  File "\lib\site-packages\interactions\api\models\member.py", line 75, in __init__
    self.roles = [
TypeError: 'NoneType' object is not iterable

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "\lib\site-packages\discord\ext\commands\bot.py", line 939, in invoke
    await ctx.command.invoke(ctx)
  File "\lib\site-packages\discord\ext\commands\core.py", line 863, in invoke
    await injected(*ctx.args, **ctx.kwargs)
  File "\lib\site-packages\discord\ext\commands\core.py", line 94, in wrapped
    raise CommandInvokeError(exc) from exc
discord.ext.commands.errors.CommandInvokeError: Command raised an exception: TypeError: 'NoneType' object is not iterable

### What version of the library did you use?

release

### Code of Conduct

- [X] I agree to follow the contribution requirements.

PR #595, opened 2022-03-01
fix!: Allow usage of builtin json instead of orjson 

sillyElvy:
## Checklist

- [x] I've ran `pre-commit` to format and lint the change(s) made.
- [x] I've checked to make sure the change(s) work on `3.8.6` and higher.
- [ ] This fixes/solves an [Issue](https://github.com/goverfl0w/discord-interactions/issues).
  - (If existent):
- [ ] I've made this pull request for/as: (check all that apply)
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix

PR #597, opened 2022-03-01
fix!:  member role attr 

sillyElvy:
## Checklist

  - (If existent): #596 
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix


sillyElvy:
please merge #595 first

interactions/api/models/member.pyi
```diff
@@ -16,7 +16,7 @@ class Member(DictSerializerMixin):
     user: Optional[User]
     nick: Optional[str]
     avatar: Optional[str]
-    roles: List[Role]
+    roles: Union[List[Role], List[int]]
```
```suggestion
    roles: List[int]
```

Upon close investigation, we've type hinted this incorrectly the whole time. The API only returns a list of integers, and there's no way to request a specific role via. ID from the endpoints without us checking from a list of role objects via. `GET Guild Roles`.

i0bs (changes requested):

i0bs (approved):

PR #633, opened 2022-03-10
feat: implement magic methods

### LINKED ISSUE #624: [REQUEST] automatically update ``_json`` in models to make user changes to them easier
### Describe the feature.

Implementing a `__setattr__` method for models like embed, button, select, etc seems pretty useful to allow easy user changes and instantly update `_json`

### Code of Conduct

sillyElvy:
Implement magic methods to make usage of models easier for users

## Checklist

  - (If existent): resolves: #624 
  - [ ] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix


sillyElvy:
oh i forgot embed

i0bs (commented):
Have you tried adding this `__setattr__` magic style into the serializer itself? The issue boils down to trying to recursively remap the attributes we've added to our serialization subclass.

sillyElvy:
I don't quite understand how you mean that

FayeDel:
I believe they were asking if you've tried implementing this:
```py
    def __setattr__(self, key, value) -> None:
        super().__setattr__(key, value)
        if key != "_json" and (key not in self._json or value != self._json.get(key)):
            if value is not None and value is not MISSING:
                self._json.update({key: value})

            elif value is None and key in self._json.keys():
                del self._json[key]
```
within the serialiser mixin itself and not just the objects in question

sillyElvy:
I didn't do that, I also don't see where it would be useful since it is only called once and the other `__setattr__` methods are updating the json correctly, so how would it help having it in the mixin?

interactions/api/gateway.py
```diff
@@ -366,7 +365,8 @@ def _dispatch_event(self, event: str, data: dict) -> None:  # sourcery no-metric
                 _name: str = _event_path[0] if len(_event_path) < 3 else "".join(_event_path[:-1])
                 __obj: object = getattr(__import__(path), _name)
 
-                if name in {"_create", "_add"}:
+                # name in {"_create", "_add"} returns False (tested w message_create)
+                if any(_ in name for _ in {"_create", "_update", "_add", "_remove", "_delete"}):
```
FayeDel, line 369: Not sure why this was changed in this PR. Mind explaining to me why its needed to have the client attribute in delete/remove events?

sillyElvy: It is not really needed, but I thought some users could have some kind of use for it so I added them

also I thought I did it in the HTTP pr whooops

FayeDel (approved):

PR #638, opened 2022-03-13
refactor: Allow dispatching to the voice ext if it is loaded.

sillyElvy:
## About

This pull request modifies paths and imports to allow the `VoiceStateUpdate` event to be dispatched to the ext.

## Checklist

  - (If existent):
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix


i0bs:
This will not work the way you want with adding onto the path string. If we see a voice event but the developer does not have the ext downloaded, we're hitting an import error. The try/except feels like a messy way of going around it

I think as the wait for ext did, we should be subclassing the web socket client for dispatching voice

sillyElvy:
Tried that but couldn't get it to work, so idk how to then

i0bs:
As discussed in the #external channel on the support server, there's been a given solution for this. As such, I do not see any reason or need for this to stay open.

PR #668, opened 2022-03-24
fix: Guild commands synchronization

GeomKid:
## About

## Checklist

  - (If existent):
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix


interactions/client.py
```diff
@@ -183,10 +183,10 @@ async def __bulk_update_sync(self, data: List[dict], delete: Optional[bool] = Fa
 
         for command in data:
             if command.get("guild_id"):
-                if guild_commands.get(command["guild_id"]):
-                    guild_commands[command["guild_id"]].append(command)
+                if guild_commands.get(int(command["guild_id"])):
+                    guild_commands[int(command["guild_id"])].append(command)
                 else:
-                    guild_commands[command["guild_id"]] = [command]
+                    guild_commands[int(command["guild_id"])] = [command]
```
arHSM: No

arHSM (changes requested):

interactions/client.py
sillyElvy: Agreed

GeomKid: other than that there's nothing to change? I di d that because the input sometimes the command["guild_id"] will be string so to make it easier I did it to int

arHSM: It's gonna be a string all the time, snowflakes are transported as strings over the api, no need to waste processing over something internal

GeomKid: but when i tested if some are string some are int the synchro will not be done and some will not be loaded

interactions/client.py
```diff
@@ -256,8 +256,8 @@ async def _synchronize(self, payload: Optional[dict] = None) -> None:
                 await self.__create_sync(payload)
         else:
             to_delete.extend(command for command in commands if command not in cache)
-        await self.__bulk_update_sync(to_sync)
-        await self.__bulk_update_sync(to_delete, delete=True)
+        if len(to_sync) >= 1:
+            await self.__bulk_update_sync(to_sync + commands)
```
i0bs: How are you actually communicating that we're deleting? If your goal is only 1 bulk call, I'd also recommend removing the deletion establishment in the bulk update sync method.

GeomKid: In the bulk_update_sync() this is the part where the updates to Discord get uploaded 
Snip1
in `self.http.overwrite_application_command`
is 
```py
async def overwrite_application_command(
        self, application_id: int, data: List[dict], guild_id: Optional[int] = None
    ) -> List[dict]:
        """
        Overwrites application command(s) from a scope to the new, updated commands.

        ..note:
            This applies to all forms of application commands (slash and context menus)

        :param application_id: Application ID snowflake
        :param data: The dictionary that contains the command (name, description, etc)
        :param guild_id: Guild ID snowflake to put them in, if applicable.
        :return: An array of application command objects.
        """
        url = (
            f"/applications/{application_id}/commands"
            if not guild_id
            else f"/applications/{application_id}/guilds/{guild_id}/commands"
        )

        return await self._req.request(Route("PUT", url), json=data)
```
In Discord API docs  
image
and 
image
which means that if the data given doesn't have those items it will be deleted
Therefore in my code the `to_sync + commands` takes account to all new and old commands together to replace it

i0bs: I don't think you understood what I meant.

In your code, you've stopped using the `delete` kwarg that our internal bulk sync method has for specifically deleting messages. If you're not going to use this, (presuming you're wanting to make only 1 bulk call now) remove the logic from this method including the kwarg allowing this.

i0bs:
Before reviewing, I have a merge conflict with `client.py`. Please inform me when such is resolved and ready for reviewal.

i0bs (approved):

i0bs:
I'll be going through with merging this pull request, thank you for contributing!

Issue #664, opened 2022-03-21
[BUG] 


dugaldp:
### Describe the bug.

When trying to return a list of members in a guild using:
```
for member in bot.guilds().members:
        print(member)
```
An error was given

### List the steps.

$ pip install -U git+https://github.com/interactions-py/library@unstable
Import interactions module
Created a slash command with the code above

### What you expected.

I expected the terminal to print out a list of members in the guild 

### What you saw.

Instead, I received this traceback error given from my Python terminal:
```
Traceback (most recent call last):
    for member in bot.guilds().members:
  File "/opt/homebrew/lib/python3.9/site-packages/interactions/client.py", line 105, in guilds
    return [Guild(**_, _client=self._http) for _ in self._http.cache.self_guilds.view]
  File "/opt/homebrew/lib/python3.9/site-packages/interactions/client.py", line 105, in <listcomp>
    return [Guild(**_, _client=self._http) for _ in self._http.cache.self_guilds.view]
TypeError: interactions.api.models.guild.Guild() got multiple values for keyword argument '_client'
```

### What version of the library did you use?

unstable

### Code of Conduct

- [X] I agree to follow the contribution requirements.

i0bs:
This Issue is really hard to understand. I get what it means, and we've appropriated a fix for this. Next time, please format your issues to be easier to read.

PR #670, opened 2022-03-24
fix!: fix ``guilds`` property in client

sillyElvy:
## About

Fixes an error occurring because of the cached guild object already having an HTTPClient

## Checklist

  - (If existent): #664 
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix


interactions/client.py
```diff
@@ -102,7 +102,7 @@ def __init__(
     @property
     def guilds(self) -> List[Guild]:
         """Returns a list of guilds the bot is in."""
-        return [Guild(**_, _client=self._http) for _ in self._http.cache.self_guilds.view]
+        return [Guild(**_) for _ in self._http.cache.self_guilds.view]
```
i0bs, line 105: What are we ditching the `_client` declaration here for? Would it not be useful in terms of DX for a bot developer to be able to do something to a guild by easily reading off of this property?

```py
for guild in bot.guilds:
    if guild.name == "foo":
        await guild.method()
```

sillyElvy: it seems to be present already

I can't see any modification in cache, but ``'Guild' got multiple values for '_client'``

sillyElvy:
```py
  File "C:\Users\\Desktop\PycharmProjects\library\main.py", line 12, in on_ready
    print(bot.guilds)
  File "C:\Users\\Desktop\PycharmProjects\library\interactions\client.py", line 105, in guilds
    return [Guild(**_, _client=self._http) for _ in self._http.cache.self_guilds.view]
  File "C:\Users\\Desktop\PycharmProjects\library\interactions\client.py", line 105, in <listcomp>
    return [Guild(**_, _client=self._http) for _ in self._http.cache.self_guilds.view]
TypeError: interactions.api.models.guild.Guild() got multiple values for keyword argument '_client'
```

i0bs:
In conjunction to `client.latency` which can be called upon prior to `client.start()`, I cannot access the `client.guilds` property method at all in an event listener in a main bot file. Trying this produces an `AttributeError`, see:

```py
@client.event
async def on_ready():
    print(len(client.guilds))
```

sillyElvy:
for me it results in an empty list if not populated in the same event... how would it raise an attribute error?

i0bs:
This is from the latest `unstable` commit:

image

sillyElvy:
but that would mean the property is nonexistent... how's that possible?

i0bs:
beats me i didn't write the code 😂

sillyElvy:
> beats me i didn't write the code 😂

still doesn't make sense that a property, what is in the code, disappears on runtime?

i0bs:
let me rephrase:

you're asking the wrong person why, i have zero clue why it's erroring. all i know is that it's broken and it needs to be fixed. pls fix

sillyElvy:
Sorry I can't even start the bot because of `ImportError: cannot import name 'WebSocketClient' from partially initialized module 'interactions.api.gateway.client' (most likely due to a circular import) (C:\Users\edinv\Desktop\PycharmProjects\library\interactions\api\gateway\client.py)` 🙂

Nanrech:
Relatable

i0bs:
Latest unstable commits have fixed the circular import issue.

sillyElvy:
```py
@bot.event
async def on_ready():
    print(bot.guilds)
    print(len(bot.guilds))
```

This results in:

grafik


I don't know what is causing the issue for you

i0bs:
Seems that my env was messed up

interactions/bot.py
```diff
@@ -102,7 +102,7 @@ def __init__(
     @property
     def guilds(self) -> List[Guild]:
         """Returns a list of guilds the bot is in."""
-        return [Guild(**_, _client=self._http) for _ in self._http.cache.self_guilds.view]
+        return [Guild(**_) for _ in self._http.cache.self_guilds.view]
```
i0bs, line 105: How come we're ditching the client? What if bot developers want to call methods in a guild from here?

sillyElvy: https://github.com/interactions-py/library/pull/670#discussion_r839827602

i0bs (approved):

Issue #662, opened 2022-03-20
[BUG] (Guild) Commands always update on startup, even if no changes have been made.


sillyElvy:
### Describe the bug.

Commands are synced, even if they should not be. This runs into particularly high rate limits when changing one command and all other commands are synced again. 

Disabling synchronization is a temporal solution, but it fails when it comes to updating single commands.

### List the steps.

1. Import module
2. Create guild command 
3. Start Bot
4. Restart Bot -> Command is synced without change

### What you expected.

1. Import module
2. Create guild command 
3. Start Bot
4. Restart Bot -> Command is not synced if not changed

### What you saw.

Check debug logs on startup; User reports on discord

### What version of the library did you use?

unstable

### Version specification

Any.

### Code of Conduct

- [X] I agree to follow the contribution requirements.

PR #672, opened 2022-03-25
fix!: constant command synchronization even without changes on the command

sillyElvy:
## About

This PR fixes an issue where commands would always be synced.

## Checklist

  - (If existent): #662 
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix

### DIFF
--- a/interactions/client.py
+++ b/interactions/client.py
@@ -84,6 +84,10 @@ class Client:
         self._token = token
         self._extensions = {}
         self._scopes = set([])
+        self.__to_sync = []
+        self.__to_delete = []
+        self.__has_commands = False
+        # determines if any command has been found in the code. if not, all existing commands will be deleted.
         self.me = None
         _token = self._token  # noqa: F841
         _cache = self._http.cache  # noqa: F841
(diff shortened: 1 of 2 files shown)


interactions/client.py
```diff
@@ -132,14 +132,34 @@ async def __compare_sync(self, data: dict, pool: List[dict]) -> bool:
         :return: Whether the command has changed or not.
         :rtype: bool
         """
-        attrs: List[str] = ["type", "name", "description", "options", "guild_id"]
+        attrs: List[str] = [
```
i0bs, line 135: So, I was thinking on this months back when I was refactoring the whole sync system. How would you feel if we did this instead?

```py
attrs: List[str] = [name for name in ApplicationCommand.__dict__.keys() if not name.startswith("_")]
```

This would simply generate a whole list of attributes to check for if they're not dundered. This will obviously ignore `_json`, and (if potentially added in the future) `_client`. Additionally, since our actual comparison condition checks for if the attribute name exists in the pool of data we're looking for, we don't have to worry about whether we're excessively checking for certain attributes or not. It also avoids us having to constantly update the list of attributes every time Discord releases new features to the application command object, as we only have to update the type annotation in the class itself. (Pretty much avoids WET code)

sillyElvy: Yeah, I'll do that, but is there a benefit of using `__dict__` instead of just `__slots__`?

i0bs: I'm not sure if `__slots__` can actually be read from in terms of reading data. If they can, then no benefit besides preference of how you're going about it.

sillyElvy: Done

i0bs (approved):

Issue #679, opened 2022-03-28
[REQUEST] Adding ``icon_url`` property for returning the guild's icon under api.models.guild.Guild


B1ue-Dev:
### Describe the feature.

Title.

### Code of Conduct

- [X] I agree to follow the contribution requirements.

i0bs:
This Issue did not follow our contribution requirements. Next time, please be sure to follow them.

PR #680, opened 2022-03-28
feat: Adding ``icon_url`` property for returning the guild's icon under api.models.guild.Guild

B1ue-Dev:
## About

This pull request is for adding ``icon_url`` property for returning the guild's icon under ``api.models.guild.Guild``, which is currently not available and make it harder for users to get the guild's icon URL for different purposes, such as adding guild's icon into an embed message.

## Checklist

  - (If existent): https://github.com/interactions-py/library/issues/679
  - [ ] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix


r-warfield:
dont you need to check for MISSING instead of None

B1ue-Dev:
Huh? I tested whenever when guild has no icon and the library returned ``None`` in the ``icon`` hash, so if not None can be used as well, since it is still None anyway.

i0bs:
`MISSING` is a type used for when we're sending something to the API but not defaulting to `None`, as it converts to `null` which some endpoints use for overwriting data.

r-warfield:
ok that makes sense

interactions/api/models/guild.py
```diff
@@ -1600,6 +1600,21 @@ async def search_members(self, query: str, limit: Optional[int] = 1) -> List[Mem
...
 
+    @property
+    def icon_url(self) -> str:
+        """
+        Returns the URL of the guild's icon.
+
+        :return: URL of the guild's icon (None will be returned if none of icon is set)
+        :rtype: str
+        """
+        url = "https://cdn.discordapp.com/"
```
AstreaTSS: `url`'s declaration here is unnecessary - this line could be removed, and L1613 could be changed to include this "base" part instead (as well as assigning the variable instead of adding to it).

interactions/api/models/guild.py
```diff
@@ -1600,6 +1600,21 @@ async def search_members(self, query: str, limit: Optional[int] = 1) -> List[Mem
...
+        if self.icon is not None:
+            url += f"icons/{int(self.id)}/{self.icon}.png"
```
AstreaTSS: This doesn't account for animated guild icons.

B1ue-Dev: Ohhh, right, I forgot about this.

interactions/api/models/guild.py
B1ue-Dev: @Astrea49 I think this is fine, since all the ``avatar_url`` or ``icon_url`` property in the library itself is doing the same thing. Check out the ``avatar_url`` under ``api.models.user.User``.

AstreaTSS: It's a *bit* different there - an icon with a value of `None` there means there's still *an* icon for that user (the default Discord ones), while a `None` value for guilds means, well, it has no icon, being represented by its initials instead (which in this case don't count as icons).

There, no matter what, the "base" URL is needed for every part of the if statement, while it's only needed in one part here.

(Ultimately, this *is* nitpicking a bit, but it would play nicer with some typehinters. It's not a *strictly* necessary change.)

B1ue-Dev: Ah, I see. Will be fixed it in a moment after I test stuff again.

B1ue-Dev:
@Astrea49 All done with the next commit.

B1ue-Dev:
Closed due to conflict.

r-warfield:
interesting

Issue #691, opened 2022-03-31
[REQUEST] Implement ``client.get()`` helper method


sillyElvy:
### Describe the feature.

This helper method was meant to be created a while ago, but it seems to have been dropped, so I will implement it now.

### Code of Conduct

- [X] I agree to follow the contribution requirements.

Catalyst4222:
Finished as of #801

PR #693, opened 2022-03-31
feat!: implement ``bot.get()``  helper metod

sillyElvy:
## About

This pull request implements the long-time wanted helper method ``bot.get()`` 

## Checklist

  - (If existent): #691 
  - [ ] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix

### DIFF
--- a/interactions/client/bot.py
+++ b/interactions/client/bot.py
@@ -1080,7 +1080,7 @@ class Client:
         # TODO: custom error formatter
 
         if isinstance(obj, Channel):
-            if not channel_id:
+            if channel_id is MISSING:
                 raise AttributeError("Please specify a channel_id for getting a channel!")
 
             if channel := self._http.cache.channels.get(str(channel_id)) and cache:
@@ -1090,7 +1090,7 @@ class Client:
                 return Channel(**await self._http.get_channel(int(channel_id)), _client=self._http)
 
         elif isinstance(obj, Guild):
-            if not guild_id:
+            if guild_id is MISSING:
                 raise AttributeError("Please specify a guild_id for getting a guild!")
 
             if not (guild := self._http.cache.self_guilds.get(str(guild_id)) and cache):
@@ -1107,7 +1107,7 @@ class Client:
             return guild
 
         elif isinstance(obj, Member):
-            if not guild_id or not user_id:
+            if guild_id is MISSING or user_id is MISSING:
                 raise AttributeError("Please specify guild_id and user_id for getting a member!")
 
             # todo: cache impl
@@ -1117,7 +1117,7 @@ class Client:
             )
 
         elif isinstance(obj, Message):
-            if not channel_id or not message_id:
+            if channel_id is MISSING or message_id is MISSING:
                 raise AttributeError(
                     "Please specify message_id and channel_id for getting a message!"
                 )
@@ -1135,7 +1135,7 @@ class Client:
             return Message(**message, _client=self._http)
 
         elif isinstance(obj, Role):
-            if not guild_id or not role_id:
+            if guild_id is MISSING or role_id is MISSING:
                 raise AttributeError("Please specify guild_id and role_id for getting a role!")
 
             if role := self._http.cache.roles.get(str(role_id)) and cache:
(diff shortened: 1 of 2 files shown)


interactions/client/bot.py
```diff
@@ -1092,6 +1097,122 @@ def reload(
         self.remove(name, package)
         return self.load(name, package, *args, **kwargs)
 
+    async def get(
+        self,
+        obj: Union[Channel, Guild, Member, Message, Role, User],
+        *,
+        cache: bool = False,
+        channel_id: Optional[int] = MISSING,
```
V3ntus, line 1105: Instead of all these ID kwargs, how about just one?

V3ntus: Oh I see. Stuff like Message, Member, etc. requires two values

V3ntus: I don't like the clutter of all these keywords. For `Member`, `Message`, and `Role`, how about we expect a tuple or something?
```py
bot.get(Message, (guild_id, message_id))
```

sillyElvy: Do you think that's easier for the user?

V3ntus: Well, it's certainly cleaner and prevents confusion, in my opinion

sillyElvy: I see, I'll keep that in mind, but I'd like to know what fl0w wants, first

V3ntus: Definitely. I'd like others' input as well, including fl0w

mAxYoLo01: Personally I think it would work better with kwargs than a single argument. If it were a tuple and you need to give two IDs, do they need to always be in the same order so we are able to correctly parse them? `(guild_id, message_id)` VS `(message_id, guild_id)`

sillyElvy: That was my thought

AstreaTSS: I was about to suggest the same. If typehinting is an issue, then typing overloads can be used to do so in a pretty reasonable way.

V3ntus: > Personally I think it would work better with kwargs than a single argument. If it were a tuple and you need to give two IDs, do they need to always be in the same order so we are able to correctly parse them? `(guild_id, message_id)` VS `(message_id, guild_id)`

I think a hierarchical process should be considered. You need the guild to get the message so (guild, message). Would have to be stressed in documentation for users otherwise we'll have to troubleshoot nasty issues probably

AstreaTSS: > > Personally I think it would work better with kwargs than a single argument. If it were a tuple and you need to give two IDs, do they need to always be in the same order so we are able to correctly parse them? `(guild_id, message_id)` VS `(message_id, guild_id)`
> 
> I think a hierarchical process should be considered. You need the guild to get the message so (guild, message)

Honestly, I don't think it's *that* major - most people will put the guild ID before the message one, but if they don't... well, it's unorthodox, but it doesn't matter as long as they provide both.

mAxYoLo01: Seems easier to implement (and not less clean) then to use kwargs, don't you think?

AstreaTSS: > Seems easier to implement (and not less clean) then to use kwargs, don't you think?

...really? Tuples are ugly in this case IMO, and would be annoying to typehint (you would still have to use overloads to typehint per object, or at least typehint different based off of different objects). I'd argue it would take maybe a bit less time to typehint for not much benefit.

i0bs:
This approach superficially tries to parse an ID or equivalent field identifier into a data model by hardcoded conditionals, which is what we're wanting to avoid. This current design does not respect our philosophy of a "dynamic" retrieval system.

Unless you're comfortable with waiting, my suggestion would be closing this PR for the time being until I am ready to share information about how to do this at a more appropriate time, as we've more important things to do on our priority list (refer to 2022 roadmap) than a new feature implementing a utility method. This is an enhancement that is not a harsh requirement to be seen in the library thus far, so I believe it can wait.

PR #702, opened 2022-04-05
feat: implement `Client.get()` dynamically

### LINKED ISSUE #691: [REQUEST] Implement ``client.get()`` helper method

Toricane:
## About

This pull request implements the helper function `Client.get()`, which lets you dynamically and effortlessly get models from the Discord API.

You use this function by assigning a variable to the result of the function call.
The keyword arguments depend on what kind of model you want to get.
Input the keyword arguments as integers, or as the HTTP methods in the `HTTPClient` require.

Here are examples of how you can use this helper function:
```py
from interactions import Channel, Guild, Member, Message, Role, User, Client
...
client: Client = Client(...)
...
channel: Channel = await client.get(
    Channel,
    channel_id=...,
)
guild: Guild = await client.get(
    Guild,
    guild_id=...,
)
member: Member = await client.get(
    Member,
    guild_id=...,
    member_id=...,
)
message: Message = await client.get(
    Message,
    channel_id=...,
    message_id=...,
)
role: Role = await client.get(
    Role,
    guild_id=...,
    role_id=...,
)
user: User = await client.get(
    User,
    user_id=...,
)
```
## Checklist

  - (If existent): #691 
  - [ ] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix

### DIFF
--- a/interactions/client/bot.py
+++ b/interactions/client/bot.py
@@ -7,7 +7,7 @@ from importlib.util import resolve_name
 from inspect import getmembers
 from logging import Logger
 from types import ModuleType
-from typing import Any, Callable, Coroutine, Dict, List, Optional, Union
+from typing import Any, Callable, Coroutine, Dict, List, Optional, TypeVar, Union
 
 from ..api import Cache
 from ..api import Item as Build
@@ -18,6 +18,7 @@ from ..api.models.flags import Intents
 from ..api.models.guild import Guild
 from ..api.models.misc import MISSING, Snowflake
 from ..api.models.presence import ClientPresence
+from ..api.models.role import Role
 from ..api.models.team import Application
 from ..base import get_logger
 from .decor import command
@@ -26,6 +27,7 @@ from .enums import ApplicationCommandType, Locale, OptionType
 from .models.command import ApplicationCommand, Option
 from .models.component import Button, Modal, SelectMenu
 
+T = TypeVar("T")
 log: Logger = get_logger("client")
 _token: str = ""  # noqa
 _cache: Optional[Cache] = None
@@ -362,6 +364,55 @@ class Client:
         :type presence: ClientPresence
         """
         await self._websocket._update_presence(presence)
+
+    async def get(
+        self, model: T, http_method: Optional[str] = None, **kwargs: Dict[str, Snowflake]
+    ) -> T:
+        """
+        A helper function for getting a model from the Discord API.
+
+        :param model: The model to get.
+        :type model: object
+        :param http_method: The HTTP method to use instead of `get_x`.
+        :type http_method: Optional[str]
+        :param kwargs: The arguments to pass to the HTTP method.
+        :type kwargs: Dict[str, Snowflake]
+        :return: The model.
+        :rtype: object
+        """
+        if model is Role:
+            if not kwargs.get("guild_id") and kwargs.get("role_id"):
+                raise TypeError(
+                    "Client.get() missing some of the following required keyword arguments: guild_id, role_id."
+                )
+            try:
+                roles: List[dict] = await self._http.get_all_roles(int(kwargs["guild_id"]))
+            except TypeError as e:
+                raise ValueError("Invalid guild_id provided.") from e
+            role = next((_role for _role in roles if _role["id"] == str(kwargs["role_id"])), None)
+            if role is None:
+                raise ValueError("Role not found.")
+            return Role(**role, _client=self._http)
+        else:
+            class_name: str = model.__name__.removesuffix("s")
+            try:
+                _http_method: Coroutine = getattr(
+                    self._http, http_method or f"get_{class_name.lower()}"
+                )
+            except AttributeError as e:
+                raise ValueError(f"Client.get() does not support the model {class_name}.") from e
+            try:
+                res: Union[dict, List[dict]] = await _http_method(**kwargs)
+            except TypeError as e:
+                raise ValueError(
+                    f"Client.get() could not find a {class_name} with the given IDs."
+                ) from e
+            if isinstance(res, dict):
+                return model(**res, _client=self._http)
+            elif isinstance(res, list):
+                return [model(**r, _client=self._http) for r in res]
+            else:
+                return res
 
     def __check_command(
         self,
--- a/interactions/client/bot.py
+++ b/interactions/client/bot.py
@@ -368,15 +368,15 @@ class Client:
     async def get(
         self, model: T, http_method: Optional[str] = None, **kwargs: Dict[str, Snowflake]
     ) -> T:
-        """
+        r"""
         A helper function for getting a model from the Discord API.
 
         :param model: The model to get.
         :type model: object
         :param http_method: The HTTP method to use instead of `get_x`.
         :type http_method: Optional[str]
-        :param kwargs: The arguments to pass to the HTTP method.
-        :type kwargs: Dict[str, Snowflake]
+        :param \**kwargs: The arguments to pass to the HTTP method.
+        :type \**kwargs: Dict[str, Snowflake]
         :return: The model.
         :rtype: object
         """
--- a/interactions/client/bot.py
+++ b/interactions/client/bot.py
@@ -7,7 +7,7 @@ from importlib.util import resolve_name
 from inspect import getmembers
 from logging import Logger
 from types import ModuleType
-from typing import Any, Callable, Coroutine, Dict, List, Optional, TypeVar, Union
+from typing import Any, Callable, Coroutine, Dict, List, Optional, Union
 
 from ..api import Cache
 from ..api import Item as Build
@@ -27,7 +27,6 @@ from .enums import ApplicationCommandType, Locale, OptionType
 from .models.command import ApplicationCommand, Option
 from .models.component import Button, Modal, SelectMenu
 
-T = TypeVar("T")
 log: Logger = get_logger("client")
 _token: str = ""  # noqa
 _cache: Optional[Cache] = None
@@ -365,54 +364,52 @@ class Client:
         """
         await self._websocket._update_presence(presence)
 
-    async def get(
-        self, model: T, http_method: Optional[str] = None, **kwargs: Dict[str, Snowflake]
-    ) -> T:
+    async def get(self, __obj: object, http_method: Optional[str] = None, **kwargs) -> object:
         r"""
-        A helper function for getting a model from the Discord API.
+        A helper method for retrieving data from the Discord API in its object representation.
 
         :param model: The model to get.
         :type model: object
         :param http_method: The HTTP method to use instead of `get_x`.
         :type http_method: Optional[str]
         :param \**kwargs: The arguments to pass to the HTTP method.
-        :type \**kwargs: Dict[str, Snowflake]
-        :return: The model.
+        :type \**kwargs: dict
+        :return: The model we're trying to get.
         :rtype: object
         """
-        if model is Role:
-            if not kwargs.get("guild_id") and kwargs.get("role_id"):
+        if __obj is Role:
+            if not (kwargs.get("guild_id") and kwargs.get("role_id")):
                 raise TypeError(
                     "Client.get() missing some of the following required keyword arguments: guild_id, role_id."
                 )
             try:
                 roles: List[dict] = await self._http.get_all_roles(int(kwargs["guild_id"]))
-            except TypeError as e:
-                raise ValueError("Invalid guild_id provided.") from e
-            role = next((_role for _role in roles if _role["id"] == str(kwargs["role_id"])), None)
-            if role is None:
-                raise ValueError("Role not found.")
-            return Role(**role, _client=self._http)
-        else:
-            class_name: str = model.__name__.removesuffix("s")
+            except TypeError:
+                raise ValueError("Invalid guild_id provided.")
+            for role in roles:
+                if role["id"] == str(kwargs["role_id"]):
+                    return Role(**role, _client=self._http)
+            raise ValueError("Role not found.")
+        elif __obj is not List:
+            class_name: str = __obj.__name__.removesuffix("s")
             try:
                 _http_method: Coroutine = getattr(
                     self._http, http_method or f"get_{class_name.lower()}"
                 )
-            except AttributeError as e:
-                raise ValueError(f"Client.get() does not support the model {class_name}.") from e
+            except AttributeError:
+                raise ValueError(f"Client.get() does not support the model {class_name}.")
             try:
                 res: Union[dict, List[dict]] = await _http_method(**kwargs)
-            except TypeError as e:
-                raise ValueError(
-                    f"Client.get() could not find a {class_name} with the given IDs."
-                ) from e
+            except TypeError:
+                raise ValueError(f"Client.get() could not find a {class_name} with the given IDs.")
             if isinstance(res, dict):
-                return model(**res, _client=self._http)
+                return __obj(**res, _client=self._http)
             elif isinstance(res, list):
-                return [model(**r, _client=self._http) for r in res]
+                return [__obj(**r, _client=self._http) for r in res]
             else:
                 return res
+        else:
+            ...  # TODO: Implement List[object]
 
     def __check_command(
         self,
--- a/interactions/client/bot.py
+++ b/interactions/client/bot.py
@@ -391,7 +391,7 @@ class Client:
                     return Role(**role, _client=self._http)
             raise ValueError("Role not found.")
         elif __obj is not List:
-            class_name: str = __obj.__name__.removesuffix("s")
+            class_name: str = __obj.__name__
             try:
                 _http_method: Coroutine = getattr(
                     self._http, http_method or f"get_{class_name.lower()}"
--- a/interactions/client/bot.pyi
+++ b/interactions/client/bot.pyi
@@ -7,31 +7,22 @@ from typing import (
     Dict,
     List,
     Optional,
-    overload,
     Tuple,
-    Type,
-    TypeVar,
     Union,
 )
 
 from ..api.cache import Cache
 from ..api.gateway import WebSocketClient
 from ..api.http.client import HTTPClient
-from ..api.models.channel import Channel
 from ..api.models.flags import Intents
 from ..api.models.guild import Guild
-from ..api.models.member import Member
-from ..api.models.message import Message
 from ..api.models.misc import MISSING, Snowflake
 from ..api.models.presence import ClientPresence
-from ..api.models.role import Role
 from ..api.models.team import Application
-from ..api.models.user import User
 from .enums import ApplicationCommandType, Locale
 from .models.command import ApplicationCommand, Option
 from .models.component import Button, Modal, SelectMenu
 
-T = TypeVar("T")
 _token: str = ""  # noqa
 _cache: Optional[Cache] = None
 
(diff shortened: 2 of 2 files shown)


interactions/client/bot.py
```diff
@@ -363,6 +365,55 @@ async def change_presence(self, presence: ClientPresence) -> None:
         """
         await self._websocket._update_presence(presence)
 
+    async def get(
+        self, model: T, http_method: Optional[str] = None, **kwargs: Dict[str, Snowflake]
```
i0bs:
```suggestion
        self, model: T, http_method: Optional[str] = None, **kwargs
```

`**kwargs` should never have strict typing involved. If you're trying to do it for documentation purposes, please note it only as being dict-like.

Toricane:
Quick note: the caching system is not used as of right now in the method since it needs a refactor.

interactions/client/bot.py
```diff
@@ -363,6 +365,55 @@ async def change_presence(self, presence: ClientPresence) -> None:
...
+        :param model: The model to get.
+        :type model: object
+        :param http_method: The HTTP method to use instead of `get_x`.
+        :type http_method: Optional[str]
+        :param \**kwargs: The arguments to pass to the HTTP method.
+        :type \**kwargs: Dict[str, Snowflake]
+        :return: The model.
+        :rtype: object
+        """
+        if model is Role:
```
i0bs: Why are we making a check for what the object actually is?

interactions/client/bot.py
```diff
@@ -363,6 +365,55 @@ async def change_presence(self, presence: ClientPresence) -> None:
...
+            if not kwargs.get("guild_id") and kwargs.get("role_id"):
```
i0bs: This will not trigger if we're missing only one.

interactions/client/bot.py
```diff
@@ -363,6 +365,55 @@ async def change_presence(self, presence: ClientPresence) -> None:
...
+                raise TypeError(
+                    "Client.get() missing some of the following required keyword arguments: guild_id, role_id."
+                )
+            try:
+                roles: List[dict] = await self._http.get_all_roles(int(kwargs["guild_id"]))
+            except TypeError as e:
+                raise ValueError("Invalid guild_id provided.") from e
+            role = next((_role for _role in roles if _role["id"] == str(kwargs["role_id"])), None)
```
i0bs: Is there a specific reason for why we're treating the role selection as an iterable?

interactions/client/bot.py
```diff
@@ -363,6 +365,55 @@ async def change_presence(self, presence: ClientPresence) -> None:
...
+            if role is None:
+                raise ValueError("Role not found.")
+            return Role(**role, _client=self._http)
+        else:
+            class_name: str = model.__name__.removesuffix("s")
+            try:
+                _http_method: Coroutine = getattr(
```
i0bs, line 396: Since you're making a typing check with the object argument to begin with, this will contradict if we're attempting to input `List[object]` as this will not suffix "s" to the end of the attribute grabbing,

interactions/client/bot.py
```diff
@@ -363,6 +365,55 @@ async def change_presence(self, presence: ClientPresence) -> None:
...
+                    self._http, http_method or f"get_{class_name.lower()}"
+                )
+            except AttributeError as e:
+                raise ValueError(f"Client.get() does not support the model {class_name}.") from e
+            try:
+                res: Union[dict, List[dict]] = await _http_method(**kwargs)
+            except TypeError as e:
+                raise ValueError(
+                    f"Client.get() could not find a {class_name} with the given IDs."
+                ) from e
```
i0bs:
```suggestion
                )
```

Please raise only the exception and none more. Developers should be able to catch for either exception and not clump it as one. If this isn't possible, please make use of our current exception system, which I will note is bound to be refactored.

interactions/client/bot.py
```diff
@@ -363,6 +365,55 @@ async def change_presence(self, presence: ClientPresence) -> None:
         """
         await self._websocket._update_presence(presence)
 
+    async def get(
```
i0bs: I don't get the purpose of adding the `get()` method when there is no way to demand data out of our cache.

interactions/client/bot.pyi
```diff
@@ -51,12 +69,58 @@ class Client:
     async def wait_until_ready(self) -> None: ...
     def event(self, coro: Coroutine, name: Optional[str] = None) -> Callable[..., Any]: ...
     def change_presence(self, presence: ClientPresence) -> None: ...
+    @overload
```
i0bs: We do not need to have numerous `@overload` assignments in our interfacing scheme. This is excessive boilerplate that will make maintaining this later on down the line messier. We're essentially hardcoding type vars, which we're wanting to avoid. My recommendation is ditching `TypeVar` as while it's great for other employments, would not suit a low profile in the library.

i0bs: Additionally, having numerous assignments of this will bring back the problem of having to further maintain it with adding even *more* object types that Discord may potentially introduce in the future.

interactions/client/bot.pyi
```diff
@@ -67,23 +131,23 @@ class Client:
         options: Optional[List[Option]] = MISSING,
         default_permission: Optional[bool] = MISSING,
         name_localizations: Optional[Dict[Union[str, Locale], str]] = MISSING,
-        description_localizations: Optional[Dict[Union[str, Locale], str]]  = MISSING,
+        description_localizations: Optional[Dict[Union[str, Locale], str]] = MISSING,
```
i0bs, line 80:
```suggestion
```

Why are you including this in a feature PR? Please revert.

interactions/client/bot.pyi
```diff
@@ -67,23 +131,23 @@ class Client:
...
     ) -> Callable[..., Any]: ...
     def message_command(
         self,
         *,
         name: str,
         scope: Optional[Union[int, Guild, List[int], List[Guild]]] = MISSING,
         default_permission: Optional[bool] = MISSING,
-        name_localizations: Optional[Dict[Union[str, Locale], str]]  = MISSING,
+        name_localizations: Optional[Dict[Union[str, Locale], str]] = MISSING,
```
i0bs, line 88:
```suggestion
```

interactions/client/bot.pyi
```diff
@@ -67,23 +131,23 @@ class Client:
...
     ) -> Callable[..., Any]: ...
     def user_command(
         self,
         *,
         name: str,
         scope: Optional[Union[int, Guild, List[int], List[Guild]]] = MISSING,
         default_permission: Optional[bool] = MISSING,
-        name_localizations: Optional[Dict[Union[str, Locale], str]]  = MISSING,
+        name_localizations: Optional[Dict[Union[str, Locale], str]] = MISSING,
```
i0bs, line 96:
```suggestion
```

interactions/client/bot.pyi
```diff
@@ -51,12 +69,58 @@ class Client:
...
+    async def get(
+        self, model: T, http_method: Optional[str] = None, **kwargs: Dict[str, Snowflake]
+    ) -> T: ...
     def __check_command(
         self,
         command: ApplicationCommand,
         coro: Coroutine,
         regex: str = r"^[a-z0-9_-]{1,32}$",
-    )-> None: ...
+    ) -> None: ...
```
i0bs, line 69:
```suggestion
```

I don't see how this is relevant to the PR.

i0bs (changes requested):

interactions/client/bot.py
```diff
@@ -363,6 +365,55 @@ async def change_presence(self, presence: ClientPresence) -> None:
...
+    ) -> T:
+        r"""
+        A helper function for getting a model from the Discord API.
+
+        :param model: The model to get.
+        :type model: object
+        :param http_method: The HTTP method to use instead of `get_x`.
+        :type http_method: Optional[str]
+        :param \**kwargs: The arguments to pass to the HTTP method.
+        :type \**kwargs: Dict[str, Snowflake]
```
i0bs:
```suggestion
        :type \**kwargs: dict
```

interactions/client/bot.py
```diff
@@ -363,6 +365,55 @@ async def change_presence(self, presence: ClientPresence) -> None:
...
+        :return: The model.
```
i0bs:
```suggestion
        :return: The model we're trying to get.
```

This should have some clarification consistent to the rest of our documentation

interactions/client/bot.py
```diff
@@ -363,6 +365,55 @@ async def change_presence(self, presence: ClientPresence) -> None:
...
+        """
+        if model is Role:
+            if not kwargs.get("guild_id") and kwargs.get("role_id"):
+                raise TypeError(
+                    "Client.get() missing some of the following required keyword arguments: guild_id, role_id."
+                )
+            try:
+                roles: List[dict] = await self._http.get_all_roles(int(kwargs["guild_id"]))
+            except TypeError as e:
+                raise ValueError("Invalid guild_id provided.") from e
```
i0bs:
```suggestion
                raise ValueError("Invalid guild_id provided.")
```

interactions/client/bot.pyi
Toricane: It was pre-commit, as you can see, it was removing a single unnecessary space.

interactions/client/bot.py
```diff
@@ -363,6 +365,55 @@ async def change_presence(self, presence: ClientPresence) -> None:
...
+            try:
+                res: Union[dict, List[dict]] = await _http_method(**kwargs)
+            except TypeError as e:
+                raise ValueError(
+                    f"Client.get() could not find a {class_name} with the given IDs."
+                ) from e
+            if isinstance(res, dict):
+                return model(**res, _client=self._http)
+            elif isinstance(res, list):
+                return [model(**r, _client=self._http) for r in res]
```
i0bs: This should be consistent with http requests too. If you're allowing a list of results to be produced, a list of requests should be callable as well. In this case, the code needs to check for if a "list"-like http endpoint method exists, otherwise to set up an async queue.

i0bs (changes requested):

interactions/client/bot.pyi
Toricane: The above reply

interactions/client/bot.py
```diff
@@ -363,6 +365,55 @@ async def change_presence(self, presence: ClientPresence) -> None:
...
         await self._websocket._update_presence(presence)
 
+    async def get(
+        self, model: T, http_method: Optional[str] = None, **kwargs: Dict[str, Snowflake]
+    ) -> T:
+        r"""
+        A helper function for getting a model from the Discord API.
+
+        :param model: The model to get.
+        :type model: object
```
i0bs, line 372:
```suggestion
        :type model: :class:`typing.TypeVar`
```

Please have this to be consistent with the type annotation in the argument signature.

interactions/client/bot.pyi
Toricane: https://github.com/interactions-py/library/pull/702#discussion_r842232870

interactions/client/bot.pyi
Toricane: https://github.com/interactions-py/library/pull/702#discussion_r842232870

interactions/client/bot.py
Toricane: Oh I see, will remove

interactions/client/bot.py
```diff
@@ -363,6 +365,55 @@ async def change_presence(self, presence: ClientPresence) -> None:
         """
         await self._websocket._update_presence(presence)
 
+    async def get(
+        self, model: T, http_method: Optional[str] = None, **kwargs: Dict[str, Snowflake]
+    ) -> T:
+        r"""
+        A helper function for getting a model from the Discord API.
```
i0bs:
```suggestion
        A helper method for retrieving data from the Discord API in its object representation.
```

revised for clarity.

i0bs (changes requested):

interactions/client/bot.py
Toricane: So should I use an equality check (`==`) instead?

interactions/client/bot.pyi
```diff
@@ -51,12 +69,58 @@ class Client:
...
+    @overload
+    async def get(
+        self,
+        model: Type[User],
+        http_method: Optional[str] = None,
+        *,
+        user_id: Optional[Snowflake] = None,
+    ) -> User: ...
+    async def get(
+        self, model: T, http_method: Optional[str] = None, **kwargs: Dict[str, Snowflake]
```
i0bs:
```suggestion
        self, model: T, http_method: Optional[str] = None, **kwargs
```

interactions/client/bot.py
Toricane: Putting it in parentheses fixed the issue, thanks

interactions/client/bot.py
i0bs: How is this raisable if we're already setting a strict type on the argument in the signature, as well as checking for if the field exists or not?

i0bs (changes requested):

interactions/client/bot.py
Toricane: Ok I will use a for loop there instead

i0bs (changes requested):
The implementation of `client.get()` should respect the ability for the typing module to be used. Much like discord.py's implementation, we should be able to run code like this and expect sanitized results:


```py
import typing

guilds: list[Guild] = await client.get(typing.List[interactions.Guild], id=X)

for guild in guilds:
    print(guild.name)
```

Toricane:
Okay, I will implement that.
For now, I'll convert this to a draft, and reopen it when that is finished.

i0bs:
I'll be crossposting this message I wrote into our #unstable channel in the support server last night to this PR in case you may have missed it, clarifying the state of your PR. Please consider that your PR may take a long time before it is appropriately introduced to the dev branch because of the reasons provided below.

image

i0bs:
> Okay, I will implement that. For now, I'll convert this to a draft, and reopen it when that is finished.

If you're to move this into a draft state, please consider what's been also written: https://github.com/interactions-py/library/pull/693#issuecomment-1086928456

There are chances that, if a PR is to be left in a draft state for a long amount of time without notable progress towards working on things being requested and/or required, may end up being closed in favor of lowering the workload for the core development team.

interactions/client/bot.py
Toricane: Removed them

interactions/client/bot.py
Toricane: It still simplifies getting models from the API for the end user.

interactions/client/bot.pyi
Toricane: Removing them

interactions/client/bot.py
Toricane: Done

interactions/client/bot.py
Toricane: Done

interactions/client/bot.py
Toricane: Removed

interactions/client/bot.py
i0bs: We shouldn't be checking the type at all. Refer to screenshot attached https://github.com/interactions-py/library/pull/702#issuecomment-1088147432 for more details.

interactions/client/bot.py
Toricane: Done

interactions/client/bot.py
Toricane: Removed the TypeVar due to one of your other suggestions.

interactions/client/bot.py
Toricane: There is no `get_role`, that's why `Role` is the only outlier

i0bs: My recommendation would be making a PR associated with implementing that feature

interactions/client/bot.py
Toricane: The code does not repeatedly request, it just requests once.
The only reason it is there is because of the `http_method` argument, which lets you put a string, such as `"get_self_guilds"` or other methods, which *returns* a list of results.

interactions/client/bot.py
Toricane: Okay thanks.

interactions/client/bot.py
Toricane: The `guild_id` could be incorrect, an example is `1234567890`.
Therefore, it would raise an error.

interactions/client/bot.pyi
Toricane: Removed

interactions/client/bot.py
i0bs: This comment refers to https://github.com/interactions-py/library/pull/702#pullrequestreview-931088201 and https://github.com/interactions-py/library/pull/702#discussion_r842230296

Toricane: Oh, I see.

interactions/client/bot.py
i0bs: Not sure why you resolved this conversation 😅 I'll be re-opening it to continue discussion on this part.

Grabbing from cache is not a suggestion, it's an absolute requirement of the `get()` method. We want this because we want to make as few HTTP calls as possible, which could easily help us hit rate limits depending on how bot developers structure their callbacks for this method. The reason why we've been postponing this for so long is because cache needs a refactor in not only the way we're storing data, but the way we *update* the cache itself. In theory, the future of this library will involve us heavily interacting with the cache and making rare HTTP requests unless absolutely needed.

Because of this, it needs to be possible for a bot developer to save themselves the hassle of potential rate limits by reading from the cache first before making an HTTP request. We do this especially with our application command synchronisation process to avoid excessive HTTP requests leading to rate limits, an effective measure that's been tried-and-tested.

interactions/client/bot.py
i0bs: > such as "get_self_guilds" or other methods, which returns a list of results.

You're not actually utilizing the http method feeder variable in this method impl itself to check for a suffix of "s," in-fact, you're actually removing it from the equation. This could be very harmful towards bot developers who want to make only 1 HTTP call instead of many.

interactions/client/bot.py
AstreaTSS: Leaching onto this - isn't `str.removesuffix` only a thing from Python 3.9+? This code won't work on 3.8.

As for the `list` problem, using `typing.get_origin` and working from there is a way around it. `Typing.List` gets resolved as `list` when using `typing.get_origin` too, so that's nice.

Though obviously you'll need to still write logic dealing with the origin if there is one, as well as dealing with "list" endpoints and the like.

interactions/client/bot.py
Toricane: No, I'm removing the "s" from the model.
Well I can remove that tbh
the only use is for `interactions.Webhooks` lmao

i0bs: There's no reason to be removing anything from the model names unless you're trying to parse a GW event name in its raw form.

interactions/client/bot.py
Toricane: Okay, thank you. I think ill just remove the `.removesuffix("s")` entirely.

interactions/client/bot.py
Toricane: There is a `get_webhook` but no `get_webhooks`.
The "s" at the end was the problem.
I think I'll just remove the `.removesuffix("x")`.

interactions/client/bot.py
Toricane: Yes I understand, I just haven't utilized the cache since it needs a refactor.
I will keep this as a draft until 4.1.1 is released and caching is implemented.
I want to continue work on your suggestions till then.

Toricane:
Okay, closing this and adding to `enhanced` for now.
May reopen later on.

Issue #717, opened 2022-04-08
[BUG] Inserting a string value as event name into ``@event``  is not possible


sillyElvy:
### Describe the bug.

Previously it was possible to do so, now it will result in an attribute error

### List the steps.

-

### What you expected.

It to be working and registering the event properly

### What you saw.

Instead, I received this traceback error given from my Python terminal:
```
Traceback (most recent call last):
  File "A:\Amogus\Python\interactions-bot\main.py", line 6
    raise SussyBaka("your code doesn't work, silly!")
          ^

SussyBaka: your code doesn't work, silly!
```

### What version of the library did you use?

unstable

### Code of Conduct

- [X] I agree to follow the contribution requirements.

i0bs:
please tell me you're not actually having `SussyBaka` as an error in your issue

sillyElvy:
> please tell me you're not actually having `SussyBaka` as an error in your issue

I do, I modified the whole language to exactly produce that error /s

No I just forgot to update it

PR #718, opened 2022-04-08
fix!: Allow `@event` usage with a name argument

sillyElvy:
This PR allows inserting a name into the event decorator.

## Checklist

  - (If existent): #717 
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix


i0bs (commented):
Do we _really_ need the coroutine to be optional? We'll always expect it to be there when registering an event for dispatch.

sillyElvy:
> Do we _really_ need the coroutine to be optional? We'll always expect it to be there when registering an event for dispatch.

It is always there. It just has to be optional, so you can do `@bot.event`. Keeping it not optional locks the name arg, removing the coro argument forces you to always do `@bot.event()`.

No matter how the decorator is used, the coro will always be present. Keeping it optional just improves UX

i0bs (approved):

Issue #738, opened 2022-04-18
[BUG] KeyError, Bot cannot start


ghost:
### Describe the bug.

interactions.py prints a KeyError just by starting it

### List the steps.

I tested it in multiple environments, but I can't launch only this bot (other bots work fine).
This issue is seen on 4.1.1 Beta 2 and 4.1.1 RC1 and not on 4.1.0 and 4.1.1 Beta 1

Console Log↓
```py
import interactions
bot = interactions.Client(token="")
bot.start()
```

### What you expected.

I hope Bot starts.

### What you saw.

Instead, I received this traceback error given from my Python terminal:
```py
>>> import interactions
>>> bot = interactions.Client(token="", disable_sync=False)
>>> bot.start()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/xs905136/miniconda3/lib/python3.9/site-packages/interactions/client/bot.py", line 115, in start
    self._loop.run_until_complete(self._ready())
  File "/home/xs905136/miniconda3/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/home/xs905136/miniconda3/lib/python3.9/site-packages/interactions/client/bot.py", line 326, in _ready
    await self._login()
  File "/home/xs905136/miniconda3/lib/python3.9/site-packages/interactions/client/bot.py", line 331, in _login
    await self._websocket._establish_connection(self._shard, self._presence)
  File "/home/xs905136/miniconda3/lib/python3.9/site-packages/interactions/api/gateway/client.py", line 189, in _establish_connection
    await self._handle_connection(stream, shard, presence)
  File "/home/xs905136/miniconda3/lib/python3.9/site-packages/interactions/api/gateway/client.py", line 257, in _handle_connection
    self._dispatch_event(event, data)
  File "/home/xs905136/miniconda3/lib/python3.9/site-packages/interactions/api/gateway/client.py", line 353, in _dispatch_event
    self._dispatch.dispatch(f"on_{name}", __obj(**data))  # noqa
  File "/home/xs905136/miniconda3/lib/python3.9/site-packages/interactions/api/models/guild.py", line 325, in __init__
    and self._client.cache.self_guilds.values[str(self.id)].members
KeyError: '889041902773604362'
>>>
```

### What version of the library did you use?

stable

### Version specification

4.1.1 Beta2 and 4.1.1 RC1

### Code of Conduct

- [X] I agree to follow the contribution requirements.

[...]

PR #739, opened 2022-04-18
fix: properly check for guild existence in cache

AstreaTSS:
## About

When filling out `Guild.members`, `Guild` did a check to see if it has the members it got from `GUILD_CREATE` - since guilds only have their `members` property filled out when they are "created", this allows for the same guild gotten after the fact to have those guild members.

However, to detect if the guild being added was gotten by a `GUILD_CREATE`, the library resorts to a rather naive approached that simply checked if *any* guild was cached from `GUILD_CREATE` already. This means if the current guild wasn't cached, the code would continue forward and run into a `KeyError` when actually trying to get said guild from the cache.

This PR fixes that by checking if the guild *in question* is cached, not just any guild.

## Checklist

  - (If existent): #738
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix

### DIFF
--- a/interactions/api/models/guild.py
+++ b/interactions/api/models/guild.py
@@ -322,7 +322,7 @@ class Guild(DictSerializerMixin):
         if (
             not self.members
             and self._client
-            and len(self._client.cache.self_guilds.view) > 1
+            and self._client.cache.self_guilds.values.get(str(self.id))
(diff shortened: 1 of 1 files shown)


ghost (approved):
Confirmed that KeyError is improved

sillyElvy (approved):

Issue #746, opened 2022-04-21
[REQUEST] implement name property for member object


sillyElvy:
### Describe the feature.

This makes the getting of the name of a member easier

### Code of Conduct

- [X] I agree to follow the contribution requirements.

PR #747, opened 2022-04-21
feat: implement name property for the member object

sillyElvy:
## Checklist

  - (If existent): #746 
  - [ ] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix


interactions/api/models/member.py
```diff
@@ -83,7 +83,7 @@ def __init__(self, **kwargs):
             self.avatar = self.user.avatar
 
     def __repr__(self) -> str:
-        return self.user.username if self.user else self.nick
+        return self.nick or (self.user.username if self.user else None)
```
Toricane: Can't you just do `return self.name` here?

interactions/api/models/member.py
```diff
@@ -105,6 +105,16 @@ def mention(self) -> str:
         """
         return f"<@!{self.user.id}>" if self.nick else f"<@{self.user.id}>"
 
+    @property
+    def name(self) -> str:
+        """
+        Returns a string of whether the users nickname or its username, if a nickname isn't present
```
Toricane: I think this should be:
`Returns the string of either the user's nickname or username.`
The one you had was confusing and with some grammar mistakes.
BTW you were also missing a period at the end.

Toricane (changes requested):
Looks good, just a couple of things.

interactions/api/models/member.py
sillyElvy: Does that exist?

interactions/api/models/member.py
sillyElvy: I see the grammar mistake now lol

I had like 5 minutes of my computer time left and wanted to finish it :lol:

sillyElvy:
```suggestion
        returns the string of either the user's nickname or username
```

testing if this works, I deleted the fork..

interactions/api/models/member.py
```diff
@@ -105,6 +105,16 @@ def mention(self) -> str:
         """
         return f"<@!{self.user.id}>" if self.nick else f"<@{self.user.id}>"
 
+    @property
+    def name(self) -> str:
+        """
+        returns the string of either the user's nickname or username
```
sillyElvy:
```suggestion
        Returns the string of either the user's nickname or username.
```

interactions/api/models/member.py
Toricane: > Does that exist?

You yourself implemented it in this PR.

sillyElvy: OH GOD I'M STUPID

sillyElvy:
```suggestion
        return self.name
```

Toricane: To clarify even more, this line is the same as your `name` property that you implemented, so why not just do:
```py
def __repr__(self) -> str:
    return self.name
```

sillyElvy: Yeah I got it

Toricane (approved):
LGTM!

PR #750, opened 2022-04-22
feat: implement `get_role` and `get_emoji` HTTP methods

### LINKED ISSUE #749: [REQUEST] Add `get_role` and `get_emoji` HTTP methods
### Describe the feature.

All of the other models have a `get_x()` HTTP method implemented, so for consistency, this should be implemented.
These methods would also be useful for a future `get()` dynamic helper method because of the improved consistency.
Also [this](https://github.com/interactions-py/library/pull/702#discussion_r842243753).

### Code of Conduct

Toricane:
## About

This pull request implements `get_role` and `get_emoji` HTTP methods for consistency and for future use with a `Client.get()` dynamic helper method.
See the linked issue below for more info.

## Checklist

  - (If existent): #749 
  - [ ] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix


interactions/api/http/guild.py
```diff
@@ -385,6 +385,20 @@ async def get_all_roles(self, guild_id: int) -> List[dict]:
...
+        :param guild_id: Guild ID snowflake
+        :param role_id: Role ID snowflake
+        :return: A Role object as a dictionary.
+        """
+        request = await self.get_all_roles(guild_id)
+
+        for role in request:
+            if role.get("id") and int(role["id"]) == role_id:
+                return role
+
```
AstreaTSS: This... this I feel iffy about.

There's clearly cases where some HTTP methods do some pre-processing (`get_user` comes to mind immediately), but from what I can see, there's *never* post-processing. At most, the client may cache the result, but I've never seen it explicitly filter things out like this. There's also no precedent for any other similar library to do something like this, either.

Worst of all, this doesn't even represent a valid Discord endpoint. While it would be admittedly nice to have a `Get Role` method, Discord explicitly doesn't have it, and we shouldn't be giving users the wrong expectation.
(You can argue aliases for HTTP methods give users the wrong expectations, but at least it's to a much lesser extent.)

This *would* make it easier for the future `bot.get`, and I recognize that, but I personally don't like this as a "solution" to that problem

interactions/api/http/emoji.py
```diff
@@ -32,6 +32,16 @@ async def get_guild_emoji(self, guild_id: int, emoji_id: int) -> dict:
         """
         return await self._req.request(Route("GET", f"/guilds/{guild_id}/emojis/{emoji_id}"))
 
+    async def get_emoji(self, guild_id: int, emoji_id: int) -> dict:
+        """
+        Gets an emote from a guild.
```
AstreaTSS: It should be mentioned that this is an alias to `get_guild_emoji`, like how [`get_self`](https://github.com/interactions-py/library/blob/1d3ec28aa5d8bd1b8cc6c52c15960e118b6a8cc8/interactions/api/http/user.py#L20) does.

Though, honestly, I do question why `get_self` exists in the first place. It's weird and unnecessary. This alias *may* be, too.

interactions/api/http/guild.py
AstreaTSS: Huh, fl0w/James *has* stated this would be fine, [as seen here.](https://github.com/interactions-py/library/pull/702#discussion_r842243753)

I disagree personally, but that's something to consider ~~and my gut feeling that flow might have is why i didnt explicitly deny this pr~~.

interactions/api/http/emoji.py
Toricane: This was mainly for consistency and for the future `get()` method since roles and emojis were the only outliers.
I will mention that it is an alias.

interactions/api/http/guild.py
Toricane: The previous solution (that my previous get method involved) was to make a conditional to check if it is a `Role` and do it manually there, which is *not* dynamic.
Resolving this outlier simplifies the `get()` method and improves consistency.

I don't know why Discord doesn't have a Get Role endpoint, so I had to use post-processing.
I don't know of a better solution than this, please let me know how you would go about this.
I have modified the description to `Gets the specified role from the list of roles in a Guild.` Should I also add that it isn't an endpoint to further clarify what it is?

AstreaTSS: I *do* have an idea in mind, but it was shot down earlier. I may re-mention it again sooner or later, but probably not here. If James/fl0w wanted it, I can't argue about it too much.

Regardless, I do think it's worth clarifying it's not a native endpoint, yeah.

Toricane: Okay, will add, thanks

sillyElvy (commented):
This is a violation of the discord api scheme

Issue #751, opened 2022-04-22
[REQUEST] Refactor some parts of the interactions.api.gateway package for readability


lilyyy411:
## Reasoning
Some of the attributes and methods for the `WebSocketClient` class have kind of misleading names and documentation. Specifically, `WebSocketClient.__receive_packet_stream`, despite its name is not actually a packet stream; it is just a packet. 

Also, the `_Heartbeat`'s `event` field is unexpectedly only used for setting and clearing and checking whether the event is set, which calls for replacement with a `bool`. Using the event in this way is very unintuitive for people that look at the library's source code. In fact, the entire `_Heartbeat` class is unneeded and doesn't really contribute anything. All functionality contained in the `_Heartbeat` class can be recreated in an arguably more readable way by adding 2 fields to `WebClientSocket`.

## Proposed Action:
- `WebSocketClient.__receive_packet_stream` should no longer be a property; it should be a method that passes no arguments, and should be a renamed to `WebSocketClient.__get_next_packet`
- Remove `WebSocketClient.__heartbeat` in favor of a more readable option
- `WebSocketClient.__task` should be renamed to `WebSocketClient.__heartbeat_task` for more clarity on what the task does
- Add a field named `._heartbeat_delay` to `WebSocketClient` to take the place of `WebSocketClient.__heartbeat.delay`
- Replace every time `WebSocketClient.__heartbeat.event` is referenced with proper getting / setting of `WebSocketClient.__received_heartbeat_ack`, which takes its place.
- Delete the `interactions.api.gateway.heartbeat` module as it would no longer be needed 

If I made a mistake, let me know because it's almost midnight and I'm kind of braindead, and this is my first time contributing to a relatively large project

### Code of Conduct

- [X] I agree to follow the contribution requirements.

PR #752, opened 2022-04-22
refactor: Refactor some attributes on the WebClientSocket for readability

lilyyy411:
## About

This pull request is about issue #751, which proposes some refactoring related to the `interactions.api.gateway` package

## Checklist

  - (If existent):
  - [ ] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix


interactions/api/gateway/client.py
```diff
@@ -507,12 +505,10 @@ def __option_type_context(self, context: object, type: int) -> dict:
     async def restart(self):
         await self.__restart()
 
-    @property
-    async def __receive_packet_stream(self) -> Optional[Dict[str, Any]]:
+    async def __get_next_packet(self) -> Union[Dict[str, Any], WSMessage, None]:
```
sillyElvy, line 508: It was meant to be a property, please keep it as that

interactions/api/gateway/client.pyi
```diff
@@ -60,13 +60,12 @@ class WebSocketClient:
...
-    def __contextualize(self, data: dict) -> object: ...
+    def __contextualize(self, data: dict) -> Any: ...
     def __sub_command_context(
         self, data: Union[dict, Option], _context: Optional[object] = MISSING
     ) -> Union[Tuple[str], dict]: ...
-    def __option_type_context(self, context: object, type: int) -> dict: ...
-    @property
-    async def __receive_packet_stream(self) -> Optional[Dict[str, Any]]: ...
+    def __option_type_context(self, context: Any, type: int) -> dict: ...
+    async def __get_next_packet(self) -> Union[Dict[str, Any], WSMessage, None]: ...
```
sillyElvy, line 68: It's a property

interactions/api/gateway/client.pyi
```diff
@@ -6,9 +6,8 @@ from asyncio import (
 from logging import Logger
 from typing import Any, Dict, List, Optional, Tuple, Union, Iterable
 
-from aiohttp import ClientWebSocketResponse
+from aiohttp import ClientWebSocketResponse, WSMessage
 
-from .heartbeat import _Heartbeat
```
sillyElvy, line 11: I'm against the removal of the heartbeat class, please revert this changes

interactions/api/gateway/client.pyi
```diff
@@ -60,13 +60,12 @@ class WebSocketClient:
...
     def _dispatch_event(self, event: str, data: dict) -> None: ...
-    def __contextualize(self, data: dict) -> object: ...
+    def __contextualize(self, data: dict) -> Any: ...
     def __sub_command_context(
         self, data: Union[dict, Option], _context: Optional[object] = MISSING
     ) -> Union[Tuple[str], dict]: ...
-    def __option_type_context(self, context: object, type: int) -> dict: ...
-    @property
-    async def __receive_packet_stream(self) -> Optional[Dict[str, Any]]: ...
+    def __option_type_context(self, context: Any, type: int) -> dict: ...
```
sillyElvy, line 67: Why are you changing the typehint?

sillyElvy (changes requested):

interactions/api/gateway/client.pyi
FayeDel: ??
What's the explanation behind this comment?

sillyElvy: I don't see the reason for a removal, imo it is an unnecessary change

AstreaTSS: I mean... removing an extra, unnecessary class makes sense if possible, and it seems like all the class did was add overhead to the gateway.

sillyElvy: If you're both fine with removing it we can do it

interactions/api/gateway/client.py
AstreaTSS: A property being asynchronous is... odd, to be fair. I can see why it was changed to be a function.

I'm not *against* keeping it as a property per se, but what was the reason for making it a property?

sillyElvy: uh.. @jameswalston? ..

PR #753, opened 2022-04-22
fix!: fix possible bug with event names for extensions

sillyElvy:
## About
Fixes a bug that was already fixed for the normal event decorator but not for ext listeners so it could possibly occur there

## Checklist

  - (If existent):
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix

### DIFF
--- a/interactions/client/bot.py
+++ b/interactions/client/bot.py
@@ -1340,7 +1340,7 @@ def extension_listener(func: Optional[Coroutine] = None, name: Optional[str] = N
 
     if func:
         # allows omitting `()` on `@listener`
-        func.__listener_name__ = func.__name__ if not name else name
+        func.__listener_name__ = name or func.__name__
         return func
 
     return decorator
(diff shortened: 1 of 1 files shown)


interactions/client/bot.py
```diff
@@ -1340,7 +1340,7 @@ def decorator(func: Coroutine):
 
     if func:
         # allows omitting `()` on `@listener`
-        func.__listener_name__ = func.__name__
+        func.__listener_name__ = func.__name__ if not name else name
```
```suggestion
        func.__listener_name__ = name or func.__name__
```

Toricane (changes requested):
One minor change

interactions/client/bot.py
sillyElvy: Suggestion: make it a suggestion so I can just commit it

Toricane: How do you do that?

sillyElvy: On deskotp its a file symbol with ± inside, when you select one or multiple lines, on mobile its just ±

Toricane: Done

Toricane (approved):
LGTM!

Issue #755, opened 2022-04-23
[BUG] `ctx.populate` for autocomplete does not send proper empty data


V3ntus:
### Describe the bug.

When responding with `ctx.populate(choices=[])`, instead of getting `No options match your search`, Discord considers the interaction as failed and responds with `Loading options failed`

### List the steps.

n/a

### What you expected.

n/a

### What you saw.

n/a

### What version of the library did you use?

stable

### Code of Conduct

- [X] I agree to follow the contribution requirements.

PR #756, opened 2022-04-23
fix: send a proper empty response when populating autocomplete with no choices

V3ntus:
## About

Instead of sending an empty list, which Discord will not accept, send an empty field for `choices`

```py
...
	await ctx.populate(choices=[])
	# or
	await ctx.populate(choices=None)
	# will now result in Discord sending "No options match your search"
	# instead of "Loading options failed" previously
```

## Checklist

  - (If existent): #755 
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix


interactions/client/context.py
```diff
@@ -518,32 +518,37 @@ async def populate(self, choices: Union[Choice, List[Choice]]) -> List[Choice]:
...
-                    )
-
-                await self.client.create_interaction_response(
-                    token=self.token,
-                    application_id=int(self.id),
-                    data={
-                        "type": InteractionCallbackType.APPLICATION_COMMAND_AUTOCOMPLETE_RESULT.value,
-                        "data": {"choices": _choices},
-                    },
+            log.debug("Populating choices...")
```
sillyElvy: Do we really need that?

V3ntus: Oops lemme remove my debug logs

sillyElvy (approved):

PR #757, opened 2022-04-23
feat: implement `get()` helper method.

sillyElvy:
## About

This PR tries to implement a get helper method.

## Checklist

  - (If existent):
  - [ ] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix


sillyElvy:
Marking as draft since waiting for the cache...

interactions/client/bot.py
```diff
@@ -1134,6 +1139,59 @@ async def modify(
...
+        ...
+
+    @overload
+    async def get(self, obj: Type[User], *, user_id: int) -> User:
+        ...
+
+    async def get(self, obj: Type[_T], **kwargs) -> _T:
+        """
+        Gets an object from discord.
+        """
```
Toricane: Probably change the docstring to this or something similar:
```py
r"""
A helper method for retrieving data from the Discord API in its object representation.

:param obj: The object to get.
:type obj: object
:param \**kwargs: The arguments to pass to the HTTP method.
:type \**kwargs: dict
:return: The object we're trying to get.
:rtype: object
"""
```

interactions/client/bot.py
```diff
@@ -1134,6 +1139,59 @@ async def modify(
...
+
+        if obj not in (Role, Emoji):
+            _func = getattr(self._http, f"get_{obj.__name__.lower()}")
+            _obj = await _func(**kwargs)
+            return obj(**_obj, _client=self._http)
+
+        else:
+            _guild = Guild(**await self._http.get_guild(kwargs.pop("guild_id")), _client=self._http)
+            _func = getattr(_guild, f"get_{obj.__name__.lower()}")
+            return await _func(**kwargs)
```
Toricane: Inverse condition:
```py
if obj in (Role, Emoji):
    _guild = Guild(**await self._http.get_guild(kwargs.pop("guild_id")), _client=self._http)
    _func = getattr(_guild, f"get_{obj.__name__.lower()}")
    return await _func(**kwargs)
else:
    _func = getattr(self._http, f"get_{obj.__name__.lower()}")
    _obj = await _func(**kwargs)
    return obj(**_obj, _client=self._http)
```

Toricane (changes requested):
I'm late but here are some changes

interactions/client/bot.py
```diff
@@ -1416,6 +1421,33 @@ async def modify(
...
+        """
+        A helper method for retrieving data from the Discord API in its object representation.
+
+        :param obj: The object to get. Should be a class object (not an instance!). For example: `interactions.Channel`.
+        :type obj: object
+        :param **kwargs: The arguments to pass to the HTTP method.
+        :type **kwargs: dict
+        :return: The object we're trying to get.
+        :rtype: object
+        """
```
```suggestion
        r"""
        A helper method for retrieving data from the Discord API in its object representation.

        :param obj: The object to get. Should be a class object (not an instance!). For example: `interactions.Channel`.
        :type obj: object
        :param \**kwargs: The arguments to pass to the HTTP method.
        :type \**kwargs: dict
        :return: The object we're trying to get.
        :rtype: object
        """
```

interactions/client/bot.py
```diff
@@ -1416,6 +1421,33 @@ async def modify(
...
+
+        if isinstance(obj, (Channel, Emoji, Guild, Member, Message, Role, Sticker, User)):
```
```suggestion
        if not isinstance(obj, type):
```

interactions/client/bot.py
```diff
@@ -1416,6 +1421,33 @@ async def modify(
...
+
+        if obj in (Role, Emoji):
+            _guild = Guild(**await self._http.get_guild(kwargs.pop("guild_id")), _client=self._http)
+            _func = getattr(_guild, _name)
+            return await _func(**kwargs)
+
+        else:
+            _func = getattr(self._http, _name)
+            _obj = await _func(**kwargs)
+            return obj(**_obj, _client=self._http)
```
Toricane: You could technically remove the else condition if you want:
```suggestion
        _func = getattr(self._http, _name)
        _obj = await _func(**kwargs)
        return obj(**_obj, _client=self._http)
```

Toricane (changes requested):
Docstring and minor code improvements

interactions/client/bot.py
Toricane: This checks if something is a class and not an instance [here is link](https://stackoverflow.com/questions/18499614/check-if-class-object)

i0bs (commented):
I've brought this up before both internally and in the support server, but doing `@overload` will eventually create a messy interface file. I think for type hinting purposes, the best clause is documenting `_T` to be a type of any data model that is recognised as a "resource" from the API.

interactions/client/bot.py
Catalyst4222: That would do the opposite effect, where passing a class would raise the error instead of when an instance is passed. I do think that not hardcoding objects would be beneficial, though `inspect.isclass` would probably be the best way to do it

interactions/client/bot.py
sillyElvy: Flake 8 complained with the `\` before tho

sillyElvy:
> I've brought this up before both internally and in the support server, but doing `@overload` will eventually create a messy interface file. I think for type hinting purposes, the best clause is documenting `_T` to be a type of any data model that is recognised as a "resource" from the API.

But how should we tell the dev what attributes he needs?

AstreaTSS:
I don't think it's fully realistic to *completely* get rid of overloads, but I have an idea:

- One overload can have a bounded `TypeVar` that is bounded to the classes that need 2 attributes, and that class can typehint the two attributes needed, possibly `guild_id=..., id=...`.
- Another can be a normal `TypeVar` that only has one attribute, like `id=...`.

interactions/client/bot.py
Toricane: Oops forgot to add `not`

interactions/client/bot.py
Toricane: That's how you do it for the docstring if you have **kwargs, it's also elsewhere, you can check the others out

sillyElvy: I'll merge it and see if Flake8 complains again

sillyElvy:
I messed up the branch (again), I'll make a new PR

PR #762, opened 2022-04-26
refactor!: optimize sync behavior

### LINKED ISSUE #667: [BUG] The Unstable Syncing of Guild Commands
### Describe the bug.

When the bot is started with `disable_sync=False` the guild commands were not synced correctly for example :
```py
@bot.command(
    scope=["guild1_id", "guild2_id"],
    name="test1",
    description="Test1",
)
async def math(self, ctx: ComponentContext):
    return await ctx.send("test1")


@bot.command(
    scope=["guild1_id", "guild2_id"],
    name="test2",
    description="Test2",)
async def math(self, ctx: ComponentContext):
    await ctx.defer()
    await ctx.send("Test2")
```
guild 1 will have test1 and test 2 
but 
guild 2 will only have test 2
This also happens inside extensions

### List the steps.

1. Import module
2. Create guild command
3. Start Bot
4. Wait till the bot started completely
5. Go to Guild1 and Guild 2 to see the slash commands by trying to use them

### What you expected.

Bith Guild1 and Guild 2 will have test1 and test 2

### What you saw.

guild 1 will have test1 and test 2 
but 
guild 2 will only have test 2

stable

### Version specification

_No response_

### Code of Conduct

### LINKED ISSUE #689: [BUG] Autocomplete fails when disable_sync=True
### Describe the bug.

Same problem than in #582, but outside of an extension. The error message is different though.

### List the steps.

1. Create a command
```py
@bot.command(name="load", description="Load all cogs or a specific one", scope=server, options=[
    interactions.Option(type=interactions.OptionType.STRING, name="cog", description="Cog to load", required=False, autocomplete=True)
])
async def load(ctx: interactions.CommandContext, cog: str = None):
```

2. Create an autocomplete callback to this command
```py
@bot.autocomplete("load", "cog")
async def load_autocomplete(ctx: interactions.CommandContext, user_input: str = ""):
```

### What you expected.

Autocomplete callback registered

### What you saw.

```
File "C:\Users\.\launch.py", line 68, in <module>
    @bot.autocomplete("load", "cog")
  File "C:\Users\.\interactions\client.py", line 861, in autocomplete
    _command_obj = self._find_command(_application_commands, command)
  File "C:\Users\.\interactions\client.py", line 812, in _find_command
    raise InteractionException(
interactions.api.error.InteractionException: Command check has failed. The command does not exist. Make sure to define your autocomplete callback after your commands Error code: 6
```

unstable

### Version specification

_No response_

### Code of Conduct

sillyElvy:
## About

This pull request puts the whole command synchronization process into one function, where it checks all commands and bulks all commands at once instead of syncing every command itself. Ideally, this should reduce rate limits. This also precisely allows for a command to be removed in any condition, even if the main file is empty.

Further testing of this would be nice, Although I have tested it, I'd feel better if you could test it again for some edge-cases and if sync works for commands on multiple guilds.

## Checklist

  - (If existent): #689 #667
  - [ ] Documentation
  - [x] Breaking change
  - [x] New feature/enhancement
  - [x] Bugfix


interactions/api/http/interaction.py
```diff
@@ -15,22 +15,34 @@ def __init__(self) -> None:
...
+        application_id: Union[int, Snowflake],
+        guild_id: Optional[int] = None,
+        with_localizations: Optional[bool] = None,
     ) -> List[dict]:
         """
         Get all application commands from an application.
 
         :param application_id: Application ID snowflake
         :param guild_id: Guild to get commands from, if specified. Defaults to global (None)
+        :param with_localizations: 	Whether to include full localization dictionaries (name_localizations and description_localizations) in the returned objects, instead of the name_localized and description_localized fields. Default false.
```
Toricane:
```suggestion
        :param with_localizations: Whether to include full localization dictionaries (name_localizations and description_localizations) in the returned objects, instead of the name_localized and description_localized fields. Default false.
```

interactions/api/http/scheduledEvent.py
```diff
@@ -54,7 +54,7 @@ async def get_scheduled_event(
         guild_id, event_id = int(guild_id), int(guild_scheduled_event_id)
         params = {}
         if with_user_count:
-            params["with_user_count"] = with_user_count
+            params["with_user_count"] = f"{with_user_count}"
```
Toricane: Why f-string instead of string casting? Example `str(with_user_count)`
```suggestion
            params["with_user_count"] = str(with_user_count)
```

interactions/client/bot.py
```diff
@@ -126,7 +131,7 @@ def __register_events(self) -> None:
         self._websocket._dispatch.register(self.__raw_message_create, "on_message_create")
         self._websocket._dispatch.register(self.__raw_guild_create, "on_guild_create")
 
-    async def __compare_sync(self, data: dict, pool: List[dict]) -> bool:
+    async def __compare_sync(self, data: dict, pool: List[dict]) -> (bool, dict):
```
Toricane: Uhh, shouldn't that be `Tuple[bool, dict]`?
```suggestion
    async def __compare_sync(self, data: dict, pool: List[dict]) -> Tuple[bool, dict]:
```
May need to import it if not already

interactions/client/bot.pyi
```diff
@@ -41,15 +44,11 @@ class Client:
     def latency(self) -> float: ...
     def start(self) -> None: ...
     def __register_events(self) -> None: ...
-    async def __compare_sync(self, data: dict) -> None: ...
-    async def __create_sync(self, data: dict) -> None: ...
-    async def __bulk_update_sync(
-        self, data: List[dict], delete: Optional[bool] = False
-    ) -> None: ...
-    async def _synchronize(self, payload: Optional[dict] = None) -> None: ...
+    async def __compare_sync(self, data: dict, pool: List[dict]) -> (bool, dict): ...
```
Toricane:
```suggestion
    async def __compare_sync(self, data: dict, pool: List[dict]) -> Tuple[bool, dict]: ...
```

interactions/api/http/client.py
```diff
@@ -3,35 +3,35 @@
...
+    GuildRequest,
+    InteractionRequest,
+    MemberRequest,
+    MessageRequest,
+    ReactionRequest,
+    ScheduledEventRequest,
+    StickerRequest,
+    ThreadRequest,
+    UserRequest,
+    WebhookRequest,
```
Toricane, line 34: Why was this done?
It doesn't relate to the PR..?

Toricane (changes requested):
Some suggestions

interactions/api/http/client.py
sillyElvy: I needed the typehints ant then left it in lol

Toricane:
@EdVraz I haven't tested this so I'm not 100% sure, but wouldn't this break autocomplete?
Looking at the code, it syncs all of the commands at once in `Client.start()`, so it would break autocomplete options since they worked by getting the command from the cache (I think), and raise an error if not found.
I feel like autocomplete needs an overhaul in general, just letting you know that this will break it.

interactions/client/bot.py
```diff
@@ -339,6 +235,131 @@ async def wait_until_ready(self) -> None:
         """Helper method that waits until the websocket is ready."""
         await self._websocket.wait_until_ready()
 
+    async def __sync(self) -> None:  # sourcery no-metrics
+        """
+        Syncs all commands to the API
+        """
```
```suggestion
        """Synchronizes all commands to the API."""
```

Toricane (changes requested):
Docstring suggestion

sillyElvy:
> @EdVraz I haven't tested this so I'm not 100% sure, but wouldn't this break autocomplete?
> Looking at the code, it syncs all of the commands at once in `Client.start()`, so it would break autocomplete options since they worked by getting the command from the cache (I think), and raise an error if not found.
> I feel like autocomplete needs an overhaul in general, just letting you know that this will break it.

It probably would, we could get the commands in the client init instead?

interactions/client/bot.py
```diff
@@ -339,6 +237,167 @@ async def wait_until_ready(self) -> None:
...
+
+            self.__guild_commands[_id] = {"commands": _cmds, "clean": True}
+
+    async def __sync(self) -> None:  # sourcery no-metrics
+        """Synchronizes all commands to the API."""
+
+        log.debug("starting command sync")
+        _guilds = await self._http.get_self_guilds()
+        _guild_ids = [int(_["id"]) for _ in _guilds]
+        self._scopes.add(_guild_ids)
```
FayeDel: Can't hash a list.

sillyElvy: How can I put the items from the list into the set? Is there like a method or do I have to do a for-loop?

FayeDel: Instead of `add()`, switch it to `update()`

sillyElvy: With the same argument?

FayeDel: Yes.

sillyElvy: Cool, thanks

sillyElvy: Done

FayeDel:
This is more of a note, but autocomplete fails to register when you introduce a new autocomplete command with a 
`KeyError: 'commands'`, and another KeyError follows if you use it via a guild command, saying that the scope ID within the dict doesn't exist as a key.

sillyElvy:
> This is more of a note, but autocomplete fails to register when you introduce a new autocomplete command with a 
> `KeyError: 'commands'`, and another KeyError follows if you use it via a guild command, saying that the scope ID within the dict doesn't exist as a key.

Oh I exactly know what I did wrong, this is pain

sillyElvy:
I'll try to fix it from mobile lol

interactions/client/bot.py
```diff
@@ -156,126 +182,113 @@ async def __compare_sync(self, data: dict, pool: List[dict]) -> bool:
...
+                                    if not clean:
+                                        break
+                                if not clean:
+                                    break
+
+                        if not clean:
+                            break
+
+                        else:
+                            continue
```
AstreaTSS: I'm usually fine with hacky code, especially in Python. Obviously, it's not *great*, but sometimes it's the only way of doing something without resorting to something arguably worse.

that being said wtf there *has* to be a better way

sillyElvy: I just realized something lol

sillyElvy: fixed

i0bs:
FYI, you can always amend `!` to the convention of your PR commit(s) to indicate that it's breaking instead of suffixing or prefixing as such in parenthesis. (e.g. `refactor!:`)

interactions/client/bot.py
```diff
@@ -872,13 +1045,10 @@ def decorator(coro: Coroutine) -> Any:
...
 
-    @staticmethod
-    def _find_command(commands: List[Dict], command: str) -> ApplicationCommand:
+    def _find_command(self, command: str) -> ApplicationCommand:
         """
         Iterates over `commands` and returns an :class:`ApplicationCommand` if it matches the name from `command`
 
-        :ivar commands: The list of dicts to iterate through
-        :type commands: List[Dict]
         :ivar command: The name of the command to match:
```
```suggestion
        :param command: The name of the command to match
```

Toricane (dismissed):
This is really complicated, gonna test it out
I left a suggestion

i0bs (approved):
crunching on a deadline. so far, the new behavior has performed well and later iterations can always be met.

Issue #779, opened 2022-05-08
[BUG] Autocomplete does not work under subcommand


B1ue-Dev:
### Describe the bug.

When trying to use ``@interactions.extension_aautocomplete()`` with an ``options`` under subcommand, it will not work. However, with normal command, it **does** work. Not sure why this can happen though.

### List the steps.

1. Create a subcommand with the following name ``/tag view``, with an option called ``tag_name``.
2. In autocomplete: ``@interactions.extension_autocomplete(command="tag", name="tag_name")``.
3. Run the code and test.

### What you expected.

I was expecting autocomplete to populate the choice in the option of subcommand.

### What you saw.

Autocomplete did not function on option selection in subcommand.

No error log in the Terminal. It just would not work.

### What version of the library did you use?

stable

### Version specification

4.2.0

### Code of Conduct

- [X] I agree to follow the contribution requirements.

B1ue-Dev:
I think this can be closed now.

PR #781, opened 2022-05-08
fix: Allow sub command autocomplete dispatch

sillyElvy:
## About
This pull request corrects the autocomplete dispatch so it works for subcommands

## Checklist

  - (If existent): #779 
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix

### DIFF
--- a/interactions/api/gateway/client.py
+++ b/interactions/api/gateway/client.py
@@ -311,14 +311,42 @@ class WebSocketClient:
 
                     if _context.data._json.get("options"):
                         for option in _context.data.options:
+                            if isinstance(option, dict):
+                                option = Option(**option)
+                            if option.type not in (OptionType.SUB_COMMAND, OptionType.SUB_COMMAND_GROUP):
+                                if option.focused:
+                                    __name, _value = self.__sub_command_context(option, _context)
+                                    _name += f"_{__name}" if __name else ""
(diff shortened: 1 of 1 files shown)


interactions/api/gateway/client.py
```diff
@@ -311,13 +311,47 @@ def _dispatch_event(self, event: str, data: dict) -> None:  # sourcery no-metric
...
+                                            __option = Option(**__option)
+                                        if __option.focused:
+                                            __name, _value = self.__sub_command_context(
+                                                __option, _context
+                                            )
+                                            _name += f"_{__name}" if __name else ""
+                                            if _value:
+                                                __args.append(_value)
+                                        break
+                                    break
```
Toricane, line 354: I don't understand how this part works.
Why does it `break` in the first loop?
It only checks if the option is focused for the *first* option if it exists.
Are you not allowed to have autocomplete options past the first option of the command?

sillyElvy: I think it only parses one option since you can only enter one, anyways, so only one is present. But it was in the code before, so I kept it, whoever made it probably made it with a reason xD

Toricane (approved):
LGTM!

Issue #793, opened 2022-05-17
[REQUEST] Implement clone channel method


Nanrech:
### Describe the feature.

It'd be very neat to have a way to automatically clone channels in a guild. I will create a PR that adds this feature as soon as this issue is public.

### Code of Conduct

- [X] I agree to follow the contribution requirements.

PR #794, opened 2022-05-17
feat: Add clone_channel method to guild

Nanrech:
## About

This PR adds the clone_channel method to the guild class.

## Checklist

  - #793 
  - [ ] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix

### DIFF
--- a/interactions/api/models/guild.py
+++ b/interactions/api/models/guild.py
@@ -779,17 +779,17 @@ class Guild(DictSerializerMixin):
         """
         if not self._client:
             raise AttributeError("HTTPClient not found!")
-        ch = Channel(**await self._client.get_channel(channel_id=channel_id))
+        res = await self._client.get_channel(channel_id=channel_id)
 
-        ch._json["permission_overwrites"] = [
-            Overwrite(**_) for _ in ch._json["permission_overwrites"]
+        res["permission_overwrites"] = [
+            Overwrite(**_) for _ in res["permission_overwrites"]
         ]
         [
-            ch._json.pop(attr, None)
+            res.pop(attr, None)
             for attr in {"flags", "last_pin_timestamp", "guild_id", "id", "last_message_id"}
         ]
 
-        return await self.create_channel(**ch._json)
+        return await self.create_channel(**res)
 
     async def modify_channel(
         self,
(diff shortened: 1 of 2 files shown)


interactions/api/models/guild.py
```diff
@@ -768,6 +768,27 @@ async def create_channel(
...
+        """
+        if not self._client:
+            raise AttributeError("HTTPClient not found!")
+        ch = Channel(**await self._client.get_channel(channel_id=channel_id))
+
+        ch._json["permission_overwrites"] = [
+            Overwrite(**_) for _ in ch._json["permission_overwrites"]
+        ]
+        [
+            ch._json.pop(attr, None)
```
sillyElvy: I just noticed, this is senseless. You can just modify the value from `get_channel`, you don't need to create an object from it

sillyElvy (changes requested):

interactions/api/models/guild.py
Nanrech: WDYM?

sillyElvy (approved):

interactions/api/models/guild.py
```diff
@@ -768,6 +768,27 @@ async def create_channel(
...
+        """
+        if not self._client:
+            raise AttributeError("HTTPClient not found!")
+        res = await self._client.get_channel(channel_id=channel_id)
+
+        res["permission_overwrites"] = [Overwrite(**_) for _ in res["permission_overwrites"]]
+        [
+            res.pop(attr, None)
+            for attr in {"flags", "last_pin_timestamp", "guild_id", "id", "last_message_id"}
+        ]
```
Toricane: Might as well make it just a for loop instead of a list comprehension if black formats it as 4 lines
Also sorted the attributes alphabetically so it looks nicer
```suggestion
        for attr in {"flags", "guild_id", "id", "last_message_id", "last_pin_timestamp"}:
            res.pop(attr, None)
```

Toricane (changes requested):
Loop modification

interactions/api/models/guild.py
```diff
@@ -768,6 +768,25 @@ async def create_channel(
 
         return Channel(**res, _client=self._client)
 
+    async def clone_channel(self, channel_id: int) -> Channel:
+        """
+        Clones a channel of the guild.
+
+        :param channel_id: The id of the channel to clone
+        :type channel_id: int
+        :return: The created (cloned) channel
```
Toricane: Probably just say cloned for consistency
```suggestion
        :return: The cloned channel
```

Toricane (changes requested):
One small suggestion

Toricane (approved):
LGTM!

PR #795, opened 2022-05-20
refactor!: migration to use attrs on the majority of models

Catalyst4222:
## About

This PR converts the DictSerializationMixin to use attrs instead of slots
Some benefits:
- Nested type conversion
- No massive `__init__`s
- The ability to rename data passed from discord into different attributes
- Stricter object creation (e.x. requiring HTTPClient for classes that make use of it)

This is very breaking

## Checklist

  - (If existent):
  - [ ] Documentation
  - [x] Breaking change
  - [x] New feature/enhancement
  - [x] Bugfix


sillyElvy:
Is there a way `client` could stay underscored? That's very big breaking change

sillyElvy:
Also why is this a feat and not a refactor?

Catalyst4222:
> Is there a way client could stay underscored? That's very big breaking change
It would be possible, but attrs strips the leading underscore by default, and I've already changed most of the things to use client. I can figure out a way to keep the underscore if needed, but it would have to wait a bit
> Also why is this a feat and not a refactor?
Because I'm used to making only feat/fix 😅, changing that now

sillyElvy:
> I can figure out a way to keep the underscore if needed, but it would have to wait a bit

I really don't want to let this PR through until that, a lot of users would have to change their code for this change and that's kinda bad

Catalyst4222:
I said that I could do it, I just need time to do so

sillyElvy:
I just wanted to clarify *why* I have a problem with it 😅

Alright, thanks

i0bs:
Creating a property method to alias client as underscored may be the best approach for now

AstreaTSS:
Napkining real quick - a way of allowing `_client` to still be a thing is by making a subclass of the mixin that has it. Something like so:

```python
@attrs.define(...) # same as the dict serializer
class ClientSerializerMixin(DictSerializerMixin):
    _client: "HTTPClient" = attrs.field(init=False, repr=False)

    def __init__(self, kwargs_dict: dict = None, **other_kwargs):
        kwargs = kwargs_dict or other_kwargs
        self._client = kwargs.get("_client")
        super().__init__(**kwargs)
```

interactions/api/models/misc.py
```diff
@@ -7,58 +7,116 @@
...
-        This does NOT convert them to its own data types, i.e. timestamps, or User within Member. This is left by
-        the object that's using the mixin.
-    """
+    ...
+
+
+@attrs.define(eq=False, init=False, on_setattr=attrs.setters.NO_OP)
+class DictSerializerMixin:
+    _json: dict = attrs.field(init=False, repr=False)
+    extras: dict = attrs.field(init=False, repr=False)
```
AstreaTSS: `extras` should be underscored *just* in case Discord adds an attribute called just that.

Catalyst4222: Changed in [7e1a139](https://github.com/interactions-py/library/pull/795/commits/7e1a13994ce64c992cc6ea550281bca170f3e6f0)

AstreaTSS:
Heads up: I'll be reviewing this in small chunks due to the *huge* amount of code changes and will likely not do a super deep analysis into every piece of code.

interactions/api/gateway/client.py
```diff
@@ -290,10 +290,10 @@ def _dispatch_event(self, event: str, data: dict) -> None:  # sourcery no-metric
                             )
                             if _type:
                                 if isinstance(option, dict):
-                                    _type[option["value"]]._client = self._http
+                                    _type[option["value"]].client = self._http
```
AstreaTSS, line 295: Shouldn't this (and pretty much the entire file here) be using `_client`?

interactions/api/models/message.py
```diff
@@ -154,1186 +155,1088 @@ class ChannelMention(DictSerializerMixin):
...
+    sticker_items: Optional[List[PartialSticker]] = field(converter=convert_list, default=None)
+    stickers: Optional[List[Sticker]] = field(
+        converter=convert_list(Sticker), default=None
+    )  # deprecated
 
-        interactions.EmbedFooter(
-            text="yo mama so short, she can fit in here",
-        )
+    def __repr__(self) -> str:
+        return self.content
```
AstreaTSS: I *think* this repr can removed - `attrs` can just use it's usual repr. Using the content isn't *really* great for representing messages when many people won't even have the guild message content intent on.

I know, this kind of sounds like a thing for a separate PR, and it honestly might be, but this was likely here in the first place because there was no auto-generating repr. There is one now.

AstreaTSS (commented):
Note: my web browser is lagging due to how big this PR is 😅 - I wanted to review a bit more, but I'll have to stop things here for now.

interactions/api/gateway/client.py
Catalyst4222: I believe `.client` is correct there, since the Context classes don't have it underscored

sillyElvy: I think it's for an option value here, not for context itself?

Catalyst4222: I've tested with options, and they work
That line isn't even touched in this pr

AstreaTSS (approved):
Looks good!

Heads up though: I don't really have anything to test the extensive amount of changes this PR does - I could only do some basic tests here and there. If anyone else sees this, more testing would be *very* appreciated.

sillyElvy (changes requested):
I won't approve this until you are core dev /shrug

Damego:
> I won't approve this until you are core dev /shrug

Now you should approve this pr

sillyElvy (changes requested):
You have merge conflicts

Catalyst4222:
This pr should be ready, all merge conflicts have been fixed and I have tested as much as I can remember

sillyElvy (approved):

PR #801, opened 2022-05-24
feat: implement ``get`` function

sillyElvy:
## About

This is a new approach for the ``get`` method. It is now located in a separate file, so the `@overload`s don't mess another file up. Having an overload decorator for every object allows to precisely show the user what arguments he has to put in, and for us to do a simple `**kwagrs` into the function that actually gets the object

## Checklist

  - (If existent):
  - [x] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix


sillyElvy:
pre commit check will (hopefully) fail until webhook pr is merged

interactions/client/get.py
```diff
@@ -0,0 +1,36 @@
...
+    if not isinstance(obj, type):
+        raise TypeError("The object must not be an instance of a class!")
+
+    _name = f"get_{obj.__name__.lower()}"
+
+    if obj in (Role, Emoji):
+        _guild = Guild(**await client._http.get_guild(kwargs.pop("guild_id")), _client=client._http)
+        _func = getattr(_guild, _name)
+
+    _func = getattr(client._http, _name)
```
Toricane: You're immediately reassigning `_func`, you should fix that

Toricane (changes requested):

interactions/client/get.py
```diff
@@ -0,0 +1,37 @@
...
+        raise TypeError("The object must not be an instance of a class!")
+
+    _name = f"get_{obj.__name__.lower()}"
+
+    if obj in (Role, Emoji):
+        _guild = Guild(**await client._http.get_guild(kwargs.pop("guild_id")), _client=client._http)
+        _func = getattr(_guild, _name)
+        return await _func(**kwargs)
+
+    _func = getattr(client._http, _name)
```
Toricane: Remove return and await, put else on 34, indent second _func to inside else block

Toricane (changes requested):
I didn't mean *that*...

interactions/client/get.py
sillyElvy: The func for role and emoji are returning a full object tho so it makes sense to return directly from it

Toricane: Ohh ok

Toricane (commented):
Looks good, just missing cache implementation.

Also, should getting a list of an object be implemented, or is that out of the scope of this?
For example, `List[User]`, with a list of user IDs inputted, which returns a list of User objects.

sillyElvy:
I hate everything xD

interactions/client/get.py
```diff
@@ -0,0 +1,158 @@
...
+        _name = f"get_{obj.__name__.lower()}"
+        if isinstance(obj, _GenericAlias):
+            _obj = get_args(obj)[0]
+            _objects: List[_obj] = []
+
+            # TODO: add cache, include getting for IDs that are `None`
+
+            if len(list(kwargs)) == 2:
+                if guild_id := kwargs.pop("guild_id", None):
+                    _guild = Guild(**await client._http.get_guild(guild_id), _client=client._http)
```
AstreaTSS: Any `await`s need to be moved out of this function, since it's synchronous. Ideally, any logic even touching HTTP should be in its own method.

(This isn't the only instance of `await`, just don't want to repeat myself.)

AstreaTSS (dismissed):

interactions/client/get.py
sillyElvy: I know, I'm working on it

AstreaTSS (commented):
I find the number of double underscored variables to be excessive and unnecessary in this case, but that's a me thing.

Otherwise, the code looks good! I haven't tested it though - busy with some other things. I also likely have missed a couple of things, which is why I'm not giving it explicit approval.

sillyElvy:
> I find the number of double underscored variables to be excessive and unnecessary in this case, but that's a me thing.
> 
> Otherwise, the code looks good! I haven't tested it though - busy with some other things. I also likely have missed a couple of things, which is why I'm not giving it explicit approval.

I only made double underscores when single underscore was used already, no? (except the two functions)

AstreaTSS:
> > I find the number of double underscored variables to be excessive and unnecessary in this case, but that's a me thing.
> > Otherwise, the code looks good! I haven't tested it though - busy with some other things. I also likely have missed a couple of things, which is why I'm not giving it explicit approval.
> 
> I only made double underscores when single underscore was used already, no? (except the two functions)

Right, true. Still don't think there's even a need for single underscores personally (while it is protected, variables in a function are already assumed to be that anyways), but again, personal preference.

interactions/client/get.py
```diff
@@ -0,0 +1,289 @@
...
+                Snowflake(kwargs.get("member_id")),
+            )  # Fuck it, I can't be dynamic on this
+        else:
+            _values = Snowflake(kwargs.get(kwarg_name))
+
+        _obj = client.cache[_object].get(_values)
+    return _obj
+
+
+def _search_iterable(*args, **kwargs) -> Optional[_A]:
```
Catalyst4222: `_search_iterable` only ever gets a single argument, why not show that in the parameters?

Catalyst4222: Same thing applies to the keyword arguments. Double stars doesn't let you set a parameter name, which makes it confusing to know what to pass when first looking at it

sillyElvy: It takes `**kwargs` because you can put any attribute in it, setting a value to it doesn't allow that. I'll change the `*args` tho

Catalyst4222 (changes requested):
You have `_http_request`, `_return_cache`, `_get_cache`, and `_search_iterable` as available functions, but you don't have them available in the .pyi file

docs/get.rst
```diff
@@ -0,0 +1,8 @@
+.. currentmodule:: interactions
+
+The ``get``-Method
```
i0bs:
```suggestion
The ``get`` utility method
```

interactions/client/get.py
```diff
@@ -0,0 +1,394 @@
...
+
+__all__ = (
+    "get",
+    "Force",
+)
+
+
+class Force(str, Enum):
+    """
+    An enum representing the force types for the get method.
```
i0bs:
```suggestion
    An enumerable object representing the force types for the get method.
```

interactions/client/get.py
```diff
@@ -0,0 +1,394 @@
...
+
+        def _check():
+            return obj == list[get_args(obj)[0]]
+
+    else:
+
+        def _check():
+            return False
+
+    if len(args) == 2 and any(isinstance(_, Iterable) for _ in args):
```
i0bs: Condense these two conditionals which share the same comparison on the main scope.

interactions/client/get.py
```diff
@@ -0,0 +1,394 @@
...
+from ..api.models.message import Emoji, Message, Sticker
+from ..api.models.misc import Snowflake
+from ..api.models.role import Role
+from ..api.models.user import User
+from ..api.models.webhook import Webhook
+from .bot import Client
+
+log = getLogger("get")
+
+_A = TypeVar("_A", Channel, Guild, Webhook, User, Sticker, Message, Emoji, Role, Message)
```
i0bs:
```suggestion
_T = TypeVar("_T")
```

This is a **huge** change in typing annotation behaviour, but I'm suggesting this as restricting the types will only serve be problematic later on in time when the API releases even more objects that we can either pull from cache or make http calls for. For example, I can't get information on:

- Application commands
- Components
- Webhooks

i0bs (changes requested):
Documentation and typing changes

interactions/client/get.py
```diff
@@ -0,0 +1,394 @@
...
+
+    if len(args) == 2 and any(isinstance(_, Iterable) for _ in args):
+        raise LibraryException(message="You can only use Iterables as single-argument!", code=12)
+
+    if len(args) == 2:
+
+        client, obj = args
+        if not isinstance(obj, type) and not isinstance(obj, _GenericAlias):
+            client: Client
+            obj: Union[Type[_A], Type[List[_A]]]
```
i0bs:
```suggestion
            obj: Union[Type[_T], Type[List[_T]]]
```

interactions/client/get.py
```diff
@@ -0,0 +1,394 @@
...
+            raise LibraryException(
+                message="The object must not be an instance of a class!", code=12
+            )
+
+        kwargs = _resolve_kwargs(obj, **kwargs)
+        http_name = f"get_{obj.__name__.lower()}"
+        kwarg_name = f"{obj.__name__.lower()}_id"
+        if isinstance(obj, _GenericAlias) or _check():
+            _obj: Type[_A] = get_args(obj)[0]
```
i0bs:
```suggestion
            _obj: Type[_T] = get_args(obj)[0]
```

interactions/client/get.py
```diff
@@ -0,0 +1,394 @@
...
+            return _http_request(obj=obj, http=client._http, _name=http_name, **kwargs)
+
+    elif len(args) == 1:
+        return _search_iterable(*args, **kwargs)
+
+
+async def _http_request(
+    obj: Type[_A],
+    http: HTTPClient,
+    request: Union[Coroutine, List[Union[_A, Coroutine]], List[Coroutine]] = None,
```
i0bs:
```suggestion
    request: Union[Coroutine, List[Union[_T, Coroutine]], List[Coroutine]] = None,
```

interactions/client/get.py
```diff
@@ -0,0 +1,394 @@
...
+    _name: str = None,
+    **kwargs,
+) -> Union[_A, List[_A]]:
```
i0bs:
```suggestion
) -> Union[_T, List[_T]]:
```

interactions/client/get.py
```diff
@@ -0,0 +1,394 @@
...
+        return obj(**_obj, _client=http)
+
+    if not isinstance(request, list):
+        return obj(**await request, _client=http)
+
+    return [obj(**await req, _client=http) if isawaitable(req) else req for req in request]
+
+
+async def _return_cache(
+    obj: Union[Optional[_A], List[Optional[_A]]]
```
i0bs:
```suggestion
    obj: Union[Optional[_T], List[Optional[_T]]]
```

interactions/client/get.py
```diff
@@ -0,0 +1,394 @@
...
+) -> Union[Optional[_A], List[Optional[_A]]]:
```
i0bs:
```suggestion
) -> Union[Optional[_T], List[Optional[_T]]]:
```

interactions/client/get.py
```diff
@@ -0,0 +1,394 @@
...
+    await sleep(0)  # iirc Bluenix meant that any coroutine should await at least once
+    return obj
+
+
+def _get_cache(
+    _object: Type[_A], client: Client, kwarg_name: str, _list: bool = False, **kwargs
```
i0bs:
```suggestion
    _object: Type[_T], client: Client, kwarg_name: str, _list: bool = False, **kwargs
```

interactions/client/get.py
```diff
@@ -0,0 +1,394 @@
...
+) -> Union[Optional[_A], List[Optional[_A]]]:
```
i0bs:
```suggestion
) -> Union[Optional[_T], List[Optional[_A]]]:
```

interactions/client/get.py
```diff
@@ -0,0 +1,394 @@
...
+                Snowflake(kwargs.get("member_id")),
+            )  # Fuck it, I can't be dynamic on this
+        else:
+            _values = Snowflake(kwargs.get(kwarg_name))
+
+        _obj = client.cache[_object].get(_values)
+    return _obj
+
+
+def _search_iterable(items: Iterable[_A], **kwargs) -> Optional[_A]:
```
i0bs:
```suggestion
def _search_iterable(items: Iterable[_T], **kwargs) -> Optional[_T]:
```

interactions/client/get.pyi
```diff
@@ -0,0 +1,116 @@
...
+from enum import Enum
+from ..api.models.channel import Channel
+from ..api.models.guild import Guild
+from ..api.models.member import Member
+from ..api.models.message import Message, Emoji, Sticker
+from ..api.models.user import User
+from ..api.models.webhook import Webhook
+from ..api.models.role import Role
+
+_T = TypeVar("_T", Channel, Guild, Webhook, User, Sticker)
```
i0bs: `_T` is really meant for allowing any typing under the type variable. I believe you're wanting to use `_A` for this manner which you're not following.

interactions/client/get.pyi
```diff
@@ -0,0 +1,116 @@
...
+_P = TypeVar("_P", Member, Emoji, Role, Message)
```
i0bs: What is `_P`'s purpose?

interactions/client/get.pyi
```diff
@@ -0,0 +1,116 @@
...
+_A = TypeVar("_A", Channel, Guild, Webhook, User, Sticker, Message, Emoji, Role, Message)
```
i0bs: See regarding comment about the behaviour of `_A`.

i0bs (changes requested):
More typing changes.

interactions/client/get.py
```diff
@@ -0,0 +1,394 @@
...
+        )
+
+    _arg = str(list(kwargs)[0])
+
+    __obj = next(
+        (
+            item
+            for item in items
+            if (
+                str(getattr(item, _arg, None)) == str(kwargs.get(_arg))
```
i0bs: This logically reads out to: `True if not True else foo`. What's the goal from the object iteration if the iterable can be any expression?

i0bs (changes requested):
This is a bit questionable..

interactions/client/get.py
sillyElvy: I don't get what you mean, sorry

interactions/client/get.py
EdVraz: I have another definition of _T in the pyi already tho

interactions/client/get.py
sillyElvy: Wdym

sillyElvy: Did smth maybe you meant that idk

interactions/client/get.py
```diff
@@ -0,0 +1,397 @@
...
+def _get_cache(
+    _object: Type[_T], client: Client, kwarg_name: str, _list: bool = False, **kwargs
+) -> Union[Optional[_T], List[Optional[_T]]]:
+    if _list:
+        _obj = []
+        if _object == Member:  # Can't be more dynamic on this
+            _values = ()
+            _guild_id = Snowflake(kwargs.get("guild_id"))
+            for _id in kwargs.get("member_ids"):
+                _values += (
```
Catalyst4222: Why are you using a tuple to gather the values? Using a list and appending to it is much faster, and lists are made to do things like this

Catalyst4222 (changes requested):

interactions/client/get.py
sillyElvy: Because from my last understanding a member needed a Tuple of guild and member id, no?

Catalyst4222: Yes, but you can have a list of tuples. As long as the key you're using is a tuple of the right ids, it doesn't matter where/how you get/store it

interactions/client/get.py
```diff
@@ -0,0 +1,397 @@
...
+    CACHE = "cache"
+    HTTP = "http"
+
+
+def get(*args, **kwargs):
+    """
+    Helper method to get an object.
+
+    This method can do the following:
+        * Get a list of a specific objects
```
mAxYoLo01:
```suggestion
        * Get a list of specific objects
```

interactions/client/get.py
```diff
@@ -0,0 +1,397 @@
...
+                )
+                # because of cache only, this can be None
+
+
+        Both examples should have given you a basic overview on how to get a single object. Now we will move on with
+        lists of objects.
+
+        .. important::
+            The ``parent_id`` represents the channel or guild id that belongs to the objects you want to get. It is
+            called ``parent_id`` bcause ``guild_or_channel_id`` would be horrible to type out every time.
```
mAxYoLo01:
```suggestion
            called ``parent_id`` because ``guild_or_channel_id`` would be horrible to type out every time.
```

interactions/client/get.py
sillyElvy: Will do

i0bs:
Before I make a final review, I'd really like to discuss my disdain for `@overload`.

This is going to obviously lead to an extremely messy and obfuscated pattern in the type hinting for numerous objects. Is there not any other way to assure to end-users that objects passed by the Discord API/Gateway are usable in this context?

sillyElvy:
> Before I make a final review, I'd really like to discuss my disdain for `@overload`.
> 
> This is going to obviously lead to an extremely messy and obfuscated pattern in the type hinting for numerous objects. Is there not any other way to assure to end-users that objects passed by the Discord API/Gateway are usable in this context?

There are 9 overloads (what already is a massive reduction from the 36 we had before) and the typehints can be controlled via the `_SA` and `_MA` TypeVars, i don't think that it is too messy

i0bs (approved):
Long awaited, and after nearly a year of debating how to develop this, I'm giving it my seal of approval.

Catalyst4222 (approved):

interactions/client/get.py
```diff
@@ -0,0 +1,398 @@
...
+            return await _func(**kwargs)
+
+        _func = getattr(http, _name)
+        _obj = await _func
+        return obj(**_obj, _client=http)
+
+    if not isinstance(request, list):
+        return obj(**await request, _client=http)
+
+    return [obj(**await req, _client=http) if isawaitable(req) else req for req in request]
```
Toricane, line 298: What about asyncio.gather?

```suggestion
    awaitables = []
    dicts = []
    for req in request:
        if isawaitable(req):
            awaitables.append(req)
        else:
            dicts.append(req)

    loop = get_event_loop()
    group = asyncio.gather(*request)
    result = group.run_until_complete(group)

    requests = awaitables + dicts
    return [obj(**req, _client=http) for req in requests]
```

sillyElvy: we want to keep it async if it is gotten from HTTP, also I *really* don't want to mess with event loops

Toricane: Instead of doing it one after the other, it would do it all at once and return the results in the same order

sillyElvy: .. ok I'll modify it but no loop stuff lol

sillyElvy: actually it would not keep it in order so I won't change it (order would not be kept between not coros and coros)

Catalyst4222: You could also wrap the non-awaitables in another thing so that you could still use gather

sillyElvy: this approach is too complicated imo

Toricane:
```suggestion
    awaitables = []
    for req in request:
        if isawaitable(req):
            awaitables.append(req)
        else:
            async def awaitable():
                return req
            awaitables.append(awaitable())

    loop = get_event_loop()
    group = asyncio.gather(*awaitables)
    result = loop.run_until_complete(group)

    return [obj(**res, _client=http) for res in result]
```

sillyElvy: Alright, will do later

Toricane: Try it out and see if there is a speed difference
If there is a major difference, use the better one
Else use the simpler one

sillyElvy: This does not work, because `Client` already has a event loop and I can't get it to work without the loop stuff, so we won't use it ig

Toricane: Wait a second, you could just *await* it lol
```suggestion
    async def awaitable(param):
        return param

    awaitables = []
    for req in request:
        if isawaitable(req):
            awaitables.append(req)
        else:
            awaitables.append(awaitable(req))

    result = await asyncio.gather(*awaitables)
    return [obj(**res, _client=http) for res in result]
```

Toricane (approved):
LGTM!

PR #825, opened 2022-06-07
fix button example

Yurab0b:
button => button1

## About

## Checklist

  - (If existent):
  - [x] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [ ] Bugfix

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -262,7 +262,7 @@ implementation of a component:
 
 .. code-block:: python
 
-    button = interactions.Button(
+    button1 = interactions.Button(
         style=interactions.ButtonStyle.PRIMARY,
         label="hello world!",
         custom_id="hello"


Damego:
1. Refer to unstable branch
2. Use conventional commits

V3ntus:
I'm not sure what the purpose of this PR is for. Is it a typo? Can you please alter the description accordingly?

This change does not fix anything, in fact, I believe it is misleading and results in an error

Damego:
Yeah its typo

FayeDel:
I'm closing this PR because the PR itself isn't clear of what it should fix, documentation wise this works the way it currently is, and as they said above, just changing this line will result in an error.

If you think this is false, then I'd suggest giving a little more context to the PR and the issue in question so this PR may be reopened. ^^

PR #835, opened 2022-06-09
fix: parse sub cmd (and groups) options properly

AstreaTSS:
## About

This PR adjusts the option conversion logic for subcommands so that it handles both the `Option` object and the dict way. This method also adjusts the checks so that they use the resulting object, since it already has been converted.

## Checklist

  - (If existent):
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix

### DIFF
--- a/interactions/client/bot.py
+++ b/interactions/client/bot.py
@@ -668,7 +668,7 @@ class Client:
                     11, message="A sub command group cannot contain more than 25 sub commands!"
                 )
             for _sub_command in _sub_group.options:
-                __check_sub_command(Option(**_sub_command), _sub_group)
+                __check_sub_command(_sub_command, _sub_group)
 
         def __check_sub_command(_sub_command: Option, _sub_group: Option = MISSING):
             nonlocal _sub_cmds_present
@@ -704,7 +704,7 @@ class Client:
                     )
                 _sub_opt_names = []
                 for _opt in _sub_command.options:
-                    __check_options(Option(**_opt), _sub_opt_names, _sub_command)
+                    __check_options(_opt, _sub_opt_names, _sub_command)
                 del _sub_opt_names
 
         def __check_options(_option: Option, _names: list, _sub_command: Option = MISSING):
--- a/interactions/client/models/command.py
+++ b/interactions/client/models/command.py
@@ -114,7 +114,12 @@ class Option(DictSerializerMixin):
     def __attrs_post_init__(self):
         # needed for nested classes
         self.options = (
-            [Option(**option) for option in self.options] if self.options is not None else None
+            [
+                Option(**option._json) if hasattr(option, "_json") else Option(**option)  # type: ignore
+                for option in self.options
+            ]
+            if self.options is not None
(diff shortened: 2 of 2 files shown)


interactions/client/bot.py
```diff
@@ -668,7 +668,7 @@ def __check_sub_group(_sub_group: Option):
                     11, message="A sub command group cannot contain more than 25 sub commands!"
                 )
             for _sub_command in _sub_group.options:
-                __check_sub_command(Option(**_sub_command), _sub_group)
```
sillyElvy, line 671: How has it become an option?

AstreaTSS: Recursion, kinda:
- When initing an `ApplicationCommand`, `attrs` starts to convert the top-level options into, well, `Option`s (there's also already logic to handle if it's already an `Option` and just not convert it, BTW). Let's pretend this is a subcomamnd group.
- When converting the object, the subcomamnd group will run into the lines in `command.py`, meaning *its* options (subcommands) also get started on being converted to `Option`s.
- ...which, once again, *also* triggers the same set of lines, converting the options in said subcommands to `Option`s. The recursion ends here.

This effectively means *every single option* in every subcommand in the subcommand group, and every option in every subcommand, is now an `Option` after initialization.

sillyElvy: ok, but why did it not happen before?

AstreaTSS: In short: ¯\\\_(ツ)_/¯

In slightly-less-short: the old `Option`, when dealing with the mixture of `Option`s and dicts (for subcommands), made it so it *only* updated the `_json` field with the standardized (read: just dict) values. It didn't even bother converting the possible dicts into `Option`s for the main attribute, making the typehinting a slight lie, funny enough.
(Realistically, this was only a "lie" when processing data coming from Discord. The amount of users using dicts for options is very low if I had to guess.)

Now you may argue the fact that if the `options` attribute wasn't touched, then they have the possibility of being *either* an `Option` *or* a dict, and the old code had bugs. But before this, the whole command goes through a weird process that can basically be summarized as:
`**kwargs > ApplicationCommand > dict > ApplicationCommand`
(As for why, it's unimportant but the short answer is standardization and necessity.)

The part to focus on is how the `ApplicationCommand` first gets converted into a dict, which is done by reading `_json`. This means all of the options in said command (and its subcommands) are *guaranteed* to be a dict, as touched upon previously. Then, when converting that dict back into an `ApplicationCommand`, the options field were never touched (as mentioned previously), meaning they were still dicts in the "final" object.

...we *actually* make sure that the options are, well `Option`s for subcommands now when initializing `ApplicaitonCommand` (or, well, really when initializing `Option`). This means we no longer have to do this conversion for checking options for subcommands.
(The underlying `_json`'s options are still a dict BTW.)

sillyElvy: ah i see, thanks!

sillyElvy (approved):

PR #841, opened 2022-06-11
fix: various fixes for embeds

AstreaTSS:
## About

This PR fixes a lot of things about `Embed`s and its classes, including typehinting and setting certain properties.
As a fun side effect for one of the fixes, `convert_type` now exists and is used behind the scenes when possible.

## Checklist

  - (If existent):
  - [ ] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement (technically)
  - [x] Bugfix

### DIFF
--- a/interactions/api/models/attrs_utils.py
+++ b/interactions/api/models/attrs_utils.py
@@ -184,6 +184,8 @@ def field(
     **kwargs,
 ):
     if converter is not None:
+        if isinstance(converter, type):
+            converter = convert_type(converter)
         converter = attrs.converters.optional(converter)
 
     metadata = kwargs.get("metadata", {})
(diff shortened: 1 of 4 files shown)


interactions/api/models/attrs_utils.py
```diff
@@ -157,6 +157,15 @@ def inner_convert_dict(dict):
     return inner_convert_dict
 
 
+def convert_type(type_: type):
```
sillyElvy, line 160: you don't use this anywhere?

AstreaTSS: See line L188 - admittedly, it's only really used behind the scenes right now, but it could be expanded into being used elsewhere in the future.

sillyElvy: Ah thanks, I didnt see it before

sillyElvy (approved):

PR #860, opened 2022-06-16
fix: various attrs-related fixes for Message

AstreaTSS:
## About

This pull request fixes up many of the converters with `Message`, as many of them were not the best. This PR also "fixes" a typehinting internally (though, funny enough, it's still wrong, but at least it's... uh, correctly wrong? Yeah.). 

## Checklist

  - (If existent):
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix


interactions/api/models/message.py
```diff
@@ -22,6 +20,9 @@
 from .team import Application
 from .user import User
 
+if TYPE_CHECKING:
+    from ...client.models.component import Component
+
```
sillyElvy, line 25: This is no where used?

sillyElvy (changes requested):

interactions/api/models/message.py
AstreaTSS: It's used for `Message`'s `components`.

sillyElvy (approved):

PR #864, opened 2022-06-17
refactor: change how `get_role` works

sillyElvy:
## About

This PR changes how `guild.get_role` works to prevent and eliminate rate limits

## Checklist

  - (If existent):
  - [ ] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix


sillyElvy:
Yes, I became kinda addicted to `next()`

interactions/api/models/guild.py
```diff
@@ -1638,14 +1638,21 @@ async def get_role(
...
+                raise LibraryException(code=13)
+            roles = await self._client.get_all_roles(guild_id=int(self.id))
+            try:
+                role = next(role for role in roles if int(role["id"]) == role_id)
+                self.roles = [Role(**_) for _ in roles]
+                return Role(**role)
+            except StopIteration as exc:
+                raise LibraryException(
+                    message="The role you looked for was not found!", code=0, severity=30
+                ) from exc
```
Toricane:
```suggestion
        for role in self.roles:
            if int(role.id) == role_id:
                return role
        else:
            if not self._client:
                raise LibraryException(code=13)
            roles = await self._client.get_all_roles(guild_id=int(self.id))
            self.roles = [Role(**_) for _ in roles]
            for role in self.roles:
                if int(role.id) == role_id:
                    break
            else:
                raise LibraryException(
                    message="The role you looked for was not found!", code=0, severity=30
                )
            return role
```
Use for else
DISCLAIMER: did this on phone, pls test and format the code if necessary

Toricane (changes requested):
Use for else
DISCLAIMER: did this on phone, pls test and format the code if necessary

interactions/api/models/guild.py
sillyElvy: What do you have against `next()`?

Toricane:
Next keeps looping until the end; it's better to loop through and break early

interactions/api/models/guild.py
Toricane: Next loops till end of list
Breaking early in for loops is better
Also no catching exceptions, else is used instead

sillyElvy: Is for/else even a thing?

Toricane: Also the code is shorter

Toricane: Yes, google it

sillyElvy: Also please remove the `from exc` from the pre-last line

Toricane: Else is run if you don't break from for loop

Toricane: Ok

Toricane: > Also please remove the `from exc` from the pre-last line

Done

sillyElvy: Alright nice

sillyElvy:
I'll revert the removal of next

i0bs:
Why is the new module path in.. german?

sillyElvy:
> Why is the new module path in.. german?

I accidentally pressed translate for Google chrome and it somehow changed everything while commiting - I'm about to fix it

sillyElvy:
> > Why is the new module path in.. german?
> 
> I accidentally pressed translate for Google chrome and it somehow changed everything while commiting - I'm about to fix it 

Ok a chrome tab on mobile is not made for this lmao

sillyElvy:
I fucked this branch too hard, making another rq

PR #865, opened 2022-06-17
refactor: make `get_role` more consistent 

sillyElvy:
## About

## Checklist

  - (If existent):
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [ ] Bugfix


interactions/api/models/guild.py
```diff
@@ -1638,14 +1638,22 @@ async def get_role(
...
+            roles = await self._client.get_all_roles(guild_id=int(self.id))
+            self.roles = [Role(**_) for _ in roles]
+            for role in self.roles:
+                if int(role.id) == role_id:
+                    break
+            else:
+                raise LibraryException(
+                    message="The role you looked for was not found!", code=0, severity=30
+                )
+            return role
```
Toricane: One extremely tiny thing
```suggestion
            for role in self.roles:
                if int(role.id) == role_id:
                    return role
            else:
                raise LibraryException(
                    message="The role you looked for was not found!", code=0, severity=30
                )
```

Toricane (changes requested):
One extremely tiny thing

interactions/api/models/guild.py
```diff
@@ -1638,14 +1638,21 @@ async def get_role(
...
+                raise LibraryException(code=13)
+            roles = await self._client.get_all_roles(guild_id=int(self.id))
+            self.roles = [Role(**_) for _ in roles]
+            for role in self.roles:
+                if int(role.id) == role_id:
+                    return role
+            else:
+                raise LibraryException(
+                    message="The role you looked for was not found!", code=0, severity=30
+                )
```
Toricane: Tbh you don't need the else bits since it's a return anyways

```suggestion
        for role in self.roles:
            if int(role.id) == role_id:
                return role
        if not self._client:
            raise LibraryException(code=13)
        roles = await self._client.get_all_roles(guild_id=int(self.id))
        self.roles = [Role(**_) for _ in roles]
        for role in self.roles:
            if int(role.id) == role_id:
                return role
        raise LibraryException(
            message="The role you looked for was not found!", code=0, severity=30
        )
```
This would make it a little less readable though, lmk what you think

Toricane (changes requested):
Another thing, i swear this is the last one

interactions/api/models/guild.py
sillyElvy: It's OK ig

Toricane (approved):
LGTM!

PR #867, opened 2022-06-18
Fix: remove misplaced `.id`

sillyElvy:
## About

## Checklist

  - (If existent):
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix


interactions/base.py
```diff
@@ -6,7 +6,7 @@
     "__authors__",
 )
 
-__version__ = "4.2.1"
+__version__ = "4.3.0-beta.1"
```
Toricane: It's already in unstable..?

Toricane (changes requested):
Version

interactions/base.py
sillyElvy: I have no idea..

Toricane: It is, so why is this showing up? Idk maybe revert that commit and fetch upstream

sillyElvy: fixed

Toricane (approved):
LGTM!

PR #876, opened 2022-06-22
fix: set all optional list attrs to have a default value of [] instead of None

mAxYoLo01:
## About

To avoid unnecessary `if not self.list: self.list = []`, the default value of all optional attributes will have a default value of an empty list instead.

## Checklist

  - (If existent):
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix


interactions/api/models/guild.py
```diff
@@ -2269,12 +2269,12 @@ class GuildPreview(DictSerializerMixin):
...
 
     id: Snowflake = field(converter=Snowflake)
-    emojis: Optional[List[Emoji]] = field(converter=convert_list(Emoji), default=None)
+    emojis: Optional[List[Emoji]] = field(converter=convert_list(Emoji), factory=list)
     name: str = field()
     icon: Optional[str] = field(default=None)
     splash: Optional[str] = field(default=None)
     discovery_splash: Optional[str] = field(default=None)
-    features: Optional[List[str]] = field(default=None)
+    features: Optional[List[str]] = field(default=[])
```
sillyElvy: Factory

sillyElvy (changes requested):

AstreaTSS:
This is both absolutely not how `Optional` works *and* a breaking change. While I'm okay with it personally, this is something that needs a bit more discussion than it seems.

sillyElvy:
> This is both absolutely not how `Optional` works *and* a breaking change. While I'm okay with it personally, this is something that needs a bit more discussion than it seems.

Does optional enforce `None`? I thought an emtpy list would be fine since `not []` = `True`

AstreaTSS:
> > This is both absolutely not how `Optional` works *and* a breaking change. While I'm okay with it personally, this is something that needs a bit more discussion than it seems.
> 
> Does optional enforce `None`? I thought an emtpy list would be fine since `not []` = `True`

`Optional` is explicitly `Union[X, None]`, yeah. These are no longer `Optional`s - they're just lists.

sillyElvy:
> > > This is both absolutely not how `Optional` works *and* a breaking change. While I'm okay with it personally, this is something that needs a bit more discussion than it seems.
> > 
> > Does optional enforce `None`? I thought an emtpy list would be fine since `not []` = `True`
> 
> `Optional` is explicitly `Union[X, None]`, yeah. These are no longer `Optional`s - they're just lists.

I see, alright

But imho it is certainly better than `if x is None: x = []` in every helper that adds/removes from the list

PR #930, opened 2022-07-15
feat: Add converters

sillyElvy:
## About

This pull request implements converters into command options

## Checklist

  - (If existent):
  - [x] Documentation
  - [x] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix

### DIFF
--- a/docs/migration.rst
+++ b/docs/migration.rst
@@ -93,11 +93,11 @@ The following example shows and explains how to create commands effortlessly and
         # the description is the first line of the docstring or "No description set".
 
     @bot.command(default_scope=False)
-    @interactions.option(str, name="opt1")  # description is optional.
-    @interactions.option(4, name="opt2", description="This is an option.")
-    @interactions.option(interactions.Channel, name="opt3", required=True)
+    @interactions.option(str, name="opt1", required=True)  # description is optional.
+    @interactions.option(4, name="opt2", description="This is an option.", required=True)
+    @interactions.option(interactions.Channel, name="opt3")
     async def command_with_options(
-        ctx, opt3: interactions.Channel, opt1: str = None, opt2: int = None,
+        ctx, opt1: str, hi: int, opt3: interactions.Channel = None
     ):
         ...  # do something here.
         # the default scope is disabled for this command, so this is a global command.
@@ -193,11 +193,11 @@ your option name! Example:
     bot = interactions.Client("TOKEN", default_scope=1234567890)
 
     @bot.command(default_scope=False)
-    @interactions.option(str, name="opt1")  # description is optional.
-    @interactions.option(4, name="opt2", description="This is an option.", converter="hi")
-    @interactions.option(interactions.Channel, name="opt3", required=True)
+    @interactions.option(str, name="opt1", required=True)  # description is optional.
+    @interactions.option(4, name="opt2", description="This is an option.", converter="hi", required=True)
+    @interactions.option(interactions.Channel, name="opt3")
     async def command_with_options(
-        ctx, opt3: interactions.Channel, opt1: str = None, hi: int = None,
+        ctx, opt1: str, hi: int, opt3: interactions.Channel = None
     ):
         ...
 
--- a/docs/migration.rst
+++ b/docs/migration.rst
@@ -97,7 +97,7 @@ The following example shows and explains how to create commands effortlessly and
     @interactions.option(4, name="opt2", description="This is an option.")
     @interactions.option(interactions.Channel, name="opt3", required=True)
     async def command_with_options(
-        ctx, opt1: str, opt2: int, opt3: interactions.Channel = None
+        ctx, opt3: interactions.Channel, opt1: str = None, opt2: int = None,
     ):
         ...  # do something here.
         # the default scope is disabled for this command, so this is a global command.
@@ -206,7 +206,7 @@ your option name! Example:
     @interactions.option(4, name="opt2", description="This is an option.", converter="hi")
     @interactions.option(interactions.Channel, name="opt3", required=True)
     async def command_with_options(
-        ctx, opt1: str, hi: int, opt3: interactions.Channel = None
+        ctx, opt3: interactions.Channel, opt1: str = None, hi: int = None,
     ):
         ...
 
--- a/docs/migration.rst
+++ b/docs/migration.rst
@@ -97,7 +97,7 @@ The following example shows and explains how to create commands effortlessly and
     @interactions.option(4, name="opt2", description="This is an option.", required=True)
     @interactions.option(interactions.Channel, name="opt3")
     async def command_with_options(
-        ctx, opt1: str, hi: int, opt3: interactions.Channel = None
+        ctx, opt1: str, opt2: int, opt3: interactions.Channel = None
     ):
         ...  # do something here.
         # the default scope is disabled for this command, so this is a global command.
--- a/docs/migration.rst
+++ b/docs/migration.rst
@@ -191,15 +191,6 @@ your option name! Example:
     import interactions
 
     bot = interactions.Client("TOKEN", default_scope=1234567890)
-    # the default scope will be applied to all commands except for those
-    # that disable the feature in the command decorator via: `default_scope=False`
-
-    @bot.command()
-    async def command_name(ctx):
-        """Command description"""
-        ...  # do something here.
-        # the name of the command is the coroutine name.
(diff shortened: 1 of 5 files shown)


interactions/client/models/command.py
```diff
@@ -106,6 +106,7 @@ class Option(DictSerializerMixin):
     :ivar Optional[bool] autocomplete?: A status denoting whether this option is an autocomplete option.
     :ivar Optional[Dict[Union[str, Locale], str]] name_localizations?: The dictionary of localization for the ``name`` field. This enforces the same restrictions as the ``name`` field.
     :ivar Optional[Dict[Union[str, Locale], str]] description_localizations?: The dictionary of localization for the ``description`` field. This enforces the same restrictions as the ``description`` field.
+    :ivar Optional[str] converter: How the argument for the option in your function is called, if any
```
Toricane:
```suggestion
    :ivar Optional[str] converter: How the argument for the option in your function is passed, if any
```

interactions/client/models/command.py
```diff
@@ -128,6 +129,7 @@ class Option(DictSerializerMixin):
     description_localizations: Optional[Dict[Union[str, Locale], str]] = field(
         default=None
     )  # so can this
+    converter: Optional[str] = field(default=None)
```
Toricane, line 132: Is this excluded from _json? I think it may break some stuff if it is in there

Toricane (changes requested):
Some suggestions

interactions/client/models/command.py
sillyElvy: Maybe `named`?

interactions/client/models/command.py
sillyElvy: How to exclude?

Toricane: Delete from _json in attrs post init?

interactions/client/models/command.py
Toricane: No, the option name does not change, it's how the option info is passed to the function.

Toricane:
Also why is this a breaking change? I don't see how previous methods can break?

interactions/client/models/command.py
```diff
@@ -128,6 +129,7 @@ class Option(DictSerializerMixin):
...
 
     def __attrs_post_init__(self):
```
Toricane, line 134:
```suggestion
    def __attrs_post_init__(self):
        del self._json["converter"]
```

interactions/client/models/command.py
Toricane: https://github.com/interactions-py/library/pull/930/files#r922564988

interactions/client/models/command.py
sillyElvy: I meant "how the argument is named" lol

sillyElvy:
> Also why is this a breaking change? I don't see how previous methods can break?

Imo it is breaking because its a not backwards compatible change (aka if you add converters your code iss going to break on a prior version)

AstreaTSS:
> > Also why is this a breaking change? I don't see how previous methods can break?
> 
> Imo it is breaking because its a not backwards compatible change (aka if you add converters your code iss going to break on a prior version) 

W-What? Do any prior slash commands break because of this change? If not, it's not breaking.

sillyElvy:
> > > Also why is this a breaking change? I don't see how previous methods can break?
> > 
> > Imo it is breaking because its a not backwards compatible change (aka if you add converters your code iss going to break on a prior version) 
> 
> W-What? Do any prior slash commands break because of this change? If not, it's not breaking.

If you add this change in your code I'm pretty sure that your code wouldn't function anymore if you use an older version - and that was my understanding of a breaking change

docs/migration.rst
```diff
@@ -97,7 +97,7 @@ The following example shows and explains how to create commands effortlessly and
     @interactions.option(4, name="opt2", description="This is an option.")
     @interactions.option(interactions.Channel, name="opt3", required=True)
     async def command_with_options(
-        ctx, opt1: str, opt2, int, opt3: interactions.Channel = None
+        ctx, opt3: interactions.Channel, opt1: str = None, opt2: int = None,
```
AstreaTSS: This is a bit confusing - I feel like these options should be renamed in a more logical way if you want to use this format.

AstreaTSS:
> > > > Also why is this a breaking change? I don't see how previous methods can break?
> > > 
> > > Imo it is breaking because its a not backwards compatible change (aka if you add converters your code iss going to break on a prior version) 
> > 
> > W-What? Do any prior slash commands break because of this change? If not, it's not breaking.
> 
> If you add this change in your code I'm pretty sure that your code wouldn't function anymore if you use an older version - and that was my understanding of a breaking change 

No? Breaking changes mean you have to change your code if you move to this version - there's nothing stating it works like that the other way around. If the old way still works, it's just a normal addition.

docs/migration.rst
sillyElvy: So make opt1 required instead?

sillyElvy:
> > > > > Also why is this a breaking change? I don't see how previous methods can break?
> > > > 
> > > > Imo it is breaking because its a not backwards compatible change (aka if you add converters your code iss going to break on a prior version) 
> > > 
> > > W-What? Do any prior slash commands break because of this change? If not, it's not breaking.
> > 
> > If you add this change in your code I'm pretty sure that your code wouldn't function anymore if you use an older version - and that was my understanding of a breaking change 
> 
> No? Breaking changes mean you have to change your code if you move to this version - there's nothing stating it works like that the other way around. If the old way still works, it's just a normal addition.

Oh ok, maybe I twisted smth in my brain lol.

docs/migration.rst
AstreaTSS: Could also just rename the channel to `opt1` and the string to `opt3`.

sillyElvy: That isn't what the decorators say tho

Toricane:
Lol

interactions/client/models/command.py
Toricane: The point of a converter in this context is to get the result or value of the option passed to the function in a different way
I think it should be `passed`

sillyElvy: OK done

docs/migration.rst
Toricane: Hmm I think I messed up the decorators
Think it would make more sense if the first two were required and the third was not
Could you please change these?

interactions/client/models/command.py
Toricane: Hmm nevermind, named works

sillyElvy: Bruh I just merged it

docs/migration.rst
```diff
@@ -93,11 +93,11 @@ The following example shows and explains how to create commands effortlessly and
         # the description is the first line of the docstring or "No description set".
 
     @bot.command(default_scope=False)
-    @interactions.option(str, name="opt1")  # description is optional.
-    @interactions.option(4, name="opt2", description="This is an option.")
-    @interactions.option(interactions.Channel, name="opt3", required=True)
+    @interactions.option(str, name="opt1", required=True)  # description is optional.
+    @interactions.option(4, name="opt2", description="This is an option.", required=True)
```
Toricane, line 97:
```suggestion
    @interactions.option(4, name="opt2", description="This is an option.", converter="hi", required=True)
```

interactions/client/models/command.py
```diff
@@ -106,6 +106,7 @@ class Option(DictSerializerMixin):
     :ivar Optional[bool] autocomplete?: A status denoting whether this option is an autocomplete option.
     :ivar Optional[Dict[Union[str, Locale], str]] name_localizations?: The dictionary of localization for the ``name`` field. This enforces the same restrictions as the ``name`` field.
     :ivar Optional[Dict[Union[str, Locale], str]] description_localizations?: The dictionary of localization for the ``description`` field. This enforces the same restrictions as the ``description`` field.
+    :ivar Optional[str] converter: How the argument for the option in your function is passed, if any
```
Toricane:
```suggestion
    :ivar Optional[str] converter: How the option value is passed to the function, if different than ``name``
```
I think this makes more sense

Toricane (changes requested):
Final suggestions

docs/migration.rst
sillyElvy: fixed in another way

Toricane (approved):
LGTM!

PR #944, opened 2022-07-18
fix: Fix another bug with cache

sillyElvy:
## About

## Checklist

  - (If existent):
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix

### DIFF
--- a/interactions/api/gateway/client.py
+++ b/interactions/api/gateway/client.py
@@ -412,6 +412,8 @@ class WebSocketClient:
                         model_name = model.__name__.lower()
                         if "guild" in model_name:
                             model_name = model_name[5:]
+                        elif model_name == "threadmembers":
+                            return
                         _obj = getattr(guild, f"{model_name}s", None)
                         if _obj is not None and isinstance(_obj, list):
                             for __obj in _obj:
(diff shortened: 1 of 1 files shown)


sillyElvy:
Don't merge this I wanna add more

interactions/api/gateway/client.py
```diff
@@ -401,37 +401,76 @@ def _dispatch_event(self, event: str, data: dict) -> None:
...
+                    if model.__name__.startswith("Guild"):
+                        if model.__name__ == "GuildScheduledEventUser":
+                            id = model.guild_scheduled_event_id
+                        elif model.__name__ in [
+                            "Invite",
+                            "GuildBan",
+                            "ChannelPins",
+                            "MessageReaction",
+                            "ReactionRemove",
+                            # Extend this for everything that should not be cached
```
Toricane, line 421: Instead of doing that, wouldn't it be better to make a list of things that should be cached?
Maybe add it as a property to Cache? idk

sillyElvy: There is in general not a problem with caching everything, we just have to especially exclude the gw events that don't provide an ID

Toricane: Oh ok

Toricane (approved):
LGTM!

PR #951, opened 2022-07-21
fix: add missed logic to ``Embed.insert_field_at``

Damego:
## About

Adds missed logic to `Embed.insert_field_at` and makes `name` and `value` are required

## Checklist

  - (If existent):
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix

### DIFF
--- a/interactions/api/models/message.py
+++ b/interactions/api/models/message.py
@@ -563,7 +563,7 @@ class Embed(DictSerializerMixin):
         self.fields = []
 
     def insert_field_at(
-        self, index: int, name: str = None, value: str = None, inline: Optional[bool] = False
+        self, index: int, name: str, value: str, inline: Optional[bool] = False
     ) -> None:
         """
         Inserts a field in the embed at the specified index
--- a/interactions/api/models/message.py
+++ b/interactions/api/models/message.py
@@ -578,6 +578,17 @@ class Embed(DictSerializerMixin):
         :type inline?: Optional[bool]
         """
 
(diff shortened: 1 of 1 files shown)


interactions/api/models/message.py
```diff
@@ -578,6 +578,14 @@ def insert_field_at(
         :type inline?: Optional[bool]
         """
 
+        try:
+            fields = self.fields
+            fields.insert(index, EmbedField(name=name, value=value, inline=inline))
+            self.fields = fields
+
+        except AttributeError as e:
+            raise AttributeError("No fields found in Embed") from e
```
Toricane, line 587:
```suggestion
        try:
            self.fields.insert(index, EmbedField(name=name, value=value, inline=inline))
        except AttributeError as e:
            raise AttributeError("No fields found in Embed") from e
```

Toricane (changes requested):
A suggestion

interactions/api/models/message.py
Damego: That's doesn't work

Toricane: why not?

Damego: It does not updates `.json` ig

Toricane: Ok I see, I will resolve this

Toricane (approved):
LGTM!

Issue #966, opened 2022-07-24
[BUG] no VoiceState in interactions


Nick-the-Descended:
### Describe the bug.

I was trying to write a bot with interactions for the first time, installed it with `pip install discord-py-interactions` switching from discord.py and met with some errors. module interactions has no attribute VoiceState is being output to the console every time a voice channel has any change, eg: someone joins/leaves, mutes/unmutes...

### List the steps.

import interactions # import interactions
cli = Client(token="...") # create client
cli.start() # start client

let anyone mute/unmute in the voice channel, the error is output in the console

### What you expected.

Expected for the bot to run without any errors since it only has 3 lines of code without any functionality 

### What you saw.

Instead, I received this traceback error given from my Python terminal:
```
The HTTP client has exhausted a per-route ratelimit. Locking route for 1.0 seconds.
An error occured dispatching voice_state_update: module 'interactions' has no attribute 'VoiceState'
An error occured dispatching voice_state_update: module 'interactions' has no attribute 'VoiceState'
An error occured dispatching voice_state_update: module 'interactions' has no attribute 'VoiceState'
          ^
...
```
this doesn't crush the bot, but still annoying to see it spammed occured, also there is a type in the error (occured -> occurred) 

### What version of the library did you use?

release

### Code of Conduct

- [X] I agree to follow the contribution requirements.

sillyElvy:
Hi! This is not a bug since the library natively doesn't support voice. For voice you should have a look at this: httos://github.com/interactions-py/voice

PR #967, opened 2022-07-24
feat: Suppress voice state console logs.

FayeDel:
## About

This pull request suppresses the console spam whenever someone joins/leaves a vc, and by extension does not dispatch the Voice-related events as there's no model for it in the core library (moreover relying on the voice ext to get the events, as it doesn't interfere with the extension).

## Checklist

- [X] This fixes/solves an [Issue](https://github.com/goverfl0w/discord-interactions/issues).
  - (If existent): #966 
  - [ ] Documentation
  - [ ] Breaking change
  - [X] New feature/enhancement
  - [ ] Bugfix

### DIFF
--- a/interactions/api/gateway/client.py
+++ b/interactions/api/gateway/client.py
@@ -390,7 +390,7 @@ class WebSocketClient:
                 log.warning(
                     "Context is being created for the interaction, but no type is specified. Skipping..."
                 )
-        elif event != "TYPING_START":
+        elif event not in {"TYPING_START", "VOICE_STATE_UPDATE", "VOICE_SERVER_UPDATE"}:
             name: str = event.lower()
             try:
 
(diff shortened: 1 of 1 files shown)


interactions/api/gateway/client.py
```diff
@@ -390,7 +390,7 @@ def _dispatch_event(self, event: str, data: dict) -> None:
                 log.warning(
                     "Context is being created for the interaction, but no type is specified. Skipping..."
                 )
-        elif event != "TYPING_START":
+        elif event not in {"TYPING_START", "VOICE_STATE_UPDATE", "VOICE_SERVER_UPDATE"}:
```
sillyElvy, line 393: I think this breaks how the voice lib works, can you rather check in the error line?

sillyElvy (approved):

interactions/api/gateway/client.py
sillyElvy: aytually nvm

PR #970, opened 2022-07-24
fix: fix some bugs with cache

Damego:
## About

This pull request fixes these bugs:
- `An error occured dispatching channel_update: 'NoneType' object has no attribute 'id'`
- `on_message_update(before, after)` before always is None When you edit a message which not in cache
- Wrong getting id of models

## Checklist

  - (If existent):
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix

### DIFF
--- a/interactions/api/gateway/client.py
+++ b/interactions/api/gateway/client.py
@@ -429,7 +429,7 @@ class WebSocketClient:
                                 else Snowflake(_data["id"])
                             )
                         elif hasattr(obj, f"{model_name}_id"):
-                            id = getattr(obj, f"{model_name}_id")
+                            id = getattr(obj, f"{model_name}_id", None)
 
                 def __modify_guild_cache():
                     if not (
--- a/interactions/api/gateway/client.py
+++ b/interactions/api/gateway/client.py
@@ -440,7 +440,6 @@ class WebSocketClient:
                     ):
                         return
                     if guild := self._http.cache[Guild].get(Snowflake(guild_id)):
-                        model_name: str = model.__name__.lower()
                         model_name: str = model.__name__
                         if "guild" in model_name:
                             model_name = model_name[5:]
--- a/interactions/api/gateway/client.py
+++ b/interactions/api/gateway/client.py
@@ -417,12 +417,11 @@ class WebSocketClient:
                         "ChannelPins",
                         "MessageReaction",
                         "MessageReactionRemove",
-                        "GuildEmojis"
                         # Extend this for everything that should not be cached
                     ]:
                         id = None
                     elif model.__name__.startswith("Guild"):
-                        model_name = model.__name__[5:].lower()
+                        model_name = model.__name__[5:]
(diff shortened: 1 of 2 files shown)


Damego:
Draft until I test everything

interactions/api/gateway/client.py
```diff
@@ -409,30 +409,27 @@ def _dispatch_event(self, event: str, data: dict) -> None:
...
+                    elif model.__name__.startswith("Guild"):
+                        model_name = model.__name__[5:]
+                        if _data := getattr(obj, model_name, None):
+                            id = (
+                                getattr(_data, "id")
+                                if not isinstance(_data, dict)
+                                else Snowflake(_data["id"])
+                            )
+                        elif hasattr(obj, f"{model_name}_id"):
+                            id = getattr(obj, f"{model_name}_id")
```
sillyElvy:
```suggestion
                            id = getattr(obj, f"{model_name}_id", None)
```

interactions/api/models/gw.py
```diff
@@ -780,9 +780,9 @@ class MessageReaction(DictSerializerMixin):
     emoji: Optional[Emoji] = field(converter=Emoji, default=None)
 
 
-class ReactionRemove(MessageReaction):
+class MessageReactionRemove(MessageReaction):
```
sillyElvy, line 783: Does this still work with all events?

sillyElvy (changes requested):

interactions/api/models/gw.py
Damego: I tested all reaction events so yeah thats works

interactions/api/gateway/client.py
```diff
@@ -450,18 +447,15 @@ def __modify_guild_cache():
...
                                 _obj.append(obj)
-
                             for index, __obj in enumerate(_obj):
                                 if __obj.id == id:
                                     if "_remove" in name or "_delete" in name:
                                         _obj.remove(__obj)
 
                                     elif "_update" in name and hasattr(obj, "id"):
-                                        _obj[index] = _data
+                                        _obj[index] = obj
```
sillyElvy, line 458: ^Can you explain this changes?

sillyElvy (changes requested):

interactions/api/gateway/client.py
Damego: `_data` is None because you tries to get `"Channel"` from channel instanse.
Then in `guild.channels` instead of normal channel appears `None` and then you get error when edit channel

image

sillyElvy: I see

sillyElvy (approved):
if it works, LGTM!

Damego:
~~I will test it more tommorow. I wanna sleep~~

Damego:
I didn't get any error so we can merge it ig

PR #983, opened 2022-07-29
fix: fix error ``Attachment`` is not JSON serializable

Damego:
## About

## Checklist

  - resolves #
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix

### DIFF
--- a/interactions/api/models/message.py
+++ b/interactions/api/models/message.py
@@ -888,7 +888,7 @@ class Message(ClientSerializerMixin, IDMixin):
             _attachments = [a._json for a in attachments]
 
         if not files or files is MISSING:
-            _files = self.attachments
+            _files = [] if attachments is MISSING else [a._json for a in self.attachments]
         elif isinstance(files, list):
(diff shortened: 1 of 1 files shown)


interactions/api/models/message.py
```diff
@@ -888,7 +888,7 @@ async def edit(
             _attachments = [a._json for a in attachments]
 
         if not files or files is MISSING:
-            _files = self.attachments
+            _files = [] if attachments is MISSING else [a._json for a in self.attachments]
```
sillyElvy:
```suggestion
            _files = []
```

sillyElvy (changes requested):
already set on attachments is missing above

interactions/api/models/message.py
Damego: ~~1. If `attachments` is not missing?~~ yeah it should override then
2. Then just `_files = []` because you have below `_files.extend`

sillyElvy (approved):

PR #988, opened 2022-07-31
feat!: add ``__str__`` to ``Emoji``

Damego:
## About

This pull request adds ``__str__`` magic method to ``Emoji``.

## Checklist

  - resolves #
  - [ ] Documentation
  - [x] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix

### DIFF
--- a/interactions/api/models/emoji.py
+++ b/interactions/api/models/emoji.py
@@ -36,6 +36,13 @@ class Emoji(ClientSerializerMixin):
     animated: Optional[bool] = field(default=None)
     available: Optional[bool] = field(default=None)
 
+    def __str__(self):
+        return (
+            f"<{'a' if self.animated else ''}:{self.name}:{self.id}>"
+            if self.id is not None
+            else self.name
+        )
+
     @classmethod
     async def get(
         cls,


Damego:
I have another idea.
What if instead of magic method use `@property`?
```py
@property
def str(self): # or `as_str/to_str`
  ...
```

sillyElvy:
> I have another idea.
> What if instead of magic method use `@property`?
> ```py
> @property
> def str(self): # or `as_str/to_str`
>   ...
> ```

That won't be called automatically so I don't think that's good

Damego:
> > I have another idea.
> > What if instead of magic method use `@property`?
> > ```python
> > @property
> > def str(self): # or `as_str/to_str`
> >   ...
> > ```
> 
> That won't be called automatically so I don't think that's good

But we will see string format of emoji in `print`. Its not good too

sillyElvy:
> > > I have another idea.
> > > What if instead of magic method use `@property`?
> > > ```python
> > > @property
> > > def str(self): # or `as_str/to_str`
> > >   ...
> > > ```
> > 
> > That won't be called automatically so I don't think that's good
> 
> But we will see string format of emoji in `print`. Its not good too

Imho it should be like that on print

Damego:
okay lol

sillyElvy (approved):

PR #990, opened 2022-08-01
feat: Add support for message_delete_bulk event

mAxYoLo01:
## About

Previously, `MESSAGE_DELETE_BULK` event was not taken into account by the library as it did not have its proper class. This PR adds the class and makes sure that the event is well dispatched.
Please tell me if anything is missing, it's my first time working on the dispatcher source code.

## Checklist

  - resolves #
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [ ] Bugfix

### DIFF
--- a/interactions/api/gateway/client.py
+++ b/interactions/api/gateway/client.py
@@ -412,7 +412,8 @@ class WebSocketClient:
                         "GuildBan",
                         "ChannelPins",
                         "MessageReaction",
-                        "MessageReactionRemove",
+                        "MessageReactionRemove"
+                        "MessageDelete",
                         # Extend this for everything that should not be cached
                     ]:
                         id = None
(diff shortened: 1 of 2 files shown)


interactions/api/gateway/client.py
```diff
@@ -487,6 +488,8 @@ def __modify_guild_cache():
                     if id:
                         old_obj = _cache.pop(id)
                         self._dispatch.dispatch(f"on_{name}", old_obj)
+                    elif "_delete_bulk" in name:
+                        self._dispatch.dispatch(f"on_{name}", obj)
```
sillyElvy, line 492: This is handled already handled below no?

sillyElvy (changes requested):

interactions/api/gateway/client.py
mAxYoLo01: No, since the event **does** contain `_delete` but it does not have a proper `id` attribute.

sillyElvy (approved):

PR #994, opened 2022-08-02
fix: add member's _client for message reactions

sillyElvy:
## About

This pull request adds `_client` to the member attribute of a reaction object

## Checklist

  - resolves #
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix

### DIFF
--- a/interactions/api/models/gw.py
+++ b/interactions/api/models/gw.py
@@ -777,7 +777,7 @@ class MessageDelete(DictSerializerMixin):
 
 
 @define()
-class MessageReaction(DictSerializerMixin):
+class MessageReaction(ClientSerializerMixin):
     """
     A class object representing the gateway event ``MESSAGE_REACTION_ADD`` and ``MESSAGE_REACTION_REMOVE``.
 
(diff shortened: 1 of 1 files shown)


Damego:
Why you don't use `add_client=True` in the `field`?

interactions/api/models/gw.py
```diff
@@ -796,6 +796,9 @@ class MessageReaction(DictSerializerMixin):
     member: Optional[Member] = field(converter=Member, default=None)
     emoji: Optional[Emoji] = field(converter=Emoji, default=None)
 
+    def __attrs_post_init__(self):
+        self.member._client = self._client
```
Damego: discord doesn't send `member` data in `message_reaction_remove` event so it will raise error `An error occurred dispatching message_reaction_remove: 'NoneType' object has no attribute '_client'`.
Use `add_client=True`

Damego (changes requested):

interactions/api/models/gw.py
sillyElvy: Done, didnt know it exists

Damego (approved):

Catalyst4222 (approved):

PR #995, opened 2022-08-02
feat: Add `__str__` to `Application`

CaedenPH:
## About

## Checklist

  - resolves #
  - [ ] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix


mAxYoLo01:
Please make sure to track the `unstable` branch, currently it is the `stable` one.

sillyElvy (changes requested):
> This pull request is about (X), which does (Y).

Also please explain why you want this change since I don't know anything about the intentions of this PR/why the change should be implemented/...


Also, as already mentioned, please target the `unstable` branch for your pr

CaedenPH:
> > This pull request is about (X), which does (Y).
> 
> Also please explain why you want this change since I don't know anything about the intentions of this PR/why the change should be implemented/...
> 
> Also, as already mentioned, please target the `unstable` branch for your pr

When printing `client.me`, a user is left with an unfriendly `Application` instance, this would mean that it prints the username.

Catalyst4222:
A better way to give more information about it would be to add a `__repr__` to the class, or to the user field. Using `__str__` to give information when printing data usually isn't the best way to do it

Catalyst4222:
> A better way to give more information about it would be to add a `__repr__` to the class, or to the user field. Using `__str__` to give information when printing data usually isn't the best way to do it

`attr_utils.field` has a `repr` param, if you set that to True it'll show up when you print it out. When I add `id` and `name` to the repr, it prints (in my case) as `Application(id=Snowflake(835925027337273354), name='CataBot 2.0')`, which is more descriptive without tricking the user into thinking it's a string

sillyElvy (changes requested):
Put `repr=True` into `@define()` above the application class as suggested.

Failure to do this within 3 days will resolve the PR being closed (due to inactivity)

sillyElvy:
Closing in favour of #1026

PR #1010, opened 2022-08-10
fix(autocomplete dispatch): ID -> name 

sillyElvy:
This allows multi-guild usage when 2 or more scopes are used on a command since the command will have a different ID but the same name in different guilds. Reported [here](https://canary.discord.com/channels/789032594456576001/1006824593731887214/1006838362772078743)

## About

## Checklist

  - resolves #
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix

### DIFF
--- a/interactions/api/gateway/client.py
+++ b/interactions/api/gateway/client.py
@@ -321,7 +321,7 @@ class WebSocketClient:
 
                     self._dispatch.dispatch("on_component", _context)
                 elif data["type"] == InteractionType.APPLICATION_COMMAND_AUTOCOMPLETE:
-                    _name = f"autocomplete_{_context.data.id}"
+                    _name = f"autocomplete_{_context.data.name}"
 
                     if _context.data._json.get("options"):
                         for option in _context.data.options:
--- a/interactions/client/bot.py
+++ b/interactions/client/bot.py
@@ -862,7 +862,8 @@ class Client:
             and command.type == ApplicationCommandType.CHAT_INPUT
         ):
             raise LibraryException(
-                11, message=f"Your command name ('{command.name}') does not match the regex for valid names ('{regex}')."
+                11,
+                message=f"Your command name ('{command.name}') does not match the regex for valid names ('{regex}').",
             )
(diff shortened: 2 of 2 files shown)


Damego:
In this case if i trigger autocomplete in first guild both funtions with different command ids will be called

image
image

sillyElvy:
> In this case if i trigger autocomplete in first guild both funtions with different command ids will be called
> 
> image
> image
>  

You want to have different responses for different guilds with the same command?

Damego:
> You want to have different responses for different guilds with the same command?

Yes! /s
I think it should call only command with equal command id

sillyElvy:
> > You want to have different responses for different guilds with the same command?
> 
> Yes! /s
> I think it should call only command with equal command id

Is there a real need for that?

Damego:
> Is there a real need for that?
Honestly idk. Just found this.

sillyElvy:
> > Is there a real need for that?
> Honestly idk. Just found this.

I think we should just keep it as it is rn

Damego (approved):
So it works fine for me (tested in 2 guilds)

Toricane:
It does make sense that the command with equivalent ID invokes only the autocomplete for it since the ID was given
I can imagine some sort of tag system or something where the command names are the same but autocomplete differently in different guilds
Would it be difficult to implement it? If it is, then I think it's fine

sillyElvy:
> It does make sense that the command with equivalent ID invokes only the autocomplete for it since the ID was given
> I can imagine some sort of tag system or something where the command names are the same but autocomplete differently in different guilds
> Would it be difficult to implement it? If it is, then I think it's fine

Yeah it's kinda hard to do it/it requires effort which is not really nessecary, especially considering the fact that you can just check `ctx.guild_id` inside autocomplete i think?

Toricane:
Ok, makes sense

Toricane (changes requested):
Please resolve merge conflict

Toricane (approved):
LGTM!

PR #1012, opened 2022-08-10
refactor: optimize how autodefer works

sillyElvy:
## About
see above

## Checklist

  - resolves #
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix


interactions/client/models/utils.py
```diff
@@ -50,7 +61,16 @@ async def deferring_func(ctx: Union["CommandContext", "ComponentContext"], *args
                 loop = get_running_loop()
             except RuntimeError as e:
                 raise RuntimeError("No running event loop detected!") from e
-            task: Task = loop.create_task(coro(ctx, *args, **kwargs))
+
+            if "self" in getfullargspec(coro).args:
```
Toricane: This method isn't reliable since users may name the `self` parameter something else.
Could you try using the qualname trick?
Iirc, you check if the dunder qualname (of the `coro` or of the `coro.__func__`, i forgot) has a "." in it.
I think Astrea used it in molter

Toricane (changes requested):
The method isn't reliable

Also please add the Extension typehint for `ctx`

interactions/client/models/utils.py
sillyElvy: I think I will make autodefer run completely different

sillyElvy: nvm.

Catalyst4222 (dismissed):

Catalyst4222 (approved):

interactions/client/models/utils.py
```diff
@@ -41,16 +52,24 @@ async def command(ctx):
...
-            task: Task = loop.create_task(coro(ctx, *args, **kwargs))
+
+            if isinstance(args[0], (ComponentContext, CommandContext)):
+                self = ctx
+                ctx = list(args).pop(0)
+
+                task: Task = loop.create_task(coro(self, ctx, *args, **kwargs))
+
+            else:
+                task: Task = loop.create_task(coro(ctx, *args, **kwargs))
```
Toricane: This is unnecessary, they will all get passed as normal

Toricane (changes requested):
Unnecessary code

interactions/client/models/utils.py
Catalyst4222: It gets the context, since it's needed later to defer. If it doesn't check for it, there will be issues with trying to defer an extension

Toricane: Oh I see now

Toricane (approved):
LGTM!

PR #1014, opened 2022-08-10
refactor(bot, gateway, http): Remove any unclosed client session messages

sillyElvy:
## About

This pull request aims to stop "unclosed client session" messages to appear after stopping the bot

## Checklist

  - resolves #
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [ ] Bugfix

### DIFF
--- a/interactions/api/gateway/client.py
+++ b/interactions/api/gateway/client.py
@@ -773,3 +773,4 @@ class WebSocketClient:
         """
         if self._client:
             await self._client.close()
+        self._closed = True
--- a/interactions/api/gateway/client.py
+++ b/interactions/api/gateway/client.py
@@ -112,7 +112,7 @@ class WebSocketClient:
         except RuntimeError:
             self._loop = new_event_loop()
         self._dispatch: Listener = Listener()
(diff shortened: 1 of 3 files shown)


interactions/api/gateway/client.py
```diff
@@ -112,7 +112,7 @@ def __init__(
         except RuntimeError:
             self._loop = new_event_loop()
         self._dispatch: Listener = Listener()
-        self._http: HTTPClient = HTTPClient(token)
```
FayeDel, line 141: I don't see why changing the http attribute fixes the unclosed client session message tbh

From my testing it pointed from the weird global-esque session definition, http variable was just fine?

sillyElvy: the attrib indeed causes a unclosed client session message, the global session definition caused smth other that was unclosed

sillyElvy: (it only happens when you don't start the bot but yeah

Catalyst4222 (approved):

PR #1018, opened 2022-08-13
feat: add helper methods for component disabling

sillyElvy:
## About

This pull request adds methods to disable components

## Checklist

  - resolves #
  - [ ] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix

### DIFF
--- a/interactions/client/context.py
+++ b/interactions/client/context.py
@@ -265,7 +265,7 @@ class _Context(ClientSerializerMixin):
 
         return payload
 
-    async def popup(self, modal: Modal) -> None:
+    async def popup(self, modal: Modal) -> dict:
         """
         This "pops up" a modal to present information back to the
         user.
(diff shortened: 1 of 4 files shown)


interactions/client/context.py
```diff
@@ -654,6 +654,37 @@ async def defer(
...
+                for component in components.components:
+                    component.disabled = True
+
+            if other_kwargs.get("components"):
+                raise LibraryException(
+                    12, "You must not specify the `components` argument in this method."
+                )
+
+            other_kwargs["components"] = self.message.components
+            await self.edit(**other_kwargs)
```
Catalyst4222: You have the return type as a Message, but if the user chooses to respond to the interaction, nothing gets returned

sillyElvy: fixed

Toricane (commented):
Could you also add a utility function that takes in a list of action rows or components and disables them all?
I think that would be useful as well

sillyElvy:
> Could you also add a utility function that takes in a list of action rows or components and disables them all? I think that would be useful as well
Done

Catalyst4222 (approved):

PR #1023, opened 2022-08-20
ref!: set GuildMember as a subclass of Member

mAxYoLo01:
## About

This pull request ensures that in the future, `Member` & `GuildMember` have the same attributes & functions.
As of now, there were still inconsistencies between the two.

## Checklist

  - resolves #
  - [ ] Documentation
  - [x] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix


sillyElvy (changes requested):
Revert method removal, the overridden methods are needed (for now) but I will make a PR some day

Also it's not breaking

mAxYoLo01:
> Revert method removal, the overridden methods are needed (for now) but I will make a PR some day

Could you  explain why? I'll revert for the moment.

> Also it's not breaking

For me it is, as `roles` passes from a list of str to int.

sillyElvy:
> > Revert method removal, the overridden methods are needed (for now) but I will make a PR some day
> 
> Could you  explain why? I'll revert for the moment.
> 
> > Also it's not breaking
> 
> For me it is, as `roles` passes from a list of str to int.
> 
> 

Guild ID


Ok that's really fair

Damego:
image

sillyElvy:
And that has to be reverted

mAxYoLo01:
Yeah my bad that's what happens when you play with 5 branches at the same time and decide to stage all files

sillyElvy:
> Yeah my bad that's what happens when you play with 5 branches at the same time and decide to stage all files

Me who uses pycharm github integration: your problems are beyond my understanding

sillyElvy (approved):

PR #1025, opened 2022-08-20
feat!: Add a guild id property for the member object

sillyElvy:
## About

This pull request adds a guild ID property for the member object

## Checklist

  - resolves #
  - [ ] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix


interactions/api/models/member.py
```diff
@@ -62,6 +63,78 @@ class Member(ClientSerializerMixin, IDMixin):
...
+            elif len(possible_guilds) == 0:
+                possible_guilds = list(self._client.cache[Guild].values.values())
+
+            for guild in possible_guilds:
+                for member in guild.members:
+                    if member.joined_at == self.joined_at:
+                        return guild.id
+
+            else:
+                return None
```
mAxYoLo01:
```suggestion
    @property
    def guild_id(self) -> Optional[Snowflake]:
        """
        Attempts to get the guild ID the member is in.
        Only works then roles or nick or joined at is present.
        :return: The ID of the guild this member belongs to.
        """

        if hasattr(self, "guild_id"):
            return self.guild_id

        if not self._client:
            raise LibraryException(code=13)

        if self.roles:
            for guild in self._client.cache[Guild].values.values():
                for role in guild.roles:
                    if role.id in self.roles:
                        return guild.id

        possible_guilds: List[Guild] = []

        if self.nick:
            def check(member: Member):
                    return member.nick == self.nick and member.id == self.id

            for guild in self._client.cache[Guild].values.values():   
                if len(search_iterable(guild.members, check=check)) > 0:
                    possible_guilds.append(guild)

        if len(possible_guilds) == 1:
            return possible_guilds[0].id

        elif len(possible_guilds) == 0:
            possible_guilds = list(self._client.cache[Guild].values.values())

        for guild in possible_guilds:
            for member in guild.members:
                if member.joined_at == self.joined_at:
                    return guild.id

        else:
            return None
```

Catalyst4222 (approved):

PR #1039, opened 2022-08-25
refactor(cache): optimize caching and dispatching behaviour

Damego:
## About

This pull request:
1. Moves getting object ID into separated method.
2. Adds method to get object IDs. It needs for `message_delete_bulk`, `emojis_update` and `stickers_update` events. It allows properly cache emojis and stickers.
3. Caches `GuildMember` like `Member`
4. Deletes messages from cache on `message_delete_bulk` event.
5. Moves `__modify_guild_cache` to own method.
6. Makes properly cache `Guild.members` and `Guild.roles`.
7. Properly dispatches remove/delete event so we no need to use raw remove/delete event.
8. Dispatches `_update` event if there are no id and ids. It allows use `on_channel_pins_update`, `on_webhooks_update` events without `raw` prefix.

## Checklist

  - resolves #
  - [x] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix

### DIFF
--- a/interactions/api/gateway/client.py
+++ b/interactions/api/gateway/client.py
@@ -441,7 +441,6 @@ class WebSocketClient:
                 model = getattr(__import__(path), _name)
                 obj = model(**data)
 
-                # I don't like this but idk what i should do then
                 guild_obj = guild_model = None
                 if model is GuildRole:
                     guild_obj = Role(**role_data) if (role_data := data.get("role")) else None
@@ -452,7 +451,6 @@ class WebSocketClient:
 
                 _cache: "Storage" = self._http.cache[model]
                 _guild_cache: "Storage" = self._http.cache[guild_model]
-                # in _guild_cache stores 'Role' and 'Member' objects
 
                 ids = None
                 id = self.__get_object_id(data, obj, model)
@@ -473,6 +471,7 @@ class WebSocketClient:
 
                 elif "_update" in name:
                     self._dispatch.dispatch(f"on_raw_{name}", obj)
+
                     if not id and ids is None:
                         return self._dispatch.dispatch(f"on_{name}", obj)
 
@@ -480,11 +479,11 @@ class WebSocketClient:
                         name, data, guild_model or model, guild_obj or obj, id, ids
                     )
                     if ids is not None:
-                        # Not cached but it needed for events guild_emojis_update and guild_stickers_update
-                        self._dispatch.dispatch(f"on_{name}", obj)
-                        return
+                        # Not cached but it needed for guild_emojis_update and guild_stickers_update events
+                        return self._dispatch.dispatch(f"on_{name}", obj)
                     if id is None:
                         return
+
                     if guild_obj:
                         old_guild_obj = _guild_cache.get(id)
                         if old_guild_obj:
--- a/interactions/api/gateway/client.py
+++ b/interactions/api/gateway/client.py
@@ -516,7 +516,9 @@ class WebSocketClient:
 
                     elif ids and "message" in name:
                         # currently only message has '_delete_bulk' event but ig better keep this condition for future.
-                        self.__delete_message_cache(ids)
+                        _message_cache: "Storage" = self._http.cache[Message]
+                        for message_id in ids:
+                            _message_cache.pop(message_id)
 
                     self._dispatch.dispatch(f"on_{name}", old_obj or obj)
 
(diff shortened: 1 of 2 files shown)


sillyElvy (changes requested):
Add emojis to global cache in the guild obj

interactions/api/gateway/client.py
```diff
@@ -441,55 +443,35 @@ def _dispatch_event(self, event: str, data: dict) -> None:
...
-                            id = (
-                                getattr(_data, "id")
-                                if not isinstance(_data, dict)
-                                else Snowflake(_data["id"])
-                            )
-                        elif hasattr(obj, f"{model_name}_id"):
-                            id = getattr(obj, f"{model_name}_id", None)
+                    ids = self.__get_object_ids(obj, model)
 
                 def __modify_guild_cache():
```
Catalyst4222: While you're optimizing, can you move `__modify_guild_cache` into its own function?

interactions/api/models/guild.py
```diff
@@ -360,6 +360,8 @@ def __attrs_post_init__(self):  # sourcery skip: last-if-guard
                     self.member_count = guild.member_count
                 if not self.presences:
                     self.presences = guild.presences
+                if not self.emojis:
+                    self.emojis = guild.emojis
```
sillyElvy, line 364: Please also add this to adding it to the global cache as above done with other objects

sillyElvy (changes requested):

interactions/api/models/guild.py
Damego: Emojis doesn't exists in global cache.

interactions/api/gateway/client.py
Damego: Sure

interactions/api/models/guild.py
sillyElvy: Why would they not?

Damego: > Why would they not? 

Discord has only one event with emoji. Its `guild_emojis_update`. On unstable it doesn't gets emojis and not stores them in cache.
To cache emojis we should get emojis from `GuildEmojis` object and do some stuff with cache. Same for stickers.

interactions/api/gateway/client.py
Damego: Done

interactions/api/gateway/client.py
```diff
@@ -435,83 +438,51 @@ def _dispatch_event(self, event: str, data: dict) -> None:
...
-                                        _obj[index] = obj
-                                    break
-                            setattr(guild, f"{model_name}s", _obj)
-                        self._http.cache[Guild].add(guild)
+                    ids = self.__get_object_ids(obj, model)
+
+                # TODO: Move to own method?
+                guild_obj = None
+                if guild_model is Role:
+                    guild_obj = Role(**data["role"]) if "role" in data else _guild_cache.get(id)
```
Damego: `guild_role_delete` doesn't contain `role` in data so i took cached state

Damego:
Looks like it works fine for me

interactions/api/gateway/client.py
```diff
@@ -431,87 +434,61 @@ def _dispatch_event(self, event: str, data: dict) -> None:
...
                 model = getattr(__import__(path), _name)
-
-                data["_client"] = self._http
                 obj = model(**data)
 
-                _cache: "Storage" = self._http.cache[model]
+                # I don't like this but idk what i should do then
+                guild_obj = guild_model = None
+                if model is GuildRole:
+                    guild_obj = Role(**data["role"]) if "role" in data else None
```
sillyElvy:
```suggestion
                    guild_obj = Role(**role_data) if (role_data := data.get("role")) else None
```

sillyElvy (changes requested):

interactions/api/gateway/client.py
Damego: Then i suggest use walrus operator

sillyElvy: Idek if that works there

Damego: This works
```py
guild_obj = Role(**role_data) if (role_data := data.get("role")) else None
```

sillyElvy: Modified suggestion

AstreaTSS (approved):
Haven't verified the code works, but it looks solid.

Damego:
New changes:
- ~~Add support for `_execution` event.~~ No need lol. I forgot we have `else` block
- Dispatches `_update` event if there are no id and ids. It allows use `on_channel_pins_update`, `on_webhooks_update` events without `raw` prefix.

sillyElvy (approved):

PR #1040, opened 2022-08-26
refactor: make Command.dispatch a function

Catalyst4222:
## About

This pull request changes Command.dispatch from a property into a function
This is mostly just to make it easier to override in exts, though it does help to simplify the code

## Checklist

  - resolves #
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [ ] Bugfix

### DIFF
--- a/interactions/client/models/command.py
+++ b/interactions/client/models/command.py
@@ -711,13 +711,17 @@ class Command(DictSerializerMixin):
         **kwargs,
     ) -> Optional[BaseResult]:
         """
-        Returns a coroutine that calls the command along with the subcommands, if any.
+        Call the command along with any subcommands
 
-        .. note::
-            The coroutine returned is never the same object.
-
-        :return: A coroutine that calls the command along with the subcommands, if any.
-        :rtype: Callable[..., Awaitable]
+        :param ctx: The context for the interaction
+        :type ctx: CommandContext
+        :param args: The args to be passed to the command
+        :param sub_command_group: The subcommand group being invoked, if any
+        :type sub_command_group: Optional[str]
+        :param sub_command: The subcommand being invoked, if any
+        :type sub_command: Optional[str]
+        :param kwargs: The kwargs to pass to the command
+        :return:
         """
         base_coro = self.coro
         base_res = BaseResult(
--- a/interactions/client/models/command.py
+++ b/interactions/client/models/command.py
@@ -702,8 +702,14 @@ class Command(DictSerializerMixin):
 
         return decorator
 
-    @property
-    def dispatcher(self) -> Callable[..., Awaitable]:
+    async def dispatcher(
+        self,
+        ctx: "CommandContext",
+        *args,
+        sub_command_group: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


interactions/client/models/command.py
```diff
@@ -702,8 +702,14 @@ def decorator(coro: Callable[..., Awaitable]) -> "Command":
 
         return decorator
 
-    @property
-    def dispatcher(self) -> Callable[..., Awaitable]:
+    async def dispatcher(
```
Toricane, line 705: Update the docstring

Toricane (changes requested):
Please update the docstring

interactions/client/models/command.py
Catalyst4222: Done

interactions/client/models/command.py
```diff
@@ -702,58 +702,54 @@ def decorator(coro: Callable[..., Awaitable]) -> "Command":
...
+        :param ctx: The context for the interaction
+        :type ctx: CommandContext
+        :param args: The args to be passed to the command
+        :param sub_command_group: The subcommand group being invoked, if any
+        :type sub_command_group: Optional[str]
+        :param sub_command: The subcommand being invoked, if any
+        :type sub_command: Optional[str]
+        :param kwargs: The kwargs to pass to the command
+        :return:
         """
```
Toricane: Add period at end of the function description sentence.
Make the opening quotes `r"""`.
Turn `args` into `\*args`.
Turn `kwargs` into `\**kwargs`.
Add `:type \*args: Tuple[Any]`
Add `:type \**kwargs: Dict[Any]`
Add descriptions for `:rtype:` and `:return:`.

Toricane (changes requested):
More changes required

interactions/client/models/command.py
Catalyst4222: Args shouldn't be typehinted as a tuple, that would imply each argument in it is a tuple

Toricane: There are no examples of documentation of args that I can point towards, but there are examples of kwargs being typehinted as dict in the docstrings, such as the ones in bot.py, so I think args should be tuple

Catalyst4222: The typehint following `*args` is what's inside the tuple

Here the typehints are for a tuple (python 3.10, same case for earlier versions)
image

If it's set to only `int`, then the typehints for each item in `args` is for ints
image

Toricane: > but there are examples of kwargs being typehinted as dict in the docstrings, such as the ones in bot.py, so I think args should be tuple

I'm not talking about typehints in the function parameters, I'm talking about the docstring. Maybe the types on the docstrings can be changed in a separate PR since they're all over the place.

Catalyst4222: The types in docstrings should match the function parameters. For now, I'll match the docstring to the rest of the library, but cleaning up docstring parameters is needed at some point

Toricane (approved):
LGTM!

PR #1046, opened 2022-08-30
fix: command options and choices doesn't syncs

Damego:
## About

This pull request fixes bug command options/choices doesn't syncs when you tryies to add/update/remove options/choice.

Reason of this bug is in overriding `_command` variable on line 259.
Also this pr removes useless return data

## Checklist

  - resolves #
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix

### DIFF
--- a/interactions/client/bot.py
+++ b/interactions/client/bot.py
@@ -194,7 +194,7 @@ class Client:
             if any(option not in _command_option_names for option in _data_option_names) or len(
                 _data_option_names
             ) != len(_command_option_names):
-                return False, command
+                return False
 
             for option in command.get("options"):
                 for _option in data.get("options"):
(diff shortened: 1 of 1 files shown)


sillyElvy (changes requested):
Was this tested properly and with enough commands? I remember having issues before adding that

Damego:
> Was this tested properly

I adds, updates, deletes options and choices in all kinds(maybe?) of commands(groups and subcommands) and it works fine.

interactions/client/bot.py
```diff
@@ -301,8 +300,7 @@ def __check_options(command, data):
 
                         elif command.get("options") and data.get("options"):
 
-                            clean, _command = __check_options(command, data)
-
+                            clean = __check_options(command, data)
```
Damego, line 303:
```suggestion
                            clean = __check_options(command, data)

```

sillyElvy (approved):
If it works it works, yay

PR #1053, opened 2022-08-31
fix: Mitigate gateway null responses.

FayeDel:
## About

This pull request properly checks whether the websocket stream exists such that dispatching a None object wouldn't happen and reorders the sequence when the heartbeat manager resets so that no heartbeats are sent when the WS client is None.

More testers are welcome :)

## Checklist

  - resolves #
  - [ ] Documentation
  - [ ] Breaking change
  - [X] New feature/enhancement
  - [X] Bugfix

### DIFF
--- a/interactions/api/gateway/client.py
+++ b/interactions/api/gateway/client.py
@@ -807,8 +807,10 @@ class WebSocketClient:
                 msg = loads(packet.data)
             except Exception as e:
                 import traceback
+
                 log.debug(
-                    f'Error serialising message: {"".join(traceback.format_exception(type(e), e, e.__traceback__))}.')
+                    f'Error serialising message: {"".join(traceback.format_exception(type(e), e, e.__traceback__))}.'
+                )
                 # There's an edge case when the packet's None... or some other deserialisation error.
                 # Instead of raising an exception, we just log it to debug, so it doesn't annoy end user's console logs.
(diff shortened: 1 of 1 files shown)


interactions/api/gateway/client.py
```diff
@@ -800,7 +803,19 @@ async def __receive_packet(self, ignore_lock: bool = False) -> Optional[Dict[str
...
 
-            return loads(packet.data) if isinstance(packet.data, str) else None
+            try:
+                msg = loads(packet.data)
+            except Exception as e:
+                import traceback
+
+                log.debug(
+                    f'Error serialising message: {"".join(traceback.format_exception(type(e), e, e.__traceback__))}.'
+                )
```
sillyElvy, line 813: Delta, again? (log.exception)

sillyElvy (changes requested):
This one + what about client being None in heartbeat or change presence?

interactions/api/gateway/client.py
FayeDel: did you uhh
read the comment below the line- :P

sillyElvy: No I did not i was in a rush xD 

sorry

sillyElvy (approved):
lgtm

PR #1057, opened 2022-09-01
feat: Add base iterators & context managers + helpers

sillyElvy:
## About

This pull request implements iterators and context managers,

Used for channel typing, channel history and guild members

## Checklist

  - resolves #
  - [x] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix

### DIFF
--- a/interactions/api/models/channel.py
+++ b/interactions/api/models/channel.py
@@ -176,8 +176,9 @@ class AsyncHistoryIterator(BaseAsyncIterator):
 
         self.objects = [Message(**msg, _client=self._client) for msg in msgs]
 
-    def flatten(self):
-        return NotImplemented
+    async def flatten(self) -> List[Member]:
+        """returns all remaining items as list""" 
+        return [item async for item in self]
 
     async def get_objects(self) -> None:
         from .message import Message
--- a/interactions/api/models/channel.py
+++ b/interactions/api/models/channel.py
@@ -6,7 +6,7 @@ from typing import TYPE_CHECKING, Any, Awaitable, Callable, ContextManager, List
 from warnings import warn
 
 from ...utils.abc.base_context_managers import BaseAsyncContextManager
-from ...utils.abc.base_iterators import BaseAsyncIterator
+from ...utils.abc.base_iterators import DiscordPaginationIterator
 from ...utils.attrs_utils import (
     ClientSerializerMixin,
     DictSerializerMixin,
@@ -102,7 +102,7 @@ class ThreadMember(ClientSerializerMixin):
     mute_config: Optional[Any] = field(default=None)  # todo explore this, it isn't in the ddev docs
 
 
-class AsyncHistoryIterator(BaseAsyncIterator):
+class AsyncHistoryIterator(DiscordPaginationIterator):
     """
     A class object that allows iterating through a channel's history.
 
--- a/interactions/api/models/channel.py
+++ b/interactions/api/models/channel.py
@@ -1,4 +1,4 @@
-from asyncio import create_task, sleep
+from asyncio import create_task, sleep, Task
 from datetime import datetime, timedelta, timezone
 from enum import IntEnum
 from math import inf
@@ -252,7 +252,15 @@ class AsyncTypingContextManager(BaseAsyncContextManager):
         obj: Union[int, str, "Snowflake", "Channel"],
         _client: "HTTPClient",
     ):
-        super().__init__(obj, _client)
+        
+        try:
+            self.loop = get_running_loop()
+        except RuntimeError as e:
+            raise RuntimeError("No running event loop detected!") from e
+
+        self.object_id = None if not obj else int(obj) if not hasattr(obj, "id") else int(obj.id)
+        self._client = _client
+        self.__task: Optional[Task] = None
 
     def __await__(self):
         return self._client.trigger_typing(self.object_id).__await__()
--- a/interactions/api/models/channel.py
+++ b/interactions/api/models/channel.py
@@ -1,6 +1,7 @@
 from datetime import datetime, timedelta, timezone
 from enum import IntEnum
 from typing import TYPE_CHECKING, Any, Callable, List, Optional, Union
+from warnings import warn
 
 from ...utils.attrs_utils import (
     ClientSerializerMixin,
@@ -18,6 +19,7 @@ from .webhook import Webhook
 
 if TYPE_CHECKING:
     from ...client.models.component import ActionRow, Button, SelectMenu
+    from ..http.client import HTTPClient
     from .guild import Invite, InviteTargetType
     from .member import Member
     from .message import Attachment, Embed, Message, Sticker
@@ -28,6 +30,7 @@ __all__ = (
     "Channel",
     "ThreadMember",
     "ThreadMetadata",
+    "AsyncHistoryIterator",
 )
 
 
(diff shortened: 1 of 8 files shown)


interactions/utils/abc/base_iterators.py
```diff
@@ -0,0 +1,97 @@
...
+class BaseAsyncIterator(metaclass=ABCMeta):
+    """A base class for async iterators."""
+
+    # I don't want to make it subclass the BaseIterator since it forces implementation of __next__ and __iter__
+
+    @abstractmethod
+    def __init__(
+        self,
+        obj: Union[int, str, "Snowflake", _T],
+        _client: Optional["HTTPClient"] = None,
```
Catalyst4222: Can you make it so that obj, client, etc. aren't implemented in the base async iterator? There are use cases where none of the parameters will be needed

interactions/utils/abc/base_iterators.py
```diff
@@ -0,0 +1,97 @@
...
+            if start_at is MISSING
+            else int(start_at)
+            if not hasattr(start_at, "id")
+            else int(start_at.id)
+        )
+        self.__stop: bool = False
+        self.objects: Optional[List[_O]] = None
+
+    @abstractmethod
+    async def get_first_objects(self) -> None:
```
Catalyst4222: Not every iterator will have a difference between getting the first object and getting the subsequent objects, or even need another method to do so

interactions/utils/abc/base_iterators.py
```diff
@@ -0,0 +1,97 @@
...
+_O = TypeVar("_O")
+
+__all__ = ("BaseAsyncIterator", "BaseIterator")
+
+if TYPE_CHECKING:
+    from ...api.http.client import HTTPClient
+    from ...api.models.misc import Snowflake
+
+
+class BaseAsyncIterator(metaclass=ABCMeta):
```
Catalyst4222, line 17: Can you add a flatten method to completely consume the generator and return the list of results?

interactions/utils/abc/base_iterators.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def __aiter__(self):
+        return self
+
+    @abstractmethod
+    async def __anext__(self) -> _O:
+        raise NotImplementedError
+
+
+class BaseIterator(metaclass=ABCMeta):
```
Catalyst4222: See previous comments about flatten, parameters, and methods to get objects

interactions/utils/abc/base_context_managers.py
```diff
@@ -0,0 +1,63 @@
...
+        self.object_id = int(obj) if not hasattr(obj, "id") else int(obj.id)
+        self._client = _client
+        self.__task: Optional[Task] = None
+
+    @abstractmethod
+    def __await__(self) -> None:
+        raise NotImplementedError
+
+    @abstractmethod
+    async def do_action(self):
```
Catalyst4222: `do_action` is a rather ambiguous name. If it's required for *every* context manager, why isn't it in the sync context manager? If it's not required, why is it in the base class?

interactions/utils/abc/base_context_managers.py
```diff
@@ -0,0 +1,63 @@
...
+    from ...api.http.client import HTTPClient
+    from ...api.models.misc import Snowflake
+
+
+class BaseAsyncContextManager(metaclass=ABCMeta):
+    """A base class for async iterators."""
+
+    # I don't want to make it subclass the BaseContextManager since it forces implementation of __enter__ and __exit__
+
+    @abstractmethod
```
Catalyst4222, line 14: See other comment on parameters

interactions/api/models/channel.py
```diff
@@ -1122,6 +1329,10 @@ async def get_history(self, limit: int = 100) -> Optional[List["Message"]]:
         :rtype: List[Message]
         """
 
+        warn(
```
Catalyst4222, line 1345: Why does this need to be deprecated? The method has nothing wrong with using it, and removing it doesn't simplify code elsewhere

interactions/api/models/guild.py
```diff
@@ -2078,6 +2181,11 @@ async def get_all_members(self) -> List[Member]:
         :return: Returns a list of all members of the guild
         :rtype: List[Member]
         """
+        warn(
```
Catalyst4222, line 2358: See above comment about depreciation

Catalyst4222 (changes requested):

interactions/utils/abc/base_iterators.py
sillyElvy: You can just leave one of the methods with an ellipsis defined in your class

interactions/utils/abc/base_iterators.py
sillyElvy: I can make them none but I don't want to remove them since the objects are intended to be used internally hence the design, it just saves like 20-30 lines of duplicate code

Catalyst4222: It also prevents them from being used internally for purposes other than pagination from discord

interactions/utils/abc/base_context_managers.py
sillyElvy: I guess I can make it not abstract, it's intended for background tasks (like typing)

interactions/utils/abc/base_iterators.py
sillyElvy: How so if they're optional?

interactions/api/models/channel.py
sillyElvy: It allows less control than the new method (for example if you want to find a specific message but don't know where it is you'll have to get all set messages) and imo it's just better to use the new method

I also compared them, the new one's were around 5 seconds faster than the deprecated

interactions/utils/abc/base_iterators.py
Catalyst4222: You can't leave the method unoverriden without errors. Adding the function with no code can easily lead to confusion

sillyElvy: Alright I'll un abstract this too

interactions/utils/abc/base_context_managers.py
```diff
@@ -0,0 +1,42 @@
...
+    def __init__(self):
+        raise NotImplementedError
+
+    @abstractmethod
+    def __await__(self) -> None:
```
Catalyst4222: What's your expectation of how `__await__` will be used?

interactions/utils/abc/base_iterators.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @abstractmethod
+    def __init__(self):
+        raise NotImplementedError
+
+    def __aiter__(self):
+        return self
+
+    @abstractmethod
+    async def flatten(self):
```
Catalyst4222, line 29: Can you make it so that `flatten` consumes the rest of the iterator and returns the list of items? It should be a common enough use case to be implemented as a concrete method (same in the sync iterator too)

interactions/utils/utils.py
```diff
@@ -231,3 +246,64 @@ def disable_components(
...
+    :type reverse?: Optional[bool]
+    :param check?: A check to ignore specific messages
+    :type check?: Optional[Callable[[Message], bool]]
+    :param maximum?: A set maximum of messages to get before stopping the iteration
+    :type maximum?: Optional[int]
+
+    :return: An asynchronous iterator over the history of the channel
+    :rtype: AsyncHistoryIterator
+    """
+    return AsyncHistoryIterator(
```
Catalyst4222, line 279: Would it be better to do a local import of the history iterator? If you do that, then it will keep it so that the `utils` submodule doesn't depend on any other parts of the library during initialization

interactions/api/models/channel.py
```diff
@@ -95,6 +102,181 @@ class ThreadMember(ClientSerializerMixin):
...
+    async def __anext__(self) -> "Message":
+        if self.objects is None:
+            await self.get_first_objects()
+
+        try:
+            obj = self.objects.pop(0)
+
+            if self.check:
+
+                _res = self.check(obj)
```
Catalyst4222: Can you add support for async checks? It should just be `if isawaitable(res): res = await res`

interactions/api/models/guild.py
```diff
@@ -210,6 +215,108 @@ class UnavailableGuild(DictSerializerMixin, IDMixin):
...
+    async def __anext__(self) -> Member:
+        if self.objects is None:
+            await self.get_first_objects()
+
+        try:
+            obj = self.objects.pop(0)
+
+            if self.check:
+
+                _res = self.check(obj)
```
Catalyst4222: See above comment about async checks

Catalyst4222 (changes requested):

interactions/api/models/channel.py
AstreaTSS: If your argument is that the new feature supersedes the old, deprecating it at the very least makes sense. If there's literally no reason to use this over the new methods, why keep it?

interactions/utils/abc/base_context_managers.py
sillyElvy: Made the abc after I made typing and didn't think about removing it

interactions/api/models/channel.py
sillyElvy: Because I didn't want to have this breaking

interactions/utils/abc/base_iterators.py
```diff
@@ -0,0 +1,96 @@
...
+class BaseIterator(metaclass=ABCMeta):
+    """A base class for iterators"""
+
+    @abstractmethod
+    def __init__(
+        self,
+        obj: Union[int, str, "Snowflake", _T] = None,
+        maximum: Optional[int] = inf,
+        start_at: Optional[Union[int, str, "Snowflake", _O]] = MISSING,
+        check: Optional[Callable[[_O], bool]] = None,
```
Catalyst4222: Can you remove the parameters? Same reason as the async iterator

interactions/utils/utils.py
```diff
@@ -231,3 +245,66 @@ def disable_components(
             for _components in components:
                 for component in _components.components:
                     component.disabled = True
+
+
+def get_channel_history(
+    http: "HTTPClient",
```
Catalyst4222: Can you make it so either the http client or the bot can be passed? Many users probably won't even know what the http client is, which is intended due to the underscore prefix

interactions/utils/utils.py
```diff
@@ -231,3 +245,66 @@ def disable_components(
...
+    :rtype: AsyncHistoryIterator
+    """
+    from ..api.models.channel import AsyncHistoryIterator
+
+    return AsyncHistoryIterator(
+        http, channel, start_at=start_at, reverse=reverse, check=check, maximum=maximum
+    )
+
+
+def get_guild_members(
```
Catalyst4222, line 289: See above comment about the httpclient and bot

Catalyst4222 (changes requested):

Catalyst4222 (approved):

PR #1061, opened 2022-09-02
fix: Change event sets, fix proper close() WS handling

FayeDel:
## About

This pull request fixes minor bugs relating to setting event locks and prevents another edge case of setting heartbeats when client isn't defined.

## Checklist

  - resolves #
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [X] Bugfix

### DIFF
--- a/interactions/api/gateway/client.py
+++ b/interactions/api/gateway/client.py
@@ -701,7 +701,7 @@ class WebSocketClient:
         Restarts the client's connection and heartbeat with the Gateway.
         """
 
-        self._ready.clear()
+        self.ready.clear()
 
         async with self.reconnect_lock:
             self.__closed.clear()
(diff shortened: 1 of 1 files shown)


interactions/api/gateway/client.py
```diff
@@ -831,9 +831,11 @@ async def _send_packet(self, data: Dict[str, Any]) -> None:
             # This is because the ratelimiter limits already accounts for this.
             await self._ratelimiter.block()
 
-        self._last_send = perf_counter()
-        log.debug(packet)
-        await self._client.send_str(packet)
+        if self._client is not None:  # this mitigates against another edge case.
```
sillyElvy, line 834: Uhhh isn't it bad that we just don't send it? Maybe just make it while and wait until the client exist? Idk?

FayeDel: The edge case here is that if the client doesn't exist whenever we reconnect, it would still send (during lib reconnection, client resets itself to None and there could be a very minor time frame where the heartbeat will still send in between heartbeat manager reset and client close. Also, iirc, the manager's supposed to reset on reconnect anyway. Of course, that could be optimised eventually so in terms of heartbeat, it still works/resets fine atm

(changing it to while wouldn't do anything)

(On current heartbeat testing, it still beats regularly, minus current gateway loop problems that seem to stem from didcord)

sillyElvy (approved):
Alright lgtm

PR #1062, opened 2022-09-02
feat: Add full support for prune endpoints

Nanrech:
## About

This pull request adds support for both `prune` endpoints, as well as their respective helper methods: `Guild.prune()` & `Guild.get_prune_count`.

## Checklist

  - resolves #
  - [ ] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [ ] Bugfix

### DIFF
--- a/interactions/api/models/guild.py
+++ b/interactions/api/models/guild.py
@@ -1909,7 +1909,7 @@ class Guild(ClientSerializerMixin, IDMixin):
         self,
         days: int = 7,
         compute_prune_count: bool = True,
-        include_roles: Optional[Union[List[Role], List[int]]] = MISSING,
+        include_roles: Optional[Union[List[Role], List[int], List[Snowflake], List[str]]] = MISSING,
     ) -> Optional[int]:
         """
         Begins a prune operation.
@@ -1924,7 +1924,7 @@ class Guild(ClientSerializerMixin, IDMixin):
             raise LibraryException(code=13)
 
         if include_roles is not MISSING:
-            _roles = [role.id if isinstance(role, Role) else role for role in include_roles]
+            _roles = [int(role.id) if isinstance(role, Role) else int(role) for role in include_roles]
         else:
             _roles = None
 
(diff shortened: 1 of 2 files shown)


interactions/api/models/guild.py
```diff
@@ -1905,6 +1905,65 @@ async def get_all_bans(self) -> List[Dict[str, User]]:
...
+    async def prune(
+        self,
+        days: int = 7,
+        compute_prune_count: bool = True,
+        include_roles: Optional[Union[List[Role], List[int]]] = MISSING,
+    ) -> Optional[int]:
+        """
+        Begins a prune operation.
+
+        :param days: Number of days to count. Defaults to 7.
```
sillyElvy: Give options (min max or smth)

interactions/api/models/guild.py
```diff
@@ -1905,6 +1905,65 @@ async def get_all_bans(self) -> List[Dict[str, User]]:
 
         return res
 
+    async def prune(
+        self,
+        days: int = 7,
+        compute_prune_count: bool = True,
+        include_roles: Optional[Union[List[Role], List[int]]] = MISSING,
```
sillyElvy: List snowflake, List str missing

sillyElvy (changes requested):
~~Also add a helper for getting prune count if it doesn't exist already~~ignore that, github didn't load properly

interactions/api/models/guild.py
Nanrech: Done

interactions/api/models/guild.py
Nanrech: Check last commit.

interactions/api/models/guild.py
sillyElvy: This is still missing from the helper methods. Most users will look at those and not the http methods

Nanrech: I completely forgot. You are 100% right.

Nanrech: Pushed!

sillyElvy (changes requested):
How is one's maximum 30 but the other one's 3?

sillyElvy (approved):
Lgtm ig

PR #1064, opened 2022-09-03
fix: add intents in example

sus2790:
## About

## Checklist

  - resolves #
  - [ ] Documentation
  - [ ] Breaking change
  - [ ] New feature/enhancement
  - [x] Bugfix

### DIFF
--- a/examples/bot.py
+++ b/examples/bot.py
@@ -10,7 +10,7 @@ import interactions
 # When you make a bot, we refer to it as the "client."
 # The client is the main object that interacts with the Gateway, what talks to Discord.
 # The client is also the main object that interacts with the API, what makes requests with Discord.
-client = interactions.Client("your bot token will go here.")
+client = interactions.Client("your bot token will go here.", intents=interactions.Intents.DEFAULT)
 
 
 # With our client established, let's have the library inform us when the client is ready.
--- a/examples/bot.py
+++ b/examples/bot.py
@@ -10,9 +10,10 @@ import interactions
 # When you make a bot, we refer to it as the "client."
 # The client is the main object that interacts with the Gateway, what talks to Discord.
 # The client is also the main object that interacts with the API, what makes requests with Discord.
-client = interactions.Client("your bot token will go here.", intents=interactions
-.Intents.DEFAULT | interactions
-.Intents.GUILD_MESSAGE_CONTENT)
+client = interactions.Client(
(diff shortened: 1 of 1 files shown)


Damego:
message content intent doesn't included in DEFAULT

sus2790:
TBR...

examples/bot.py
```diff
@@ -10,7 +10,10 @@
 # When you make a bot, we refer to it as the "client."
 # The client is the main object that interacts with the Gateway, what talks to Discord.
 # The client is also the main object that interacts with the API, what makes requests with Discord.
-client = interactions.Client("your bot token will go here.")
+client = interactions.Client(
+    "your bot token will go here.",
+    intents=interactions.Intents.DEFAULT | interactions.Intents.GUILD_MESSAGE_CONTENT,
```
i0bs, line 15: Can we make this just the default?

Would also add comments

i0bs (changes requested):

examples/bot.py
Damego: > Can we make this just the default?

Example bot has `on_message_create` event so we need message intent for this

i0bs: Would rather have an example just to show bare connection tbh, other example files should be considered

Catalyst4222: I agree, a separate example about intents should be added

sus2790: Wait...., why it's `on_ready` not `on_start`?

mAxYoLo01: > Wait...., why it's `on_ready` not `on_start`?

The gateway can reconnect multiple times during the bot's life, and `on_ready` is triggered for each reconnection. However, `on_start` is triggered only **once** when the bot has been started for the first time, so it's better to use it.

sus2790: Dead.

sus2790: is time to resolving conversation

i0bs: Thoughts haven't changed: people would rather have a separate example page for intents, which would mean gutting out the `on_message_create` event handler code in total here.

Catalyst4222:
Can we get an update on this? A guide on intent would be very appreciated, though having them in the first example could be confusing to users

EdVraz:
Closing in 3 days if there is no more activity.

PR #1067, opened 2022-09-03
refactor: Change type of parameters in scheduled event requests from Snowflake to int

Nanrech:
## About

This pull request refactors the scheduledEvent http methods to take in `int` objects instead of `Snowflake`.
There's also a fix for `get_scheduled_events`, so the method now works properly when the parameter `with_counts` is set to True.

## Checklist

  - resolves #
  - [ ] Documentation
  - [ ] Breaking change
  - [x] New feature/enhancement
  - [x] Bugfix

### DIFF
--- a/interactions/api/http/scheduledEvent.py
+++ b/interactions/api/http/scheduledEvent.py
@@ -47,7 +47,7 @@ class ScheduledEventRequest:
 
         :param guild_id: Guild ID snowflake.
         :param guild_scheduled_event_id: Guild Scheduled Event ID snowflake.
-        :param with_user_count: A boolean to include number of users subscribed to the associated event, if given.
+        :param with_user_count: Whether the number of users subscribed to the events is returned.
         :return: A dictionary containing the guild scheduled event object on success.
         """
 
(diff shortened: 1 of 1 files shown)


Toricane (changes requested):
Please modify the docstrings, mainly changing parameter descriptions and adding parameter types

Nanrech:
> Please modify the docstrings, mainly changing parameter descriptions and adding parameter types

Their current state is identical to other docstrings in the http directory, though. What changes do you suggest?

Toricane:
They still say they are snowflakes, and the `:type X: ...` is missing for them as well

Damego:
In other http files same thing

Nanrech:
> They still say they are snowflakes, and the `:type X: ...` is missing for them as well

That happens to be the same across ALL http files. If we really want that changed, we should do so in a separate PR addressing all those files.

Toricane:
Oh ok, thanks for clarifying

Toricane (approved):
LGTM!

PR #1081, opened 2022-09-10
feat: implement a debug setting for the client

sillyElvy:
## About

This change makes it easier to enable debug logging and gives debug logging.

## Checklist

  - [ ] For the documentation
  - [x] To add a new feature
  - [ ] To fix an existing bug
  - [ ] To resolve #ISSUENUMBER

This is:
  - [ ] A breaking change

### DIFF
--- a/interactions/client/bot.py
+++ b/interactions/client/bot.py
@@ -58,8 +58,8 @@ class Client:
     :type default_scope?: Optional[Union[int, Guild, List[int], List[Guild]]]
     :param disable_sync?: Controls whether synchronization in the user-facing API should be automatic or not.
     :type disable_sync?: Optional[bool]
-    :param debug?: Whether debug logging should be enabled
-    :type debug?: Optional[bool]
+    :param logging?: Set to ``True`` to enable debug logging or set to a log level to use a specific level
+    :type logging?: Optional[Union[bool, logging.DEBUG, logging.INFO, logging.NOTSET, logging.WARNING, logging.ERROR, logging.CRITICAL]]
 
     :ivar AbstractEventLoop _loop: The asynchronous event loop of the client.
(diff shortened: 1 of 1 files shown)


Toricane (commented):
Should there be a way to change the level to stuff other than debug?

sillyElvy:
> Should there be a way to change the level to stuff other than debug?

Idk, if it'd be wanted I can bring back the previous commit but modify it a bit

Toricane:
Maybe name it `logging` or something instead of `debug`, where you can input a logging level or bool which defaults to debug or off

sillyElvy:
> Maybe name it `logging` or something instead of `debug`, where you can input a logging level or bool which defaults to debug or off

Done

interactions/client/bot.py
```diff
@@ -102,6 +104,21 @@ def __init__(
...
+            # thx i0 for posting this on the retux Discord
+
+            if _logging is True:
+                _logging = logging.DEBUG
+
+            _format = (
+                "%(asctime)s [%(levelname)s] - .%(funcName)s(): %(message)s"
+                if _logging == logging.DEBUG
+                else "%(asctime)s [%(levelname)s] - %(message)s"
+            )
```
Toricane, line 118: Why is the format different? Is funcName given only in DEBUG level?

sillyElvy: I don't think so - but why would you want the func name if you are not debugging

Toricane: Maybe to know where it errored or warned? Idk lmk if it's really unnecessary and ill approve

sillyElvy: I think it's unnecessary

Toricane (approved):
LGTM!

PR #1096, opened 2022-09-24
fix: `ComponentContext.label` not updated to reflect component models

GeomKid:
## About

This pull request is about fixing the bug of label, which causes the following error:
```py
  File "C:\Users\Asus\Documents\GitHub\Python-Slash-cmd\newcogs\cstop.py", line 795, in choice0
    ctx.label
  File "C:\Users\Asus\Documents\GitHub\Python-Slash-cmd\.venv\lib\site-packages\interactions\client\context.py", line 680, in label
    for component in action_row["components"]:
TypeError: 'ActionRow' object is not subscriptable
```
Original Code:
```py
@extension_component("drjoin")
async def choice0(self, ctx: ComponentContext):
    ctx.label 
```
## Checklist

  - [x] For the documentation
  - [ ] To add a new feature
  - [x] To fix an existing bug
  - [ ] To resolve #ISSUENUMBER

This is:
  - [ ] A breaking change


interactions/client/context.py
```diff
@@ -738,7 +738,19 @@ def custom_id(self) -> Optional[str]:
 
     @property
     def label(self) -> Optional[str]:
+        """
+        Gives the label of button
```
Toricane:
```suggestion
        The label of the interacted component.
```

interactions/client/context.py
```diff
@@ -738,7 +738,19 @@ def custom_id(self) -> Optional[str]:
...
+        :rtype: Optional[str]
+        """
+        if not self.data.component_type == ComponentType.BUTTON:
+            return None
```
Toricane: Don't selects have labels too? Remove this

interactions/client/context.py
```diff
@@ -738,7 +738,19 @@ def custom_id(self) -> Optional[str]:
...
+        if self.message is None:
+            return None
```
Toricane:
```suggestion
            return
```

interactions/client/context.py
```diff
@@ -738,7 +738,19 @@ def custom_id(self) -> Optional[str]:
...
+        if self.message.components is None:
+            return None
```
Toricane:
```suggestion
            return
```

interactions/client/context.py
```diff
@@ -738,7 +738,19 @@ def custom_id(self) -> Optional[str]:
...
         for action_row in self.message.components:
-            for component in action_row["components"]:
-                if component["custom_id"] == self.custom_id and component["type"] == 2:
-                    return component.get("label")
+            for component in action_row.components:
+                if component.custom_id == self.custom_id:
+                    return component.label
+        else:
+            return None
```
Toricane: Don't need this as the function will return None when the loop ends

Toricane (changes requested):
A few changes

interactions/client/context.py
GeomKid: I checked selects don't have labels

Toricane: Oops i confused it with the custom_id

interactions/client/context.py
```diff
@@ -738,7 +738,17 @@ def custom_id(self) -> Optional[str]:
 
     @property
     def label(self) -> Optional[str]:
+        """
+        The label of the interacted component.
```
Toricane: One last suggestion
```suggestion
        The label of the interacted button.
```

Toricane (changes requested):
One last suggestion

Toricane (approved):
LGTM!

PR #1104, opened 2022-09-26
feat: Implement guild_join event

ItsRqtl:
## About

This pull request is about on_guild_create event, which save the guilds array from ready (start) event, then excluding them from dispatching (then remove them from the list so the event will dispatch if the bot rejoined).

## Checklist

  - [ ] For the documentation
  - [x] To add a new feature
  - [ ] To fix an existing bug
  - [ ] To resolve #ISSUENUMBER

This is:
  - [ ] A breaking change

### DIFF
--- a/interactions/api/gateway/client.py
+++ b/interactions/api/gateway/client.py
@@ -470,10 +470,13 @@ class WebSocketClient:
                     ids = self.__get_object_ids(obj, model)
 
                 if "_create" in name or "_add" in name:
-                    if id and str(id) in self.__start_guild:
-                        self.__start_guild.remove(str(id))
-                    else:
-                        self._dispatch.dispatch(f"on_{name}", obj)
+                    if name == "guild_create":
+                        if id and str(id) in self.__start_guild:
+                            self.__start_guild.remove(str(id))
+                        else:
+                            self._dispatch.dispatch("on_join", obj)
+
+                    self._dispatch.dispatch(f"on_{name}", obj)
 
                     if id:
                         _cache.merge(obj, id)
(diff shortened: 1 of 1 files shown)


Damego:
Will be better implement new event `guild_join` instead of limiting `guild_create`

ItsRqtl:
reverted guild_create, dispatching new guild_join instead

interactions/api/gateway/client.py
```diff
@@ -319,6 +319,7 @@ async def _handle_stream(self, stream: Dict[str, Any]):
             self.session_id = data["session_id"]
             self.resume_url = data["resume_gateway_url"]
             if not self.__started:
+                self.__start_guild = [i["id"] for i in data["guilds"]]
```
Damego: `READY` event can be dispatched multiply times during runtime and every time it will contain a list with guilds. SO after a second `READY` event you will get `guild_join` event dispatching for all your guilds. I suggest move this from if statement.
Also rename `__start_guild` to `__unavailable_guilds`

interactions/api/gateway/client.py
```diff
@@ -469,6 +470,12 @@ def _dispatch_event(self, event: str, data: dict) -> None:
                     ids = self.__get_object_ids(obj, model)
 
                 if "_create" in name or "_add" in name:
+                    if name == "guild_create":
+                        if id and str(id) in self.__start_guild:
+                            self.__start_guild.remove(str(id))
```
Damego: rename `__start_guild` to `__unavailable_guilds`

Damego (changes requested):

Damego (approved):

Catalyst4222 (approved):

PR #1112, opened 2022-10-06
docs: Huge Docs revamp

### LINKED ISSUE #1054: [REQUEST] Update function parameters
### Describe the feature.

There are quite a few places where the function parameters don't quite line up with what it expects. This is mostly referencing `*args: Tuple[...]` and `**kwargs: Dict`, which when accessed inside of a function or when calling it, give the wrong typehints. In many docstrings, not only does the `*args`/`**kwargs` discrepancy show up, but on some IDEs (pycharm, possibly more), parameters can fail to appear when accessing the formatted docstring (often caused by `*args` or `parameter?`)

### Code of Conduct

mAxYoLo01:
## About

This PR's goal is to massively revamp the docs to a more readable version.

In this PR is included:
- Rethink the general layout of the docs to make them more user-friendly, and highlight the important part
- Remove certain "private" files from being generated in docs (cache, error, gateway, ...) as they should not be public
- Hide private methods & functions from the docs
- Fix multiple docs generation warnings & errors
- Remove all `?` from docstrings
- Combining `:param:` and `:type:` into a single line in docstrings

## Checklist

  - [x] For the documentation
  - [ ] To add a new feature
  - [x] To fix an existing bug
  - [ ] To resolve #ISSUENUMBER

This is:
  - [ ] A breaking change


sillyElvy (changes requested):
target branch. Also why the removal of error docs?

mAxYoLo01:
Here is the new layout of the docs:
image
image

docs/api.http.rst
```diff
@@ -3,6 +3,10 @@
 HTTP
 ====
 
+.. warning::
```
sillyElvy:
```suggestion
.. important::
```
maybe?

mAxYoLo01:
> target branch. Also why the removal of error docs?

I feel like since this is such a massive update, it belongs in unstable. Not only docs have been changed. Also, the docs branch is quite behind.

The error page had nothing important in it, kind of like the cache one, nothing that a random person needs to see.

sillyElvy:
> > target branch. Also why the removal of error docs?
> 
> I feel like since this is such a massive update, it belongs in unstable. Not only docs have been changed. Also, the docs branch is quite behind.
> 
> The error page had nothing important in it, kind of like the cache one, nothing that a random person needs to see.

Alright, that is fair

and yeah I didn't see the code changes tbf

mAxYoLo01:
I'm going to have to regenerate all the locale docs too with all the modifications, should I do it here or another PR?

docs/events.rst
```diff
@@ -19,18 +19,18 @@ Generally, you can listen to an event like this:
 
     # possibility 1
     @bot.event
-    async def on_(...):
+    async def on_xxx(...):
```
Damego, line 22: Can you add this link https://discord.com/developers/docs/topics/gateway-events#receive-events

docs/events.rst
```diff
@@ -74,7 +74,7 @@ as long as you don't absolutely need it.
 
 
 Event: ``on_start``
-^^^^^^^^^^^^^^^^^^^
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Damego, line 77: Why you added more `^` symbols?

interactions/api/dispatch.py
```diff
@@ -27,12 +27,9 @@ def dispatch(self, __name: str, *args, **kwargs) -> None:
...
         Dispatches an event given out by the gateway.
 
-        :param __name: The name of the event to dispatch.
-        :type __name: str
-        :param *args: Multiple arguments of the coroutine.
-        :type *args: list[Any]
-        :param **kwargs: Keyword-only arguments of the coroutine.
-        :type **kwargs: dict
+        :param str __name: The name of the event to dispatch.
+        :param list[Any] \*args: Multiple arguments of the coroutine.
```
Damego: `*args` is an `Any` and `args` is a `Tuple[Any]`, not a list. Same thing for `kwargs`

docs/events.rst
mAxYoLo01: Idk, I think I didn't understand yet the "rule" that it doesn't matter how many characters there are, as long as it's more than the length of the title

docs/events.rst
mAxYoLo01: this link is already present in this page line 154

sillyElvy (changes requested):
Has http been removed from index and searching?

mAxYoLo01:
> Has http been removed from index and searching?

Yes, as these methods should not be used by developers except if they have very good reasons to, which in this case I would guess they would check the source code at this point.

sillyElvy:
> > Has http been removed from index and searching?
> 
> Yes, as these methods should not be used by developers except if they have very good reasons to, which in this case I would guess they would check the source code at this point.

Yeah I'm pro removing, sick of seeing "why is `interactions.api.http.guild.GuildRequest.get_member(id)` not working" xD

sillyElvy (approved):
I think this is fine and I don't want to make you a hell conflict when I make another refactor, or the other way around

PR #1125, opened 2022-10-17
fix: 429 error not shown

Toricane:
## About

This pull request shows the 429 error properly.

I've been seeing lots of `JSONDecodeError`s from mostly Replit users (there is a documented case of it happening with Heroku as well), which leads to a `TypeError`. The common "fix" given to them was just to do `kill 1` in the shell with no additional information as to what happened and why.

When I looked further into the code, I saw that when Cloudflare ratelimits you, it returns an HTML page instead of JSON, causing the `JSONDecodeError` (because the library tries to JSONify the response immediately without any checks), which is typical for Replit users. The suggestion of `kill 1` basically runs the code on a new server in Replit, sometimes with a new IP, so it temporarily works again.

Ideally, this error shouldn't occur at all because the library handles ratelimits awesomely, but when it *does* occur, it isn't handled properly.

This PR highlights that it is a 429 error instead of not handling the exception.

## Checklist

  - [ ] For the documentation
  - [ ] To add a new feature
  - [x] To fix an existing bug
  - [ ] To resolve #ISSUENUMBER

This is:
  - [ ] A breaking change

### DIFF
--- a/interactions/api/http/request.py
+++ b/interactions/api/http/request.py
@@ -174,7 +174,7 @@ class _Request:
                         # This "redundant" debug line is for debug use and tracing back the error codes.
 
                         raise LibraryException(message=message, code=code, severity=40, data=data)
-                    elif isinstance(data, dict) and code == 0 and message:
+                    elif isinstance(data, dict) and code in (0, 403) and message:
                         log.debug(f"RETURN {code}: {dumps(data, indent=4, sort_keys=True)}")
                         # This "redundant" debug line is for debug use and tracing back the error codes.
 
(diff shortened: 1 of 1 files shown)


interactions/api/http/request.py
```diff
@@ -146,10 +146,19 @@ async def request(self, route: Route, **kwargs) -> Optional[Any]:
                 async with self._session.request(
                     route.method, route.__api__ + route.path, **kwargs
                 ) as response:
+                    if response.content_type == "application/json":
+                        data = await response.json()
+                        __import__("pprint").pprint(data)
```
sillyElvy: ?

Toricane: I turned it into draft to debug it because there were attribute errors but then I had to sleep so I'll resolve them later today, maybe in 10 hours or so

sillyElvy: alright

Toricane:
@EdVraz ready for review

interactions/api/http/request.py
```diff
@@ -161,40 +171,40 @@ async def request(self, route: Route, **kwargs) -> Optional[Any]:
...
                             f"RETURN {response.status}: {dumps(data, indent=4, sort_keys=True)}"
                         )
                         # This "redundant" debug line is for debug use and tracing back the error codes.
 
-                        raise LibraryException(
-                            message=data["message"], code=data["code"], severity=40, data=data
-                        )
-                    elif isinstance(data, dict) and data.get("code") == 0 and data.get("message"):
+                        raise LibraryException(message=message, code=code, severity=40, data=data)
+                    elif isinstance(data, dict) and code in (0, 403) and message:
```
sillyElvy: Why is 403 here?

sillyElvy (changes requested):

interactions/api/http/request.py
Toricane: Isn't that also an error where the token is wrong?

sillyElvy: I think it's only 0 and 403 always has a proper message? I'm unsure, but it was only 0 before and I'm confused about the change

Toricane: Ok I'll remove

Catalyst4222: Can we get an update on this?

Toricane: Removed it, sorry for the delay

Toricane: Done, sorry for the delay

sillyElvy (approved):
Lgtm I think

PR #1134, opened 2022-10-26
fix: The issue of multiple client unable to load the commands from the same extension

GeomKid:
## About

This pull request is about fixing the issue of multiple client unable to load the commands from the same extension
```py
bot1 = Client(
    token=login["token1"],
)
bot2 = Client(
    token=login["token2"],
)
for filename in os.listdir("./cogs"):
    if filename.endswith(".py"):
        bot1.load(f"cogs.{filename[:-3]}")
        bot2.load(f"cogs.{filename[:-3]}")
```
Will load the cogs into both Clients However due to some unidentified reasons the commands `self.client` will always be bot2
## Checklist

  - [ ] For the documentation
  - [ ] To add a new feature
  - [x] To fix an existing bug
  - [ ] To resolve #ISSUENUMBER

This is:
  - [ ] A breaking change

### DIFF
--- a/interactions/client/bot.py
+++ b/interactions/client/bot.py
@@ -509,7 +509,7 @@ class Client:
             This is an internal method. Do not call it unless you know what you are doing!
         """
         for cmd in self._commands:
-            if cmd.resolved:
+            if cmd.coro.__qualname__ in [Cmd.__qualname__ for Cmd in self.__command_coroutines]:
                 continue
 
             cmd.listener = self._websocket._dispatch
@@ -555,7 +555,6 @@ class Client:
                     self._scopes.add(cmd.scope if isinstance(cmd.scope, int) else cmd.scope.id)
 
             self.event(coro, name=f"command_{cmd.name}")
-            cmd.resolved = True
 
     async def __sync(self) -> None:  # sourcery no-metrics
         """
--- a/interactions/client/bot.py
+++ b/interactions/client/bot.py
@@ -1386,6 +1386,7 @@ class Client:
         else:
             log.debug(f"Loaded extension {name}.")
             self._extensions[_name] = module
+            del sys.modules[name]
             return extension
 
     def remove(
--- a/interactions/client/bot.py
+++ b/interactions/client/bot.py
@@ -1424,7 +1424,6 @@ class Client:
                         self._loop.create_task(
                             _extension.teardown(remove_commands=remove_commands)
                         )  # made for Extension, usable by others
-            del sys.modules[_name]
 
         else:
             with contextlib.suppress(AttributeError):
--- a/interactions/client/bot.py
+++ b/interactions/client/bot.py
@@ -509,7 +509,7 @@ class Client:
             This is an internal method. Do not call it unless you know what you are doing!
         """
         for cmd in self._commands:
-            if cmd.coro.__qualname__ in [Cmd.__qualname__ for Cmd in self.__command_coroutines]:
+            if cmd.coro.__qualname__ in [_cmd.__qualname__ for _cmd in self.__command_coroutines]:
                 continue
 
             cmd.listener = self._websocket._dispatch
--- a/interactions/client/models/command.py
+++ b/interactions/client/models/command.py
@@ -860,8 +860,7 @@ class Command(DictSerializerMixin):
             if has_args:  # foo(ctx, ..., *args, ..., **kwargs) OR foo(ctx, *args, ...)
                 return await _coro(
                     ctx,
-                    # pos before *args
-                    *(kwargs[opt] for opt in par_opts if opt in kwargs),
+                    *(kwargs[opt] for opt in par_opts if opt in kwargs),  # pos before *args
                     *args,
                     *(
                         kwargs[opt]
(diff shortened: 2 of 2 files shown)


interactions/client/bot.py
```diff
@@ -509,7 +509,7 @@ def __resolve_commands(self) -> None:
             This is an internal method. Do not call it unless you know what you are doing!
         """
         for cmd in self._commands:
-            if cmd.resolved:
+            if cmd.coro.__qualname__ in [Cmd.__qualname__ for Cmd in self.__command_coroutines]:
```
sillyElvy: why are you doing this change? it feels unnessecary to do an extra list comprehension for... for what even?

interactions/client/models/command.py
```diff
@@ -862,7 +860,8 @@ async def __call(
             if has_args:  # foo(ctx, ..., *args, ..., **kwargs) OR foo(ctx, *args, ...)
                 return await _coro(
                     ctx,
-                    *(kwargs[opt] for opt in par_opts if opt in kwargs),  # pos before *args
+                    # pos before *args
+                    *(kwargs[opt] for opt in par_opts if opt in kwargs),
```
sillyElvy: ?

interactions/client/models/command.py
```diff
@@ -411,7 +411,6 @@ class Command(DictSerializerMixin):
     :ivar Dict[str, int] num_options: The dictionary of the number of options per subcommand.
     :ivar Dict[str, Union[Callable[..., Awaitable], str]] autocompletions: The dictionary of autocompletions for the command.
     :ivar Optional[str] recent_group: The name of the group most recently utilized.
-    :ivar bool resolved: Whether the command is synced. Defaults to ``False``.
```
sillyElvy, line 414: state a reason for the removal of this attribute

interactions/client/models/command.py
```diff
@@ -436,7 +435,6 @@ class Command(DictSerializerMixin):
     )
     recent_group: Optional[str] = field(default=None, init=False)
     error_callback: Optional[Callable[..., Awaitable]] = field(default=None, init=False)
-    resolved: bool = field(default=False, init=False)
```
sillyElvy, line 439: state a reason for the removal of this attribute

sillyElvy (changes requested):

interactions/client/bot.py
Catalyst4222: It changed commands from being resolved globally after loaded by one client to being resolved once per client

sillyElvy: Hm I see. still it would make more sense to have it before the loop wouldn't it?

interactions/client/models/command.py
Catalyst4222: When a command is created in an extension, only a single Command instance is made. Having a single attribute to show it the command is resolved prevents multiple clients from losing that command

interactions/client/bot.py
Catalyst4222: `self.__command_coroutines` changes inside of the loop, so not recomputing it adds in the possibility of edge cases. Regardless, it's not expensive to compute, and if we were that worried about speed we wouldn't be using python 😉

sillyElvy: LMAO fair ok

interactions/client/models/command.py
GeomKid: Oops looks like my IDE just reformatted that
I'll fix that

interactions/client/bot.py
```diff
@@ -1387,6 +1386,7 @@ def load(
         else:
             log.debug(f"Loaded extension {name}.")
             self._extensions[_name] = module
+            del sys.modules[name]
```
Damego, line 1389: This raises an error

```py
Task exception was never retrieved
future: <Task finished name='Task-24' coro=<ext() done, defined at D:\python_gh\discord_bots\ipy_bot\venv\lib\site-packages\interactions\client\models\command.py:930> exception=KeyError('exts.someext')>
  File "D:\python_gh\discord_bots\ipy_bot\venv\lib\site-packages\interactions\client\models\command.py", line 970, in wrapper
    raise e
  File "D:\python_gh\discord_bots\ipy_bot\venv\lib\site-packages\interactions\client\models\command.py", line 939, in wrapper
    return await coro(ctx, *args, **kwargs)
  File "D:\python_gh\discord_bots\ipy_bot\main.py", line 55, in ext
    bot.remove("exts.someext")
  File "D:\python_gh\discord_bots\ipy_bot\venv\lib\site-packages\interactions\client\bot.py", line 1434, in remove
    del sys.modules[_name]
KeyError: 'exts.someext'
```

Damego (changes requested):

interactions/client/bot.py
Damego:
```suggestion
            if cmd.coro.__qualname__ in [_cmd.__qualname__ for _cmd in self.__command_coroutines]:
```

Damego (commented):
one thing

Damego (approved):
Okay it works fine for me

sillyElvy (approved):

PR #1150, opened 2022-11-05
refactor!: remove all unwanted `__repr__` and `__str__`

### LINKED ISSUE #1140: [REQUEST] Remove or improve `__repr__` and `__str__` on models
### Describe the feature.

Both of the `__repr__` and `__str__` magic methods can be very useful when used correctly. However, there are several places where their inclusion causes confusion. For example, the initial issue in [this help thread](https://canary.discord.com/channels/789032594456576001/1036495364737474612/1036495364737474612) was caused by the repr of the guild object just being the name. The default repr given by `attrs` would be better than returning the name of the object, due to it being clearer what kind of data it is. Similarly, the `__str__` magic method in Member can add extra confusion without providing a needed use case.

### Code of Conduct

sillyElvy:
## About

This pull request removes the `__repr__` and `__str__` so they do not override the attrs repr

## Checklist

  - [ ] For the documentation
  - [ ] To add a new feature
  - [x] To fix an existing bug
  - [x] To resolve #1140

This is:
  - [ ] A breaking change

### DIFF
--- a/docs/migration.rst
+++ b/docs/migration.rst
@@ -232,3 +232,7 @@ your option name! Example:
 ~~~~~~~~~~~~~~~
 
 The ``Color`` class was refactored into an enum. You may now access its contents via ``interactions.Color.COLOR``.
+Due to `complaints <https://www.github.com/interactions-py/library/issues/1040>`_ on str and repr magic methods, we have
+decided to `remove all implementations of them in the model classes <https://www.github.com/interactions.py/library/pulls/1050>`_.
+This affects the way you can send emojis. You can no longer use ``str(emoji)`` or ``f"{emoji}"``.
+Instead you will have to use ``emoji.format`` or ``f"{emoji.format}"`` to transform an emoji into a send-able form.
--- a/interactions/api/models/channel.py
+++ b/interactions/api/models/channel.py
@@ -478,9 +478,6 @@ class Channel(ClientSerializerMixin, IDMixin):
                 if not self.recipients:
                     self.recipients = channel.recipients
 
-    def __repr__(self) -> str:
-        return self.name
-
     @property
     def typing(self) -> Union[Awaitable, ContextManager]:
         """
--- a/interactions/api/models/emoji.py
+++ b/interactions/api/models/emoji.py
@@ -12,7 +12,7 @@ if TYPE_CHECKING:
 __all__ = ("Emoji",)
 
 
-@define(repr=False)
+@define()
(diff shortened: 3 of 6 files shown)


Catalyst4222:
What about emojis and users? 
I think that the repr of the emoji is helpful, just in the wrong place. Could you make it a property instead?

sillyElvy:
> What about emojis and users? I think that the repr of the emoji is helpful, just in the wrong place. Could you make it a property instead?

I don't see how making it a property is helpful? It is a very good thing that when you do `"{emoji}"` the emoji gets immediately converted to it's sendable form and I don't see a reason to change that

Catalyst4222:
First, fstrings convert using `__str__` in that case, and emojis have `__repr__`. Second, it shouldn't be too much effort for the user to use an attribute/property. Your alternative (adding the `__str__` method) would slightly reduce typing for users using the fstring, but it can add confusing to devs printing out objects, which was the cause of the linked issue

sillyElvy:
> First, fstrings convert using `__str__` in that case, and emojis have `__repr__`. Second, it shouldn't be too much effort for the user to use an attribute/property. Your alternative (adding the `__str__` method) would slightly reduce typing for users using the fstring, but it can add confusing to devs printing out objects, which was the cause of the linked issue

I have changed it to `__str__`. I understand the confusion as I encountered it myself, too. However I think in case of the emoji object this confusion is at a lower level. I have enabled the attrs repr so if any developers are still highly confused about the emoji object printout they can `print(repr(emoji))`

Catalyst4222:
*Many* developers (including me) will print an object instead of the repr of the object, often because they're the same

What's the reason not to have it as a property?

sillyElvy:
> _Many_ developers (including me) will print an object instead of the repr of the object, often because they're the same
> 
> What's the reason not to have it as a property?

It is an unnecessary breaking change imo.

Also, in most cases you won't bother about the attributes of an emoji, would you? I'd assume mostly you just want to send it or create an reaction with it. I really thing having emojis like that is fine. If major complaints emerge, we still can change it at a later point.

Catalyst4222 (changes requested):
A complain already has emerged, which was the reason behind the issue in the first place. Having the magic method just adds unexpected cases, while a property will only impact code when it's accessed intentionally. A property makes it more clear what code is doing, which `__str__` does not

Catalyst4222 (approved):

PR #1151, opened 2022-11-06
feat!: rework how i.py handles the event loop

AstreaTSS:
## About

This pull request reworks how `interactions.py` handles the event loop to follow by the standards set by `asyncio.run()` - basically, instead of making and holding the event loop, it should instead let `asyncio.run()` make it, and get it with `asyncio.get_running_loop()`.

An attempt has been made to keep breaking changes minimal, but there are two exceptions:
- Running any method in `Client` before running `await client.astart()` or `client.start()` that relies on `_websocket` *will not work*, erroring out instead. That shouldn't be a huge problem, as `await client.wait_until_ready()` will work regardless.
- If an extension is loaded before the bot is started *and* relies on the event loop (IE creating a task to run an asynchronous function), it will most likely error out (unless `bot.load` is run in an asynchronous function). To avoid this, users are suggested to do the following:

```python
async def main():
    bot.load("ext_that_uses_loop")
    await bot.astart()

asyncio.run(main())
```

## Draft Notice

Nothing about this is final, and feedback is important. I've only done basic testing, so more needs to be done to find any bugs, and the solution to `client.wait_until_ready()` needs to be looked at too. Really, this PR was just to get the ball rolling.

## Checklist

  - [x] For the documentation
  - [x] To add a new feature
  - [ ] To fix an existing bug
  - [ ] To resolve #ISSUENUMBER

This is:
  - [x] A breaking change

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -50,11 +50,13 @@ What does that mean? Well, we'll show you:
 
 .. code-block:: python
 
+    import asyncio
     import interactions
+    import discord
     from discord.ext import commands
 
     client = interactions.Client(token="...")
-    dpy = commands.Bot(prefix="/")
+    dpy = commands.Bot(command_prefix="/", intents=...)
 
     @dpy.command()
     async def hello(ctx):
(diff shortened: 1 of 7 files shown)


AstreaTSS:
Closing this due to lack of activity. I'll open it again if we decide to re-approach this topic.

PR #1173, opened 2022-11-20
feat(channel): Add new overwrite helper methods

sillyElvy:
## About

Guys don't code when you're sick

## Checklist

- [ ] I've ensured the change(s) work on `3.8.6` and higher.

  - [ ] For the documentation
  - [x] To add a new feature
  - [ ] To fix an existing bug
  - [ ] To resolve #ISSUENUMBER

This is:
  - [ ] A breaking change

### DIFF
--- a/interactions/api/models/channel.py
+++ b/interactions/api/models/channel.py
@@ -1963,6 +1963,8 @@ class Channel(ClientSerializerMixin, IDMixin):
         reason: Optional[str] = None,
     ) -> "Channel":
         """
+        .. versionadded:: 4.4.0
+
         Removes the overwrite for the given ID.
 
         :param Union[int, str, Snowflake, User, Role, Overwrite] id: The ID of the User/Role to create the overwrite on.
--- a/interactions/api/models/channel.py
+++ b/interactions/api/models/channel.py
@@ -1898,6 +1898,8 @@ class Channel(ClientSerializerMixin, IDMixin):
         :param Optional[Literal[0, 1, "0", "1"]] type: The type of the overwrite. 0 for Role 1 for User.
         :param Optional[Union[int, Permissions, str]] allow: Permissions to allow
         :param Optional[Union[int, Permissions, str]] deny: Permissions to deny
+        :return: The updated channel
+        :rtype: Channel
         """
 
         if not deny and not allow:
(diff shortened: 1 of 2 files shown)


interactions/api/models/channel.py
```diff
@@ -1870,6 +1882,67 @@ async def get_permissions_for(self, member: "Member") -> Permissions:
...
+    async def add_permission_overwrites(self, overwrites: Iterable[Overwrite]) -> "Channel":
+        """
+        .. versionadded:: 4.4.0
+
+        Add multiple overwrites to the channel.
+
+        :param Iterable[Overwrite] overwrites: The overwrites to add to the channel.
+        """
+
+        ow = self.permission_overwrites or []
```
Damego: Rename ´ow´ to ´_overwrites´

Damego (commented):
Missed return and rtype in the docstrings

interactions/api/models/channel.py
```diff
@@ -1870,6 +1882,73 @@ async def get_permissions_for(self, member: "Member") -> Permissions:
 
         return Permissions(permissions)
 
+    async def add_permission_overwrite(
+        self,
+        id: Union[int, str, Snowflake, User, Role],
+        type: Optional[Literal[0, 1, "0", "1"]] = None,
+        allow: Optional[Union[int, Permissions, str]] = None,
+        deny: Optional[Union[int, Permissions, str]] = None,
```
Damego: Why here is None, not MISSING?

interactions/api/models/channel.py
```diff
@@ -1870,6 +1882,109 @@ async def get_permissions_for(self, member: "Member") -> Permissions:
...
+        :param Union[int, str, Snowflake, User, Role] id: The ID of the User/Role to create the overwrite on.
+        :param Optional[Literal[0, 1, "0", "1"]] type: The type of the overwrite. 0 for Role 1 for User.
+        :param Optional[Union[int, Permissions, str]] allow: Permissions to allow
+        :param Optional[Union[int, Permissions, str]] deny: Permissions to deny
+        :param Optional[str] reason: The reason to be shown in the audit log
+        :return: The updated channel
+        :rtype: Channel
+        """
+
+        if not deny and not allow:
```
Damego, line 1907:
```suggestion
        if deny is MISSING and allow is MISSING:
```

Damego (commented):
I think this better

interactions/api/models/channel.py
sillyElvy: that makes it worse

MISSING is not (`__bool__`). So it we check for missing only the user could just pass `None` for both

Damego: Okay then

Damego:
Does it work with new changes?

sillyElvy:
> Does it work with new changes?

on my machine yep

Damego (approved):

PR #1176, opened 2022-11-22
docs: Add missed docstring to context; Hide _Context

Damego:
## About

## Checklist

- [ ] I've ensured the change(s) work on `3.8.6` and higher.

  - [x] For the documentation
  - [ ] To add a new feature
  - [ ] To fix an existing bug
  - [ ] To resolve #ISSUENUMBER

This is:
  - [ ] A breaking change


interactions/client/context.py
```diff
@@ -38,12 +38,6 @@ class _Context(ClientSerializerMixin):
     that the user-facing API is able to allow developers to
     easily access information presented from any event in
     a "contextualized" sense.
-
-    :ivar Optional[Message] message: The message data model.
-    :ivar Member author: The member data model.
-    :ivar User user: The user data model.
-    :ivar Optional[Channel] channel: The channel data model.
-    :ivar Optional[Guild] guild: The guild data model.
```
sillyElvy, line 46: keep it

sillyElvy (changes requested):
see cmment

interactions/client/context.py
```diff
@@ -600,10 +605,27 @@ class ComponentContext(_Context):
     """
     A derivation of :class:`_Context`
```
sillyElvy: remove link aso for commandctx but cant mention it

sillyElvy (changes requested):

interactions/client/context.py
Damego: I added these ivars to Command & Component Contexts.
Also _Context is hidden now so why we need to keep it?

sillyElvy: i mean users can see it but if you think its not needed leave it out

sillyElvy (approved):

PR #1185, opened 2022-12-01
feat: Paginate guilds when syncing

mAxYoLo01:
## About

image
Previously, the library did not account for bots which were present in more than 200 guilds BUT used guild commands. Now, the function when syncing uses a paginator to make sure all guilds are acquired.

## Checklist

  - [ ] For the documentation
  - [x] To add a new feature
  - [x] To fix an existing bug
  - [ ] To resolve #ISSUENUMBER

This is:
  - [ ] A breaking change

### DIFF
--- a/interactions/api/models/guild.py
+++ b/interactions/api/models/guild.py
@@ -49,6 +49,7 @@ from .webhook import Webhook
 
 if TYPE_CHECKING:
     from ..http.client import HTTPClient
+    from ...client.bot import Client
     from .gw import AutoModerationRule, VoiceState
     from .message import Message
 
--- a/interactions/api/models/guild.py
+++ b/interactions/api/models/guild.py
@@ -3036,6 +3036,37 @@ class Guild(ClientSerializerMixin, IDMixin):
             else None
         )
 
+    @classmethod
+    async def get_all_guilds(cls, http: "HTTPClient") -> List["Guild"]:
+        """
+        .. versionadded:: 4.4.0
+
+        Gets all guilds that the bot is present in.
+
+        :return: List of guilds
+        :rtype: List[Guild]
+        """
+
+        _after = None
+        _all: list = []
+
+        res: list = await http.get_self_guilds(limit=200)
+
+        while len(res) >= 1000:
+
+            for guild in res:
+                _all.append(Guild(**guild))
+            _after = int(res[-1]["id"])
+
+            res = await http.get_self_guilds(
+                after=_after,
+            )
+
+        for guild in res:
+            _all.append(Guild(**guild))
+
+        return _all
+
 
 @define()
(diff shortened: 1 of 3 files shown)


interactions/api/models/guild.py
```diff
@@ -3036,6 +3036,37 @@ def discovery_splash_url(self) -> Optional[str]:
             else None
         )
 
+    @classmethod
+    async def get_all_guilds(cls, http: "HTTPClient") -> List["Guild"]:
```
sillyElvy: Maybe change it/set Union to bot? That might be an option to hide http (your idea)

mAxYoLo01: Good idea, I'd rather expose bot than http!

AstreaTSS (commented):
Going through every single guild is a *very* slow operation. Perhaps a warning would be best? Maybe advise the user to not do this if more than one iteration is needed?

Also, while this doesn't relate to this PR, maybe adding in a toggle to disable guild-based syncing would be a good feature too, as many only need to update global commands.

interactions/api/models/guild.py
```diff
@@ -3036,6 +3037,38 @@ def discovery_splash_url(self) -> Optional[str]:
...
+        :return: List of guilds
+        :rtype: List[Guild]
+        """
+
+        _after = None
+        _all: list = []
+
+        res: list = await client._http.get_self_guilds(limit=200)
+
+        while len(res) >= 1000:
```
AstreaTSS: Just a question: why 1000?

sillyElvy: Copy paste from get all bans I guess

mAxYoLo01: you're right, wrong copy paste oops, fixed!

mAxYoLo01:
Ok, I have modified the PR to make the function private to Client, as we already have `bot.guilds` and it would act as a duplicate. Also allows to reduce the time taken as it's not converted to a Guild object

interactions/client/bot.py
```diff
@@ -507,13 +507,39 @@ async def wait_until_ready(self) -> None:
         """
         await self._websocket.wait_until_ready()
 
+    async def __get_all_guilds(self) -> List[dict]:
```
Catalyst4222: Is there a reason why this method needs two prefixed underscores? It's not something that could mess with the bot is mismanaged, so at most it should only have one underscore

Catalyst4222 (changes requested):

interactions/client/bot.py
mAxYoLo01: good point, fixed

Catalyst4222 (approved):

mAxYoLo01:
PR was tested and it resolves the issue that one person had

PR #1196, opened 2022-12-11
fix: Pass HTTPClient to all objects when using get() with cache.

mAxYoLo01:
## About

This bug was found on [the support server](https://discord.com/channels/789032594456576001/1051526179846373457).
Basically, some objects do not have the HTTPClient passed through when getting the objects from the cache. This PR makes sure that, if the object is found in cache, it has the HTTPClient added to it.

PS: The `get()` function is a pain to debug, I would like to suggest to rewrite it in a way that it will be less hard for the next contributor to fix it.

## Checklist

  - [ ] For the documentation
  - [ ] To add a new feature
  - [x] To fix an existing bug
  - [ ] To resolve #ISSUENUMBER

This is:
  - [ ] A breaking change

### DIFF
--- a/interactions/utils/get.py
+++ b/interactions/utils/get.py
@@ -192,7 +192,7 @@ def get(client: "Client", obj: Type[_T], **kwargs) -> Optional[_T]:
         http_name = f"get_{_obj.__name__.lower()}"
         kwarg_name = f"{_obj.__name__.lower()}_ids"
         _objects: List[Union[_obj, Coroutine]] = (
-            _get_cache(_obj, client, kwarg_name, _list=True, **kwargs) if force_http else []
+            [] if force_http else _get_cache(_obj, client, kwarg_name, _list=True, **kwargs)
         )  # some sourcery stuff i dunno
 
         if force_cache:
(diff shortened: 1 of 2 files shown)


interactions/utils/get.py
```diff
@@ -192,7 +192,7 @@ def get(client: "Client", obj: Type[_T], **kwargs) -> Optional[_T]:
         http_name = f"get_{_obj.__name__.lower()}"
         kwarg_name = f"{_obj.__name__.lower()}_ids"
         _objects: List[Union[_obj, Coroutine]] = (
-            _get_cache(_obj, client, kwarg_name, _list=True, **kwargs) if force_http else []
+            [] if force_http else _get_cache(_obj, client, kwarg_name, _list=True, **kwargs)
```
Damego:
```suggestion
            _get_cache(_obj, client, kwarg_name, _list=True, **kwargs) if not force_http else []
```

interactions/utils/get.py
```diff
@@ -287,13 +287,19 @@ def _get_cache(
                 )
                 for _id in kwargs.get("member_ids")
             ]
-            _obj.extend(client._http.cache[_object].get(item, None) for item in _values)
+            for item in _values:
+                _obj = client._http.cache[_object].get(item, None)
+                if isinstance(_obj, _object):
```
Damego: I think checking for None will be enough

Damego (approved):

PR #1200, opened 2022-12-14
fix: Remove unwanted arguments passed to HTTP methods

Nanrech:
## About

EDIT: This PR's purpose is now refactoring all edit/modify helper methods so that only the provided arguments are sent in the http payload

## Checklist

  - [ ] For the documentation
  - [ ] To add a new feature
  - [x] To fix an existing bug
  - [ ] To resolve #ISSUENUMBER

This is:
  - [ ] A breaking change
  ^ Not sure

### DIFF
--- a/interactions/api/models/channel.py
+++ b/interactions/api/models/channel.py
@@ -846,10 +846,7 @@ class Channel(ClientSerializerMixin, IDMixin):
             payload["auto_archive_duration"] = auto_archive_duration
         if locked is not MISSING:
             payload["locked"] = locked
-        if position is MISSING:
-            if self.position is not None:
-                payload["position"] = self.position
-        else:
+        if position is not MISSING:
             payload["position"] = position
 
         res = await self._client.modify_channel(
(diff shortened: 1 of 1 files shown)


interactions/api/models/channel.py
```diff
@@ -848,6 +846,11 @@ async def modify(
             payload["auto_archive_duration"] = auto_archive_duration
         if locked is not MISSING:
             payload["locked"] = locked
+        if position is MISSING:
+            if self.position is not None:
+                payload["position"] = self.position
```
mAxYoLo01:
```suggestion
            if self.position is not None:
                payload["position"] = self.position
```
From what I see, the payload expects to get only the parameters which **need** to be modified. Therefore, why pass the current position, if we're not interested in changing it? Same logic appears for all parameters btw.

Nanrech: That's a really good point. Committed out the redundant part.

mAxYoLo01: Can you do the same for all non-necessary parameters?

Damego:
Closed due to inactive

PR #1230, opened 2023-01-20
feat: Add command enhancements

i0bs:
This proposal targets the `5.x` branch and allows commands to infer from the coroutine's name instead of requiring kwargs usage of the decorator.

Additionally, I added a `.respond()` alias to `.send()` for sending.

### DIFF
--- a/interactions/client/mixins/send.py
+++ b/interactions/client/mixins/send.py
@@ -95,3 +95,48 @@ class SendMixin:
             if delete_after:
                 await message.delete(delay=delete_after)
             return message
+        
+    async def respond(
+        self,
+        content: Optional[str] = None,
+        *,
+        embeds: Optional[Union[Iterable[Union["Embed", dict]], Union["Embed", dict]]] = None,
+        embed: Optional[Union["Embed", dict]] = None,
+        components: Optional[
+            Union[
+                Iterable[Iterable[Union["BaseComponent", dict]]],
+                Iterable[Union["BaseComponent", dict]],
+                "BaseComponent",
+                dict,
+            ]
+        ] = None,
+        stickers: Optional[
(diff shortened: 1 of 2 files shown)


interactions/models/internal/application_commands.py
```diff
@@ -839,7 +841,7 @@ def wrapper(func) -> SlashCommand:
 
 
 def context_menu(
-    name: str | LocalisedName,
+    name: Absent[str | LocalisedName] = MISSING,
     context_type: "CommandTypes",
```
LordOfPolls: Non-default parameter follows a default parameter

LordOfPolls (changes requested):
Context Menus don't appear to work with this (even after fixing the default param issue): 
```python
@context_menu(context_type=CommandTypes.USER)
async def user_menu(ctx):
    ...
```

image

interactions/models/internal/application_commands.py
```diff
@@ -839,7 +841,7 @@ def wrapper(func) -> SlashCommand:
 
 
 def context_menu(
-    name: str | LocalisedName,
+    name: Absent[str | LocalisedName] = MISSING,
```
i0bs:
```suggestion
    name: str | LocalisedName,
```

LordOfPolls:
Why have you reverted support for the nameless definition on context menus?

interactions/client/mixins/send.py
```diff
@@ -95,3 +95,48 @@ async def send(
             if delete_after:
                 await message.delete(delay=delete_after)
             return message
+
+    async def respond(
```
silasary, line 99: Any reason we can't just do `respond = send`?

silasary (changes requested):

LordOfPolls:
Closing due to inactivity. Feel free to reopen but bare in mind:
- respond aliasing has now been integrated upstream 
- this branch is **very** out-of-sync

PR #1243, opened 2023-01-23
fix: `client`, `command` and `extension` are missed in `on_command`

### LINKED ISSUE #1133: [BUG] Context is missing `command` attribute in the `on_component` event
### Describe the bug.

Taken from [`#AttributeError 'NoneType' object has no attribute 'name'`](https://canary.discord.com/channels/789032594456576001/1033886778165305424/1033886778165305424) in the help forum

In the `on_command` event, there is a possibility that `ctx.command` is None instead of the associated Command instance. This is due to a race condition where some attribute processing for Context is done when the command is being invoked. This issue likely applies to other attributes as well, though that has not been tested yet

A workaround is to call `asyncio.sleep` with a short delay, but users likely won't find that solution on their own

### List the steps.

Set up the client with a command and a listener for the `on_command` event. In the event, access `ctx.command`

### What you expected.

`ctx.command` to be a Command instance 

### What you saw.

`ctx.command` is None

release

### Version specification

4.3.2

### Code of Conduct

Damego:
## About

This pull request moves `on_command` event dispatching to other place where command's func is called. Resolves #1133 

## Checklist

  - [ ] For the documentation
  - [ ] To add a new feature
  - [x] To fix an existing bug
  - [x] To resolve #1133 

This is:
  - [ ] A breaking change

### DIFF
--- a/interactions/client/models/command.py
+++ b/interactions/client/models/command.py
@@ -897,6 +897,8 @@ class Command(DictSerializerMixin):
             ctx.command = self
             ctx.extension = self.extension
 
+            self.listener.dispatch("on_command", ctx)
+
             try:
                 if self.extension:
                     return await coro(self.extension, ctx, *args, **kwargs)
--- a/interactions/api/gateway/client.py
+++ b/interactions/api/gateway/client.py
@@ -404,7 +404,6 @@ class WebSocketClient:
                     _option = self.__sub_command_context(option, _context)
                     __kwargs.update(_option)
 
-            self._dispatch.dispatch("on_command", _context)
         elif data["type"] == InteractionType.MESSAGE_COMPONENT:
             _name = f"component_{_context.data.custom_id}"
 


interactions/client/models/command.py
```diff
@@ -897,6 +897,8 @@ async def wrapper(ctx: "CommandContext", *args, **kwargs):
             ctx.command = self
             ctx.extension = self.extension
 
+            self.listener.dispatch("on_command", ctx)
```
Toricane, line 900: Is it better to dispatch `on_command` before the command is executed or after?

Damego: I put dispatch before calling a command function. Can't after because we are raising an error

i0bs: @Toricane We should probably be dispatching beforehand anyhow, so that developers can have that data exposed to them before a potential error is thrown

Damego: Isn't what I did?

i0bs (approved):

PR #1248, opened 2023-01-25
refactor: run black dwarf on lib

LordOfPolls:
## About

This PR runs my toy project, BlackDwarf, on the lib to help eliminate wildcard imports

## Checklist

  - [ ] For the documentation
  - [ ] To add a new feature
  - [ ] To fix an existing bug
  - [ ] To resolve #ISSUENUMBER

This is:
  - [ ] A breaking change


interactions/__init__.py
```diff
@@ -1,7 +1,360 @@
-from .client import *
-from .client.const import *
-from .models import *
-from .api import *
+from .client import (
+    __version__,
+    __repo_url__,
+    __py_version__,
+    __api_version__,
+    get_logger,
```
i0bs: Can "Black dwarf" filter out some of the builtin imports? I don't want to access `logging` functions in the root namespace

interactions/__init__.py
```diff
@@ -1,7 +1,360 @@
...
+    GLOBAL_SCOPE,
+    MISSING,
+    MENTION_PREFIX,
+    PREMIUM_GUILD_LIMITS,
+    Absent,
+    T,
+    T_co,
+)
+from .models import (
+    misc,
```
i0bs: I'm not sure why we're importing other namespaces when we expose everything from them underneath

interactions/__init__.py
LordOfPolls: I'm a touch confused; None of these are buitlins? Regarding `get_logging`, this is mirroring the previous behavior of which it was available at the root namespace 

For context, all it's doing is extracting the `__all__` from the imported package. If one doesn't exist it will infer what should be imported, but in that case, it's importing anything that isnt private.

interactions/__init__.py
LordOfPolls: That looks like an error with inferencing - I'll see if I can filter that out programmatically

PR #1268, opened 2023-02-03
feat: add component to property of ComponentContext

GeomKid:
## About

## Checklist

- [x] I have added the `versionadded`, `versionchanged` and `deprecated` to any new or changed user-facing function I committed.

  - [x] For the documentation
  - [x] To add a new feature
  - [ ] To fix an existing bug
  - [ ] To resolve #ISSUENUMBER

This is:
  - [ ] A breaking change


Damego (commented):
`self.message` could be None. Add a check for this

interactions/client/context.py
```diff
@@ -846,3 +846,21 @@ def label(self) -> Optional[str]:
...
+        .. versionadded:: 4.4.0
+
+        The component that you interacted.
+
+        :rtype: Optional[Union[Button, SelectMenu]]
+        """
+        if self.message is None:
+            return
+        if self.message.components is None:
+            return
```
Damego:
```suggestion
        if self.message is None or self.message.components is None:
            return

```

Damego (approved):

PR #1274, opened 2023-02-08
feat💥: make subcmds able to inherit checks from base cmds

AstreaTSS:
## About

This pull request allows (and enables by default) for `SlashCommand`s and `PrefixedCommand`s to inherit checks from their base command definition, if they exist. Furthermore, this also adds an `add_check` function to `BaseCommand` to make it easier to add checks to base commands.

For example,
```python
async def a_check(ctx: interactions.InteractionContext):
    await ctx.send("base command test")
    return True

cmd = interactions.SlashCommand(name="test", description="test")
cmd.add_check(a_check)

@cmd.subcommand(sub_cmd_name="sub", sub_cmd_description="desc", inherit_checks=True)
async def sub(ctx: interactions.InteractionContext):
    await ctx.send("sub command")
```

will produce something like this:
A demonstation of inherit checks working.

💥 This also replaces `hierarchical_checking` in `PrefixedCommand`s with this. These two do pretty similar things, and I figured it was better to have consistency between the slash and prefixed commands. As an upside, it also makes prefixed command logic simpler.

## Checklist

  - [ ] For the documentation
  - [x] To add a new feature
  - [ ] To fix an existing bug
  - [ ] To resolve #ISSUENUMBER

This is:
  - [x] A breaking change

### DIFF
--- a/interactions/ext/prefixed_commands/command.py
+++ b/interactions/ext/prefixed_commands/command.py
@@ -346,13 +346,6 @@ class PrefixedCommand(BaseCommand):
         ),
         default=True,
     )
-    hierarchical_checking: bool = attrs.field(
-        metadata=docs(
-            "If `True` and if the base of a subcommand, every subcommand underneath it will run this command's checks"
-            " and cooldowns before its own. Otherwise, only the subcommand's checks are checked."
-        ),
-        default=True,
-    )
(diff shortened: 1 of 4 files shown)


LordOfPolls (approved):
lgtm

PR #1275, opened 2023-02-08
feat💥: rework slash command parameter processing

AstreaTSS:
## About

This pull request replaces the expensive `BaseCommand` processing (which was only used for slash commands anyways) with a dummy `call_callback`, and makes a new `call_callback` in `SlashCommand` that is much more efficient than the old one, using pre-processing to speed things up.

Most *normal* commands will function as usual. However, if users did anything strange (or relied on `Annotated` to not properly work for `SlashCommand` for some reason), there is a *very* high chance it broke. Also, `CMD_*` arguments are gone due to their pointlessness.

While I did test it out myself, I can't say that it's 100% bulletproof. I would appreciate if other people could test it out.

(I also had to fix autocompletes for this since not doing so would have made it impossible to test--)

## Checklist

  - [ ] For the documentation
  - [x] To add a new feature
  - [x] To fix an existing bug
  - [ ] To resolve #ISSUENUMBER

This is:
  - [x] A breaking change


interactions/client/client.py
```diff
@@ -1597,12 +1600,12 @@ async def _dispatch_interaction(self, event: RawGatewayEvent) -> None:
                     elif autocomplete := self._global_autocompletes.get(str(auto_opt.name)):
                         callback = autocomplete
                     else:
-                        breakpoint()
+                        # breakpoint()
```
LordOfPolls: You can safely remove this, it was removed in #1273

interactions/client/client.py
```diff
@@ -1597,12 +1600,12 @@ async def _dispatch_interaction(self, event: RawGatewayEvent) -> None:
...
                         raise ValueError(f"Autocomplete callback for {str(auto_opt.name)} not found")
 
                     await self.__dispatch_interaction(
                         ctx=ctx,
-                        callback=callback(ctx),
+                        callback=callback(ctx, **ctx.kwargs),
```
LordOfPolls: I would argue passing kwargs to autocompletes is both unexpected for most use cases, and unnecessary; namely because in the unlikely event someone actually needs them, they are available through `ctx.kwargs`
Adding this only adds bloat and a barrier of entry

LordOfPolls (changes requested):

interactions/client/client.py
AstreaTSS: Really? Because this example doesn't work without it: 
https://github.com/interactions-py/interactions.py/blob/5.x/docs/src/Guides/03%20Creating%20Commands.md#i-need-more-than-25-choices

AstreaTSS: Regardless, reverted. This can be addressed in another PR.

PR #1280, opened 2023-02-14
fix(context): `edit` method doesn't properly works after `send`

GeomKid:
## About

This pull request is about Fixing edit after slash cmd
```py
@testbot1.command(name="test3", scope=scope, description="Super Test")
async def test3(ctx: CommandContext):
    await ctx.defer(ephemeral=True)
    await ctx.send("Test Start")
    # await testbot1._logout()
    await asyncio.sleep(10)
    await ctx.edit(
        "Test Done", components=Button(custom_id="Baka", label="Boo", style=ButtonStyle.PRIMARY)
    )
```
## Checklist

  - [ ] For the documentation
  - [ ] To add a new feature
  - [ ] To fix an existing bug
  - [ ] To resolve #ISSUENUMBER

This is:
  - [ ] A breaking change

### DIFF
--- a/interactions/client/context.py
+++ b/interactions/client/context.py
@@ -555,17 +555,13 @@ class CommandContext(_Context):
         msg = None
 
         if self.deferred or self.responded:
-            try:
-                res = await self._client.edit_interaction_response(
-                    data=payload,
-                    files=files,
-                    token=self.token,
-                    application_id=str(self.application_id),
-                )
-            except LibraryException as e:
-                raise e from e
-            else:
-                self.message = msg = Message(**res, _client=self._client)
+            res = await self._client.edit_interaction_response(
+                data=payload,
+                files=files,
+                token=self.token,
+                application_id=str(self.application_id),
+            )
+            self.message = msg = Message(**res, _client=self._client)
         else:
             self.callback = InteractionCallbackType.UPDATE_MESSAGE
             await self._client.create_interaction_response(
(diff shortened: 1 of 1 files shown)


interactions/client/context.py
```diff
@@ -554,43 +554,22 @@ async def edit(
...
+            try:
+                res = await self._client.edit_interaction_response(
+                    data=payload,
+                    files=files,
+                    token=self.token,
+                    application_id=str(self.application_id),
+                )
+            except LibraryException as e:
+                if e.code in {10015, 10018}:
+                    log.warning(f"You can't edit hidden messages." f"({e.message}).")
```
Damego: I think we don't need this anymore

interactions/client/context.py
```diff
@@ -554,43 +554,18 @@ async def edit(
...
+        if self.deferred or self.responded:
+            try:
+                res = await self._client.edit_interaction_response(
+                    data=payload,
+                    files=files,
+                    token=self.token,
+                    application_id=str(self.application_id),
+                )
+            except LibraryException as e:
+                raise e from e
```
Damego: And also we don't need try except block

Damego (approved):
I think it's fine

Damego:
Add more information in your pr (code for example) why `edit` doesn't work

Damego:
I'm gonna merge it

PR #1288, opened 2023-03-01
Voice rcv

LordOfPolls:
## About

Supports basic voice receiving and recording. From my testing this seems to work. 

This is heavily based on the prototype dis-snek/NAFF code contributed to by @silasary and @zevaryx, the only difference is this one should actually work 

## Checklist

  - [ ] For the documentation
  - [x] To add a new feature
  - [ ] To fix an existing bug
  - [ ] To resolve #ISSUENUMBER

This is:
  - [ ] A breaking change


LordOfPolls:
Example Usage:

```python
@slash_command("record", description="Record audio in your voice channel")
@slash_option("duration", "The duration of the recording", opt_type=interactions.OptionType.NUMBER, required=True)
async def record(ctx: interactions.SlashContext, duration: int):
    if ctx.author.voice.channel is None:
        return await ctx.send("You must be in a voice channel to use this command.")
    voice_channel = ctx.author.voice.channel
    voice_state = await voice_channel.connect()

    recorder = voice_state.start_recording()
    await ctx.send(f"Recording for {duration} seconds")
    await asyncio.sleep(duration)
    voice_state.stop_recording()

    await ctx.send(
        "Here is your recording", files=[File(f, file_name=f"{user_id}.mp3") for user_id, f in recorder.output.items()]
    )
```

LordOfPolls:
This is now ready for review

interactions/api/voice/audio.py
```diff
@@ -5,19 +5,70 @@
...
+                header_ext_length = int.from_bytes(decrypted[2:4], byteorder="big")
+                decrypted = decrypted[4 + 4 * header_ext_length :]
+            self.decoded = self._recorder.get_decoder(self.ssrc).decode(decrypted)
+            return self.decoded
+
+    @property
+    def user_id(self) -> Optional[int]:
+        """The ID of the user who made this audio."""
+        while not self._recorder.state.ws.user_ssrc_map.get(self.ssrc):
+            time.sleep(0.05)
```
Toricane, line 69: Isn't this blocking?

LordOfPolls: Yes, if it were to be called from an async function, which it never is. RawAudio is only ever used from within a thread

Toricane: Oh ok

PR #1307, opened 2023-03-21
docs: add naff migration guide

LordOfPolls:
## About

Adds a NAFF migration guide


docs/src/Guides/99 2.x Migration_NAFF.md
```diff
@@ -0,0 +1,51 @@
+Oh hey! So you're migrating from NAFF to interactions.py? Well lets get you sorted.
+
+First and foremost, you'll need to install the new library. You can do this by running `pip install interactions.py` in your terminal.
+Then, the first thing you'll need to do is change your imports. You'll need to change `from naff import _` to `from interactions import _`. To be honest, assuming your code is relatively simple, you should be able to use find and replace to do this.
+
+## Prefixed Commands
+I.py moves prefixed commands to an extension, rather than being a part of the client. So to use them you'll nneed to load them.
```
VincentRPS:
```suggestion
I.py moves prefixed commands to an extension, rather than being a part of the client. So to use them you'll need to load them.
```

Damego: pog

PR #1340, opened 2023-04-11
feat: improve slash_option decorator

Toricane:
## Pull Request Type

- [x] Feature addition
- [ ] Bugfix
- [x] Documentation update
- [x] Code refactor
- [ ] Tests improvement

## Description

Improves the slash_option decorator so that the name, option type, and required arguments are inferred based on the parameter.
Example before and after:
```py
@slash_command(name="my_command", ...)
@slash_option(
    name="integer_option",
    description="Integer Option",
    required=True,
    opt_type=OptionType.INTEGER
)
async def my_command_function(ctx: InteractionContext, integer_option: int):
    await ctx.send(f"You input {integer_option}")
```
```py
@slash_command(name="my_command", ...)
@slash_option("Integer Option")
async def my_command_function(ctx: InteractionContext, integer_option: int):
    await ctx.send(f"You input {integer_option}")
```

## Changes

- The description of the option is required and can be passed to the decorator.
- The name of the option defaults to the name of the parameter.
- The type of the option can be specified as a type annotation to the parameter.
- Whether the option is required is inferred from the parameter's default value. If there is a default value specified, the option is not required.

## Test Scenarios

Use the examples in the guide in the documentation.

## Python Compatibility

## Checklist


interactions/models/internal/application_commands.py
```diff
@@ -1175,10 +1175,10 @@ def wrapper(func: AsyncCallable) -> ModalCommand:
 
 
 def slash_option(
-    name: str,
     description: str,
```
i0bs:
```suggestion
    description: Absent[str] = MISSING,
```

interactions/models/internal/application_commands.py
```diff
@@ -1208,11 +1208,29 @@ def wrapper(func: SlashCommandT) -> SlashCommandT:
...
 
+        if not hasattr(func, "options"):
+            func.options = []
+
+        parameters = list(inspect.signature(func).parameters.values())
+        param = parameters[-1 - len(func.options)]
+        option_name = param.name if name is MISSING else name
+        option_type = param.annotation if opt_type is MISSING else opt_type
+        option_required = param.default is param.empty if required is MISSING else required
+
```
i0bs, line 1220:
```suggestion
        option_description = "No description" if description is MISSING else description

```

interactions/models/internal/application_commands.py
```diff
@@ -1208,11 +1208,29 @@ def wrapper(func: SlashCommandT) -> SlashCommandT:
...
+        if not isinstance(option_type, int):
+            option_type = OptionType.from_type(option_type)
+
         option = SlashCommandOption(
-            name=name,
-            type=opt_type,
+            name=option_name,
+            type=option_type,
             description=description,
```
i0bs:
```suggestion
            description=option_description,
```

interactions/models/internal/application_commands.py
i0bs: Making `description` the first pos. argument is a huge breaking change for client-facing code.

i0bs (changes requested):
I like inferring data for a slash command option, but there's some things I want to note:

- Changing `description` to be the first positional argument is a big breaking change, and shouldn't be changed. `name` and `description` should both respectively stay where they were originally.
- This proposal forces `description` to be filled in, while `name` does not. Seeing the internal behaviour for `@slash_command`, I don't see why we can't set at least a default value for a description.

LordOfPolls:
Closing as stale

PR #1345, opened 2023-04-13
feat: add reset_with_key and get_cooldown_time_with_key

chronosirius:
## Pull Request Type

- [x] Feature addition
- [ ] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement
- [x] Other: For consistency, get_cooldown_with_key is async

## Description

Cooldown resets with keys was a problem I had earlier; this solution is useful.
I made get_cooldown_with_key async for consistency as I said, so one doesn't have to remember which one is which.

`await command.cooldown.reset_with_key(other_user)`
`await command.cooldown.get_cooldown_time_with_key(other_user)`

vsc pylance did NOT complain.

## Changes

- made get_cooldown_with_key async
-  added reset_with_key
- added get_cooldown_time_with_key

## Related Issues

none

## Test Scenarios

none
## Python Compatibility

- [ ] I've ensured my code works on Python `3.10.x` (I don't have an installation of this.)

## Checklist

- [x] I've run the `pre-commit` code linter over all edited files (couldn't install idk why but vsc pylance gave no problems) (Github pre-commit workflow)
- [x] I've tested my changes on supported Python versions (3.11)

### DIFF
--- a/interactions/models/internal/cooldowns.py
+++ b/interactions/models/internal/cooldowns.py
@@ -401,8 +401,7 @@ class Cooldown:
         cooldown = await self.get_cooldown_with_key(key, create=create)
         if cooldown is not None:
             return cooldown.get_cooldown_time()
-        else:
-            return 0
+        return 0
 
     async def on_cooldown(self, context: "BaseContext") -> bool:
         """
@@ -457,8 +456,7 @@ class Cooldown:
         if cooldown is not None:
             cooldown.reset()
             return True
-        else:
-            return False
+        return False
 
 
 class MaxConcurrency:
--- a/interactions/models/internal/cooldowns.py
+++ b/interactions/models/internal/cooldowns.py
@@ -398,13 +398,11 @@ class Cooldown:
             key: The key to get the cooldown system for
             create: Whether to create a new cooldown system if one does not exist
         """
-
         cooldown = await self.get_cooldown_with_key(key, create=create)
         if cooldown is not None:
             return cooldown.get_cooldown_time()
         else:
             return 0
-
 
     async def on_cooldown(self, context: "BaseContext") -> bool:
         """
@@ -455,7 +453,6 @@ class Cooldown:
         Returns:
             True if the key existed and was reset successfully, False if the key didn't exist.
         """
-
         cooldown = await self.get_cooldown_with_key(key)
         if cooldown is not None:
             cooldown.reset()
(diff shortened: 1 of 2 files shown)


LordOfPolls:
These don't need to be async - all doing that does is limit what situations you can use these methods

chronosirius:
Yes that's true, I will remove async. When I was going through the other files though I'm pretty sure I saw a "this doesn't really need to be async, but it is for consistency." or something to that extent. (`Cooldown.reset_all`, found it!) I will remove async, if you already haven't made it non-async I can fix that too.

LordOfPolls:
Yeah, that method was a contentious point for ux. Rule of thumb is if it doesn't need to be async, it shouldn't be - however yes `reset_all` is an outlier and as changing it would be a breaking change, it remains

chronosirius:
Alright then I will return it to the way it was (reset_all only).

chronosirius:
Could you check it on 3.10 for me?

chronosirius:
How do I update documentation?
Edit: never mind, the documentation builder takes docstrings.

chronosirius:
haha, index.md still said `bot.load`! I fixed it to `bot.load_extension`.

LordOfPolls:
Your last change was out-of-scope. please revert it and create a new pr for that

chronosirius:
Alright.

interactions/models/internal/cooldowns.py
```diff
@@ -387,6 +387,22 @@ async def get_cooldown_time(self, context: "BaseContext") -> float:
         cooldown = await self.get_cooldown(context)
         return cooldown.get_cooldown_time()
 
+    async def get_cooldown_time_with_key(self, key: Any, *, create: bool = False) -> float:
```
LordOfPolls: I cant decide if these two added methods should be sync or async. 

They dont *need* to be async, but i can see why you chose to make them that way

chronosirius: maybe they should be sync, people with dashboards might be able to hack it and give owners a button to reset it, using flask they can only use sync. However, that does make it harder to remember (reset with context is async, but reset with key isn't), but if they're using a correct IDE (like VSC) it will complain. I'm actually more in favor of the former but I won't make a commit until you decide.

LordOfPolls: Make then sync

chronosirius: ok

i0bs: Keep in mind that slapping `async` on a method with an already defined type return won't mean the type being returned is the same.
image

chronosirius: yes, but it will complain if a sync function (the chances are more likely that one will write async than forget it) is awaited.

PR #1367, opened 2023-05-01
feat: pass event object based on listeners signature

artem30801:
## Pull Request Type

- [X] Feature addition
- [X] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description & changes
Event object is now passed (or not) based on whether there are any parameters in listener

## Test Scenarios
Tested on my bot and with test script - both as standalone function listeners and as extension methods

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/models/internal/listener.py
+++ b/interactions/models/internal/listener.py
@@ -3,7 +3,7 @@ import inspect
 from typing import Callable
 
 from interactions.api.events.internal import BaseEvent
-from interactions.client.const import MISSING, Absent, AsyncCallable, get_logger
+from interactions.client.const import MISSING, Absent, AsyncCallable
 from interactions.client.utils import get_event_name
 from interactions.models.internal.callback import CallbackObject
 
--- a/interactions/models/internal/listener.py
+++ b/interactions/models/internal/listener.py
@@ -3,7 +3,7 @@ import inspect
 from typing import Callable
 
 from interactions.api.events.internal import BaseEvent
-from interactions.client.const import MISSING, Absent, AsyncCallable
+from interactions.client.const import MISSING, Absent, AsyncCallable, get_logger
 from interactions.client.utils import get_event_name
 from interactions.models.internal.callback import CallbackObject
 
@@ -30,6 +30,7 @@ class Listener(CallbackObject):
         delay_until_ready: bool = False,
         is_default_listener: bool = False,
         disable_default_listeners: bool = False,
+        pass_event_object: bool = True,
     ) -> None:
         super().__init__()
 
@@ -41,6 +42,9 @@ class Listener(CallbackObject):
         self.delay_until_ready = delay_until_ready
         self.is_default_listener = is_default_listener
         self.disable_default_listeners = disable_default_listeners
+        self.pass_event_object = pass_event_object
+
+        self.warned_no_event_arg = False
 
     def __repr__(self) -> str:
         return f"<Listener event={self.event!r} callback={self.callback!r}>"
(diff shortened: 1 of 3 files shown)


silasary (approved):

LordOfPolls:
Purely because I'm certain we are going to get people saying "hOw Do I gEt ThE EvEnT" can you make the listener throw a warning if an event that would normally have a Param (ie any apart from startup and ready) don't on instantiation?

LordOfPolls:
Also, would it be worth moving the inspect call to the listeners instantiation too, as inspect is very expensive, and this is a high traffic part of the lib

interactions/models/internal/listener.py
```diff
@@ -88,12 +93,16 @@ def wrapper(coro: AsyncCallable) -> "Listener":
                 if not name:
                     name = coro.__name__
 
+            # If the coroutine has no parameters, we don't need to pass the event object
+            pass_event_object = len(coro_signature.parameters) != 0
```
LordOfPolls: This doesn't account for `self`

LordOfPolls (changes requested):

interactions/models/internal/listener.py
artem30801: It does account for self as callback we are getting (from extensions) is a _bound method_ and doesn't have self parameter in the signature

LordOfPolls: I tested it, it doesnt - and no, they arent bound methods proxy

artem30801: They were when the check was in event dispatcher... Welp, fixed now

interactions/models/internal/listener.py
```diff
@@ -73,6 +77,14 @@ def wrapper(coro: AsyncCallable) -> "Listener":
             if not asyncio.iscoroutinefunction(coro):
                 raise TypeError("Listener must be a coroutine")
 
+            coro_parametrers = inspect.signature(coro).parameters.copy()
+            if coro_parametrers and list(coro_parametrers.keys())[0] == "self":
+                # remove self; those who name it differently are on their own
```
silasary: I see how it is 😂

silasary (approved):

interactions/models/internal/listener.py
artem30801: Unfortunately as far as I was able to find it's impossible to tell if the decorated function is a method othervise. `inspect.ismethod` won't work because the decorated function becomes method only *after* being decorated.

LordOfPolls: I think I have a workaround. Just got home. I'll have a toy in a bit

interactions/models/internal/listener.py
```diff
@@ -73,6 +77,14 @@ def wrapper(coro: AsyncCallable) -> "Listener":
...
+                coro_parametrers.pop("self")
+
+            # If the coroutine has no parameters, we don't need to pass the event object
+            pass_event_object = len(coro_parametrers) != 0
```
mAxYoLo01:
```suggestion
            coro_parameters = inspect.signature(coro).parameters.copy()
            if coro_parameters and list(coro_parameters.keys())[0] == "self":
                # remove self; those who name it differently are on their own
                coro_parameters.pop("self")

            # If the coroutine has no parameters, we don't need to pass the event object
            pass_event_object = len(coro_parameters) != 0
```
I hate typos sorry

interactions/models/internal/listener.py
LordOfPolls: 8d3141cce765e496c98cfddf3593d61b25723307 should resolve this for all parties

silasary (approved):
That's a lot nicer looking 😄

LordOfPolls (approved):

PR #1370, opened 2023-05-01
fix!: pass a Member as author of MessageReactionRemove

mAxYoLo01:
## Pull Request Type

- [ ] Feature addition
- [x] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
Reported on the Support server, MessageReactionRemove does not pass the author as a Member even if the event is fired from a guild

## Test Scenarios
```py
@interactions.listen(interactions.events.MessageReactionRemove)
async def on_reaction_remove(event: interactions.events.MessageReactionRemove):
    print(type(event.author))
```

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/api/events/processors/reaction_events.py
+++ b/interactions/api/events/processors/reaction_events.py
@@ -15,7 +15,10 @@ class ReactionEvents(EventMixinTemplate):
         if member := event.data.get("member"):
             author = self.cache.place_member_data(event.data.get("guild_id"), member)
         else:
-            author = await self.cache.fetch_user(event.data.get("user_id"))
+            if guild_id := event.data.get("guild_id"):
+                author = await self.cache.fetch_member(guild_id, event.data.get("user_id"))
+            else:
+                author = await self.cache.fetch_user(event.data.get("user_id"))
 
         emoji = PartialEmoji.from_dict(event.data.get("emoji"))  # type: ignore
         message = self.cache.get_message(event.data.get("channel_id"), event.data.get("message_id"))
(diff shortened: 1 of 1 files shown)


interactions/api/events/processors/reaction_events.py
```diff
@@ -15,7 +15,10 @@ async def _handle_message_reaction_change(self, event: "RawGatewayEvent", add: b
         if member := event.data.get("member"):
             author = self.cache.place_member_data(event.data.get("guild_id"), member)
         else:
-            author = await self.cache.fetch_user(event.data.get("user_id"))
+            if guild_id := event.data.get("guild_id"):
```
Damego: You can make it like
```py
elif guild_id := event.data.get("guild_id"):
```

Damego (approved):

interactions/api/events/processors/reaction_events.py
```diff
@@ -14,6 +15,8 @@ class ReactionEvents(EventMixinTemplate):
     async def _handle_message_reaction_change(self, event: "RawGatewayEvent", add: bool) -> None:
         if member := event.data.get("member"):
             author = self.cache.place_member_data(event.data.get("guild_id"), member)
+        elif guild_id := event.data.get("guild_id"):
+            author = await self.cache.fetch_member(guild_id, event.data.get("user_id"))
         else:
             author = await self.cache.fetch_user(event.data.get("user_id"))
```
LordOfPolls, line 21: You need to account for the edge case where `guild_id` is set, the member is not, and the member is not cached

LordOfPolls (changes requested):

interactions/api/events/processors/reaction_events.py
mAxYoLo01: Isn't that already the case? If the member is not cached, then it will fetch it automatically no?

LordOfPolls: I cant even remember what my thought process was 14 hours ago. resolving

LordOfPolls (approved):

PR #1386, opened 2023-05-05
docs: Suppress Mkdocs warnings

mAxYoLo01:
## Pull Request Type

- [ ] Feature addition
- [ ] Bugfix
- [x] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description

This PR will focus solely on docs.
For the moment, it is kept as a draft until I find how to resolve a few more issues I could see that would make our current docs better.

## Python Compatibility

## Checklist


mAxYoLo01:
Wow just saw that my isort is not happy with the repo lol

interactions/ext/prefixed_commands/context.py
```diff
@@ -80,7 +80,7 @@ async def reply(
         content: Optional[str] = None,
         embeds: Optional[Union[Iterable[Union[Embed, dict]], Union[Embed, dict]]] = None,
         embed: Optional[Union[Embed, dict]] = None,
-        **kwargs,
+        **kwargs: dict,
```
silasary: This is not correct.
`**` typehints are implicitly `dict[str, T]` where T is your provided hint.  You're saying "Any additional args must be dicts", where I'd normally expect **kwargs to be hinted `Any`

silasary (changes requested):

interactions/ext/prefixed_commands/context.py
mAxYoLo01: In the libs code, I found there are probably 10 different ways kwargs are typehinted, the simplest one was just a basic dict so I went with that to please MkDocs

mAxYoLo01:
PR ready, I forgot it existed lol

interactions/ext/prefixed_commands/context.py
silasary: You're misunderstanding [PEP 484](https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values) here.  
This is what you've annotated the parameter as: 
image

Yes, there are some other places in the library that also do it incorrectly.  (And also some places where the typing is in fact [Dict[str, Dict[str, Any]]](https://github.com/interactions-py/interactions.py/blob/stable/interactions/api/http/http_client.py#L350)) 

You can fix them, or you can not.  But I do not want to merge a PR that makes it worse.

mAxYoLo01: Ok I indeed was misunderstanding how it typehinted kwargs, thanks for that info, I have modified it.

silasary (approved):

PR #1395, opened 2023-05-09
fix: correctly handle message references

AstreaTSS:
## Pull Request Type

- [ ] Feature addition
- [x] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
Just two bugfixes relating to message references that could cause strange behavior otherwise.

## Changes

- Since Discord annoyingly doesn't always give `referenced_message` even if it is a valid referenced message and a message reference object exists, add a workaround to set `_referenced_message_id` as the message reference object's message ID.
- Correctly catch and return none for `Message.fetch_referenced_message` to match the other fetches.

## Test Scenarios
It's a bit weird to trigger this, but I've found that creating a large amount of replies will make at least one message not `referenced_message` and so trigger this.

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/models/discord/message.py
+++ b/interactions/models/discord/message.py
@@ -400,6 +400,8 @@ class Message(BaseMessage):
         """
         if self._referenced_message_id is None:
             return None
+
+
         return await self._client.cache.fetch_message(self._channel_id, self._referenced_message_id, force=force)
 
     def get_referenced_message(self) -> Optional["Message"]:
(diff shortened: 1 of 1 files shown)


LordOfPolls (approved):
I'm not able to replicate the edgecase you've found, however your changes look fine

My test case:
[ send 50 messages replying to one message]
image

[ reboot bot ] 
```
history = channel.history(after=1105400191508545566, limit=50)
async for message in history:
    assert message.message_reference is not None
```

AstreaTSS:
Wack. I did trigger it through using my own account and creating a chain of replies manually until I saw the elif was triggered, but who knows.

PR #1406, opened 2023-05-17
feat: Handle async checks in wait_for_component

AbbyRH:
## Pull Request Type

- [x] Feature addition
- [ ] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
Following a help thread on discord I found that `client.wait_for_component()` did not support async checks, despite the fact that the underlying `client.wait_for()` and `Wait` class both do. 
This PR copies that functionality to apply to the check passed to `wait_for_component()`

## Changes
- use `asyncio.iscoroutinefunction()` to determine if we need to await on a check or call it normally

## Test Scenarios
I am unsure how to go about unit testing this due to my limited experience with asyncio and pytest. However I have validated this fixes the issues with the following code

```python
@slash_command(name="test", scopes=[1094385160536993884])
    async def test(self, ctx: SlashContext):
        components = Button(
            custom_id="my_button_id",
            style=ButtonStyle.GREEN,
            label="Click Me",
        )

        message = await ctx.channel.send("Look a Button!", components=components)

        # define the check
        async def check(res: events.Component) -> bool:
            await res.ctx.send("You clicked the button!", ephemeral=True)
            return res.ctx.author.display_name.startswith("A")

        try:

            used_component = await self.bot.wait_for_component(components=components, check=check, timeout=30)

        except TimeoutError:
            print("Timed Out!")

            components.disabled = True
            await message.edit(components=components)

        else:
            await used_component.ctx.send("Your name starts with 'a'")
```

I forgot to install pre-commit before making my commits, but `pre-commit run --all-files` passes

```
poetry run pre-commit run --all-files
[INFO] Initializing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Initializing environment for https://github.com/charliermarsh/ruff-pre-commit.
[INFO] Initializing environment for https://github.com/psf/black.
[INFO] Installing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Installing environment for https://github.com/charliermarsh/ruff-pre-commit.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Installing environment for https://github.com/psf/black.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
Requirements.............................................................Passed
Debugging................................................................Passed
Trailing Whitespace......................................................Passed
EOF Newlines.............................................................Passed
YAML Structure...........................................................Passed
TOML Structure...........................................................Passed
Merge Conflicts..........................................................Passed
ruff.....................................................................Passed
Black Formatting.........................................................Passed
```

## Python Compatibility

- [X] I've ensured my code works on Python `3.11.x`

## Checklist

### DIFF
--- a/interactions/client/client.py
+++ b/interactions/client/client.py
@@ -1105,7 +1105,7 @@ class Client(
                 dict,
             ]
         ] = None,
-        check: Optional[Callable] = None,
+        check: Absent[Optional[Union[Callable[..., bool], Callable[..., Awaitable[bool]]]]] = None,
         timeout: Optional[float] = None,
     ) -> "events.Component":
         """
--- a/interactions/client/client.py
+++ b/interactions/client/client.py
@@ -1136,7 +1136,7 @@ class Client(
         if custom_ids and not all(isinstance(x, str) for x in custom_ids):
             custom_ids = [str(i) for i in custom_ids]
 
-        def _check(event: events.Component) -> bool:
+        async def _check(event: events.Component) -> bool:
             ctx: ComponentContext = event.ctx
             # if custom_ids is empty or there is a match
             wanted_message = not message_ids or ctx.message.id in (
@@ -1144,6 +1144,8 @@ class Client(
             )
             wanted_component = not custom_ids or ctx.custom_id in custom_ids
             if wanted_message and wanted_component:
+                if asyncio.iscoroutinefunction(check):
+                    return bool(check is None or await check(event))
                 return bool(check is None or check(event))
             return False
 


AbbyRH:
@interactions-py/core over to you to review :)

AstreaTSS:
Editing the title and the PR description to note that this is actually a feature: while it's an odd exclusion and inconsistent with `wait_for`, `check` only handling synchronous functions was noted correctly in its typehints.

AstreaTSS (approved):
Code looks good - I haven't tested it though, so I prefer someone else to do so (though I doubt this code would break).

One little thing to note for any contributor doing similar things to this: `maybe_coroutine` in `utils` exists for handling functions that may or may be synchronous (which would have been useful here). Frankly, I think people (even Polls as indicated by the `Wait` class) forgot about this function because it's only extensively used in prefixed commands.
Note though that I don't think this PR needs to be changed to use it - I think I'll do a PR after this is merged to clean up multiple pieces code that should use it anyhow.

LordOfPolls:
@AstreaTSS 
> While it's an odd exclusion and inconsistent with wait_for, check only handling synchronous functions 

Originally we had no intent to make them async, this was a conclusion reached a while ago - however at the request of users we opened up predicates to be support coroutines - evidently `wait_for_component` slipped under the radar, as most people treat it purely as a filter predicate (as it was designed to be), rather than a part of the ux flow

With that said, I have no issue with this being merged, as the performance penalty is purely at the discretion of the user. However, I won't be able to test it until tomorrow evening at the earliest 



Good catch on `maybe_coroutine`, completely forgot that was in utils

i0bs (approved):
our current predicates allow coroutine yields so we can prob let this slide too, i don't see anything stopping us from using it as an async workflow

PR #1421, opened 2023-05-24
fix: correctly work with Permissions.NONE when passed into Guild.create_role

### LINKED ISSUE #1420: [BUG] Unable to create new `Role` with `Permissions.NONE`
### Library Version

5.5.1

### Describe the Bug

When attempting to create a new guild role with permissions set to `Permissions.NONE` (i.e. the role should have all permissions disabled), the role is instead created with Discord's default set of permissions for new guild roles.

### Steps to Reproduce

Call the function `create_role` on a `interactions.Guild` object, and attempt to create a new role, passing in `permissions=interactions.Permissions.NONE`. Then through the Discord UI, observe the permissions enabled for the role. Notice that the default permissions for guild roles are all enabled.

### Expected Results

When calling `interactions.Guild.create_role` with `permissions=interactions.Permissions.NONE`, the role created should have all guild permissions disabled on it.

### Minimal Reproducible Code

```python
import interactions as ipy

bot = ipy.Client(intents=ipy.Intents.DEFAULT, debug_scope=<your_guild_id>)

@ipy.slash_command()
async def create_role(ctx: ipy.SlashContext):
    role = await ctx.guild.create_role(name="test role", permissions=ipy.Permissions.NONE)
    await ctx.send(f"role's new permissions: {role.permissions}")

if __name__ == "__main__":
    bot.start("<token>")
```

### Traceback

_No response_

### Checklist

- [X] I have searched the open issues for duplicates.
- [X] I have shown the entire traceback, if possible.
- [X] I have removed my token from display, if visible.
- [X] I have attempted to debug this myself, and I believe this issue is with the library

### Additional Information

I think the bug is just an improper falsey check on the `permissions` argument.

ShajeshJ:
## Pull Request Type

- [ ] Feature addition
- [x] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description

The `interactions.Guild.create_role` method does not work correctly if you pass in `interactions.Permissions.NONE`. Rather than creating the role in Discord with every permission disabled, it instead creates the role with discord's default permissions enabled.

## Changes

- Update `interactions.Guild.create_role` method
  - Updated the arg check for adding `permissions` to the payload to check equality to`interactions.Permissions.NONE` in addition to truthy values

## Related Issues

fixes #1420

## Test Scenarios

- Setup a bot like below:
  ```python
  import interactions as ipy
  
  bot = ipy.Client(intents=ipy.Intents.DEFAULT, debug_scope=<your_guild_id>)
  
  @ipy.slash_command()
  async def create_role(ctx: ipy.SlashContext):
      role = await ctx.guild.create_role(name="test role", permissions=ipy.Permissions.NONE)
      await ctx.send(f"role's new permissions: {role.permissions}")
  
  if __name__ == "__main__":
      bot.start("<token>")
  ```
- Prior to this change, you will see the bot respond with the permission set value being `137408993025601` (discord's default)
- After this change, you should instead see the bot respond with the permission set value being `0`

I didn't add any unit tests because it seemed like overkill for such a small edge case. But I can add one in if folks disagree.

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/models/discord/guild.py
+++ b/interactions/models/discord/guild.py
@@ -1457,7 +1457,7 @@ class Guild(BaseGuild):
         if name:
             payload["name"] = name
 
-        if permissions or permissions == Permissions.NONE:
+        if permissions is not MISSING and permissions is not None:
             payload["permissions"] = str(int(permissions))
 
         if colour := colour or color:
--- a/interactions/models/discord/guild.py
+++ b/interactions/models/discord/guild.py
@@ -1457,7 +1457,7 @@ class Guild(BaseGuild):
         if name:
             payload["name"] = name
 
-        if permissions:
+        if permissions or permissions == Permissions.NONE:
             payload["permissions"] = str(int(permissions))
 
         if colour := colour or color:


interactions/models/discord/guild.py
```diff
@@ -1457,7 +1457,7 @@ async def create_role(
         if name:
             payload["name"] = name
 
-        if permissions:
+        if permissions or permissions == Permissions.NONE:
```
AstreaTSS: It's probably better to do `permissions is not MISSING` for the entire statement instead, instead of awkwardly working around a truthy.

AstreaTSS (commented):
Seems good, just one suggestion.

interactions/models/discord/guild.py
ShajeshJ: Cool, yeah that sense. I'll make that change.

ShajeshJ: Actually I just noticed the type hint on this function indicates `Optional` as well. Do you think `permissions is not MISSING and permissions is not None` would be more appropriate?

AstreaTSS: The optionals seem like a mistake looking at it, but that's a tad bit out of the scope of the PR. I'd say go for the change.

AstreaTSS (approved):

interactions/models/discord/guild.py
ShajeshJ: Great I've made the change. And thanks for the quick review 🙏

PR #1440, opened 2023-06-05
feat:  Warn if people try to run d.py alongside interactions

silasary:
## Pull Request Type

- [x] Feature addition
- [ ] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
We frequently recieve support requests where provided code snippits start with the line "import discord" and inevitably don't work. 
Due to the nature of humans generally importing stuff alphabetically, we can do runtime check for this import, and yell at them.

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/__init__.py
+++ b/interactions/__init__.py
@@ -1,3 +1,4 @@
+import sys
 from .client import (
     __api_version__,
     __py_version__,
@@ -680,6 +681,10 @@ __all__ = (
     "WebSocketOPCode",
 )
 
+if "discord" in sys.modules:
+    get_logger().error(
+        "`import discord` import detected.  Interactions.py is a completely separate library, and is not compatible with d.py models.  Please see https://interactions-py.github.io/interactions.py/Guides/100%20Migration%20From%20D.py/ for how to fix your code."
+    )
 
 ########################################################################################################################
 # Noteworthy Credits
--- a/interactions/client/const.py
+++ b/interactions/client/const.py
@@ -283,8 +283,3 @@ except AssertionError as e:
     raise RuntimeError(
         f"NON_RESUMABLE_WEBSOCKET_CLOSE_CODES contains codes that are not in RECOVERABLE_WEBSOCKET_CLOSE_CODES: {diff}"
     ) from e
-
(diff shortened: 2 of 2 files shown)


LordOfPolls:
Alternate idea. Move this check to client's sanity check method

LordOfPolls:
Alternate idea. Move this check to client's sanity check method

silasary:
Considered, implemented, then reverted upon further consideration 😂 

Take this thread as an example: https://discord.com/channels/789032594456576001/1110958056327626753/1110959282213621860

```py
import discord
from discord import app_commands
from discord.ext import commands
from interactions import slash_command, SlashContext, Permissions, slash_default_member_permission, Extension
...
bot = commands.Bot (command_prefix="*", intents = discord.Intents.all())
slash = slash_command(bot, sync_commands=True)
command_count = 0
staff = Permissions.MANAGE_ROLES

class Player(commands.Cog):
    def __init__(self, bot):
        self.bot = bot

    @slash_command(name="X", description="X")
    @slash_default_member_permission(staff)
    async def hello(self, ctx: SlashContext):
        await ctx.send(f"Hey {ctx.author.mention}! Test")


bot.add_cog(Player(bot))

@bot.event
async def on_ready():
    print(f"Bot conectado como {bot.user.name}")
    await slash.sync_all_commands()

token = os.getenv('TOKEN')
bot.run(token)
```

They're using interactions decorators on a d.py client.  They're still going to run into issues, but they won't ever call sanity check.

LordOfPolls:
Id argue that's extreme user error at that point. But fair point, then I have one more suggestion. Put it in the root init

silasary:
That was my initial plan, but it felt weird.

LordOfPolls:
It's standard practice for checks like this. Const feels weird, imo

silasary:
Understandable.  I'll move it to init

LordOfPolls:
oh fml. this targetted stable... dammit sil, i trusted you

silasary:
Ooops. I blame the fact that stable and unstable were at the same place when I made the PR.

PR #1456, opened 2023-06-21
feat: add permissions attr to Member

AstreaTSS:
## Pull Request Type

- [x] Feature addition
- [ ] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
Quick PR to add in `permissions` to `Member`, as that's given in interaction objects and can be useful.

## Changes
- Add `permissions` to `Member`.

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/models/discord/user.py
+++ b/interactions/models/discord/user.py
@@ -311,6 +311,12 @@ class Member(DiscordObject, _SendDMMixin):
         converter=optional_c(timestamp_converter),
         metadata=docs("When a member's timeout will expire, `None` or a time in the past if the user is not timed out"),
     )
+    permissions: Optional[Permissions] = attrs.field(
+        repr=False,
+        default=None,
+        converter=optional_c(Permissions),
+        metadata=docs("Calculated permissions for the member, only given in interactions"),
+    )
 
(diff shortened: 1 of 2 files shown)


LordOfPolls:
The shouldnt be in the member object as this is instance-specific, not guild specific. Imo the context object should intercept and store this data, that way its ephemeral just like this data is

AstreaTSS:
...how then? Do note that this is given for member objects in options too.

LordOfPolls:
I said? The context object

AstreaTSS:
Do you... want a mapping? Like `ctx.permissions[member]`? That doesn't sound that great for UX, those I suppose that could simplify channel permissions too.

LordOfPolls:
I have an idea. Give me 24 hours to draft something

i0bs:
Do we statically dispatch events, or are events aware of changes while being processed? If it's static – a mapping would work well, we'd just need to be considerate of indexing bounds

AstreaTSS:
Closing this PR early as regardless of what happens, this likely isn't the way this feature will be implemented.

PR #1462, opened 2023-07-02
docs: make autocomplete usage clearer

### LINKED ISSUE #1349: [Docs]: autocomplete options data (args, kwargs)
### Page

https://interactions-py.github.io/interactions.py/API%20Reference/API%20Reference/models/Internal/context/

### Problem description

Due to significant changes to version 5, the way that options are fetched for autocomplete contexts has changed drastically. While we used to fetch this from the context payload data, it is now being stored in the args and kwargs attributes.
I was not able to find anywhere on the documentation about these changes as well as anything to suggest that the data is being stored in here, I had to rely on reading source code to figure out there's already a method that processes them while not providing inline documentation of what it's attempting of doing

### Suggested changes

- Add autocomplete options to the [guide](https://interactions-py.github.io/interactions.py/Guides/03%20Creating%20Commands/#i-need-more-than-25-choices) linked in the migration documentation
- Improve definition of kwargs/args in context documentation

### Additional context

_No response_

AstreaTSS:
## Pull Request Type

- [ ] Feature addition
- [ ] Bugfix
- [x] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
This PR makes the autocomplete part of the slash command guide clearer. This basically resolves #1349's original complaint, though this doesn't change anything about `AutocompleteContext`'s API reference.

## Changes

- Make it clearer what `ctx.input_text` is through a variable.
- Note how to get other options in the command (through `ctx.kwargs`).
- Adjust responding comment wording.
- 

## Related Issues
#1349 

## Python Compatibility

## Checklist

### DIFF
--- a/docs/src/Guides/03 Creating Commands.md
+++ b/docs/src/Guides/03 Creating Commands.md
@@ -260,20 +260,23 @@ from interactions import AutocompleteContext
 
 @my_command.autocomplete("string_option")
 async def autocomplete(self, ctx: AutocompleteContext):
-    # make sure this is done within three seconds
+    string_option_input = ctx.input_text  # can be empty
+    # you can use ctx.kwargs.get("name") for other options - note they can be empty too
+
+    # make sure you respond within three seconds
     await ctx.send(
         choices=[
             {
-                "name": f"{ctx.input_text}a",
-                "value": f"{ctx.input_text}a",
+                "name": f"{string_option_input}a",
+                "value": f"{string_option_input}a",
             },
             {
-                "name": f"{ctx.input_text}b",
-                "value": f"{ctx.input_text}b",
+                "name": f"{string_option_input}b",
+                "value": f"{string_option_input}b",
             },
             {
-                "name": f"{ctx.input_text}c",
-                "value": f"{ctx.input_text}c",
+                "name": f"{string_option_input}c",
+                "value": f"{string_option_input}c",
             },
         ]
     )


i0bs (commented):
I'm sorry, but I'm -1 for this. I don't see any benefit we get out of assigning `ctx.input_text` to a `string_option_input` variable. I think most developers will look at the already given example and use IntelliSense to determine what the `input_text` property method will do, including the docstring and type return.

I'm open to us maybe putting comments in the choice population instead, but I'm unable to find any good rationale for bringing this to our documentation.

AstreaTSS:
The idea is for the code to be understandable without an IDE or the like - admittedly, using `ctx.input_text` alone might make that clear, but I felt it couldn't hurt to add more clarification and be explicit rather than implicit.

I'm willing to change it if need be, but I genuinely don't see the harm in it.

i0bs:
> The idea is for the code to be understandable without an IDE or the like [...]

Honestly, if developers aren't using IntelliSense for their Python projects when we already heavily rely on people to infer off of typings, docstrings, etc., that sounds like an L on their part. I think it'd be asinine if someone *wasn't* using a tool like that, especially when there's such prevalence for it among n/vim, JB, Atom and VSC editors.

AstreaTSS:
I disagree (at least in the context for a guide), but I think it's just better if someone else weighs in their opinion.

silasary (approved):
Seems like a good change to me

PR #1463, opened 2023-07-02
feat: Add support for return-style command callbacks

silasary:
## Pull Request Type

- [x] Feature addition
- [ ] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
This was in the original v5 plans, but fell to the side in the push for actually releasing in a reasonable time.

Allows commands to return a string or Embed rather than calling `await ctx.send()`.  Useful for lambda-style one-line commands.

## Changes
- Add typecheck to command return-value, and pass it to Context.send if applicable.

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/models/internal/command.py
+++ b/interactions/models/internal/command.py
@@ -22,9 +22,12 @@ from interactions.client.mixins.serialization import DictSerializationMixin
 from interactions.client.utils.attr_utils import docs
 from interactions.client.utils.misc_utils import get_parameters, get_object_name, maybe_coroutine
 from interactions.client.utils.serializer import no_export_meta
+from interactions.models.discord.embed import Embed
 from interactions.models.internal.callback import CallbackObject
 from interactions.models.internal.cooldowns import Cooldown, Buckets, MaxConcurrency, CooldownSystem
 from interactions.models.internal.protocols import Converter
+from interactions.client.mixins.send import SendMixin
+
 
 if TYPE_CHECKING:
     from interactions.models.internal.extension import Extension
@@ -122,7 +125,7 @@ class BaseCommand(DictSerializationMixin, CallbackObject):
 
         try:
             if await self._can_run(context):
-                if self.pre_run_callback is not None:
+                if self.pre_run_callback is not None:  # pragma: no branch
                     await self.call_with_binding(self.pre_run_callback, context, *args, **kwargs)
 
                 if self.extension is not None and self.extension.extension_prerun:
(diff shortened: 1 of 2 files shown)


LordOfPolls:
This is already supported. Please can you confirm if the existing system isn't working? If it is broken please can you fix it rather than rolling a new solution on top?

i0bs:
> This is already supported. Please can you confirm if the existing system isn't working? If it is broken please can you fix it rather than rolling a new solution on top?

I thought we only supported it in the sense of `return .send()` logic, not strings, is this not the case?

interactions/models/internal/command.py
```diff
@@ -195,7 +198,11 @@ def param_config(self, annotation: Any, name: str) -> Tuple[Callable, Optional[d
         return (annotation, getattr(annotation, name, None))
 
     async def call_callback(self, callback: Callable, context: "BaseContext") -> None:
-        await self.call_with_binding(callback, context, **context.kwargs)  # type: ignore
+        val = await self.call_with_binding(callback, context, **context.kwargs)  # type: ignore
+        if isinstance(val, str) and isinstance(context, SendMixin):
```
i0bs, line 202: So, crazy idea – what if, on top of allowing string and embed returns, we allow for numerous embeds and/or files to be returned for interaction responses as well? This could also be really useful for one-line lambda-esque returns, however, I can see the cons being presented for this.

What would stop us from doing this instead of telling users to use `.send()` instead?

LordOfPolls:
Reverted to draft as this pr requires additional changes per discussion

interactions/models/internal/command.py
```diff
@@ -195,7 +198,11 @@ def param_config(self, annotation: Any, name: str) -> Tuple[Callable, Optional[d
...
+            await context.send(val)
+        elif isinstance(val, Embed) and isinstance(context, SendMixin):
+            await context.send(embed=val)
```
i0bs, line 205: Haven't tested this, but this brings forth the proposal in suggestion form.

```suggestion
            await context.send(embed=val)
        elif all(isinstance(val, list[Embed])) and isinstance(context, SendMixin):
            await context.send(embeds=[embed for embed in val])
        elif isinstance(val, File) and isinstance(context, SendMixin):
            await context.send(file=val)
        elif all(isinstance(val, list[File])) and isinstance(context, SendMixin):
            await context.send(files=[file for file in val])
```

i0bs (changes requested):

interactions/models/internal/command.py
silasary: Fun fact: `isinstance(..., list[T])` doesn't do what you think it does. It just checks that the object is a list, and ignores the generic parameter entirely.  

There are a couple bigger changes I need to make to the PR before it lands, and I'm hoping to have time for them this weekend

LordOfPolls:
Closing as stale

PR #1438, opened 2023-06-03
fix OrTrigger fire endless

Donbur4156:
## Pull Request Type

- [ ] Feature addition
- [X] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement
- [ ] CI/CD pipeline enhancement
- [ ] Other: [Replace with a description]

## Description

setting "last_call_time" in _fire() changes it for the OrTrigger but not for the relevant Triggers. 
The function set_last_call_time() set it for normal Trigger and in case of OrTrigger to the "current_trigger". 
The current_trigger is the trigger for which triggers next.

## Changes

- add var current_trigger: BaseTrigger in class OrTrigger
- extract evaluate current_trigger to function _set_current_trigger
- add function set_last_call_time to BaseTrigger and modified it in OrTrigger

## Related Issues

#1437

## Test Scenarios

```python
@listen()
async def on_startup():
    cron_print.start()

@Task.create(
    OrTrigger(
        IntervalTrigger(2),
        TimeTrigger(hour=19, minute=1, utc=False),
        DateTrigger(target_datetime=datetime(day=3,month=6,year=2023, hour=19, minute=1, second=30)),
    )
)
async def cron_print():
    print(datetime.now().strftime("%H:%M:%S.%f")[:-3])
```
set the timings in near future and probably comment out some triggers

## Python Compatibility

- [X] I've ensured my code works on Python `3.10.x`
- [ ] I've ensured my code works on Python `3.11.x`

## Checklist

- [ ] I've run the `pre-commit` code linter over all edited files
- [ ] I've tested my changes on supported Python versions
- [ ] I've added tests for my code, if applicable
- [ ] I've updated / added  documentation, where applicable

PR #1478, opened 2023-07-13
feat: add args, kwargs to task

kennhh:
## Pull Request Type

- [x] Feature addition
- [ ] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description

Adds variable length argument list and keyword arguments to task.
Example:
```py
@Task.create(...)
async def task(*args, **kwargs):
     ...

...
task.start(*args, **kwargs)
```

## Changes

- added args, kwargs to task

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/models/internal/tasks/task.py
+++ b/interactions/models/internal/tasks/task.py
@@ -75,25 +75,25 @@ class Task:
         self.on_error_sentry_hook(error)
         interactions.Client.default_error_handler("Task", error)
 
-    async def __call__(self) -> None:
+    async def __call__(self, *args, **kwargs) -> None:
         try:
             if inspect.iscoroutinefunction(self.callback):
-                val = await self.callback()
+                val = await self.callback(*args, **kwargs)
             else:
-                val = self.callback()
+                val = self.callback(*args, **kwargs)
 
             if isinstance(val, BaseTrigger):
                 self.reschedule(val)
         except Exception as e:
             self.on_error(e)
 
-    def _fire(self, fire_time: datetime) -> None:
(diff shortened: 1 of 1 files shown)


interactions/models/internal/tasks/task.py
```diff
@@ -75,25 +75,25 @@ def on_error(self, error: Exception) -> None:
...
 
             if isinstance(val, BaseTrigger):
                 self.reschedule(val)
         except Exception as e:
             self.on_error(e)
 
-    def _fire(self, fire_time: datetime) -> None:
+    def _fire(self, fire_time: datetime, *args, **kwargs) -> None:
         """Called when the task is being fired."""
-        self.trigger.set_last_call_time(fire_time)
```
AstreaTSS, line 92: Why is this changed?

AstreaTSS (changes requested):

interactions/models/internal/tasks/task.py
kennhh: i just couldn't traceback to set_last_call_time. might be an issue on my end but the quickest fix i found was just to use last_call_time

AstreaTSS: Suggesting this was added in another PR to fix something (#1438), I'm not sure if this is really the right solution.

kennhh: yeah, you're definitely right. seems i missed that pr (oops) so ill change it back. thanks for telling me!

AstreaTSS (approved):

i0bs (approved):

PR #1493, opened 2023-07-16
feat: add ability to use different arg name than option name

AstreaTSS:
## Pull Request Type

- [x] Feature addition
- [ ] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
This PR adds the ability for the argument name of an option for slash commands to be different from the option name itself. For example, this example works, despite the difference between `test` and `_test`.
```python
@interactions.slash_command(name="test", description="A test command.")
@interactions.slash_option(
    "test",
    "A thing.",
    interactions.OptionType.STRING,
    required=True,
    argument_name="_test",
)
async def test(ctx: interactions.SlashContext, _test: str):
    await ctx.send(_test)
```

To keep this non-breaking, `argument_name`/`name` had to be inserted into the current methods at... awkward positions, but it should work.

## Changes
- Introduce a new parameter to `SlashCommandOption`, `argument_name`.
  - On its own, it does basically nothing but exist. It's even removed when transformed into a dict, just to not break serialization (and since it isn't a Discord concept).
- Add options to the various ways of declaring slash command options to specify either the argument name or the name of the option as appropriate.
  - If they're not given, they default to their old behavior.
- Add an extra final step to parsing parameters for slash commands to find options with `argument_name` specified and essentially link `SlashCommandParameter` and the actual option name through it. Of course, this required a new parameter to `SlashCommandParameter`.
- Slightly adjust the callback logic to handle the translation between the two names smoothly.
- Make hybrid commands respect this logic (or, well, be able to handle it at least).

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/ext/hybrid_commands/hybrid_slash.py
+++ b/interactions/ext/hybrid_commands/hybrid_slash.py
@@ -351,7 +351,7 @@ def slash_to_prefixed(cmd: HybridSlashCommand) -> _HybridToPrefixedCommand:  # n
             # there isn't much we can do here
             raise ValueError("Autocomplete is unsupported in hybrid commands.")
 
-        name = str(option.name)
+        name = option.argument_name or str(option.name)
         annotation = inspect.Parameter.empty
         default = inspect.Parameter.empty
         kind = inspect.Parameter.POSITIONAL_ONLY if cmd._uses_arg else inspect.Parameter.POSITIONAL_OR_KEYWORD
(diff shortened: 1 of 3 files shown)


silasary (approved):

i0bs (approved):

LordOfPolls:
Please resolve the merge conflict and this can be merged

PR #1494, opened 2023-07-17
docs💥: revamp extension guide

AstreaTSS:
## Pull Request Type

- [ ] Feature addition
- [ ] Bugfix
- [x] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
This PR completely revamps the extension guide. While it was agreed it didn't need changing, it was one of the least favorite guides according to statistics, and was in need of a revamp anyhow (I might have gone overboard with it, but that's a good thing I hope 😅).

This PR also merges the normal and advanced extension guides, since both were short on their own. It technically makes it breaking for documentation, so...

## Changes
Like the event PR, see the diff.
- Revamp the extension guide (as well as merging the advanced guide). This updates many of the demonstrations of extensions to fit better with modern usage, and also showcases a number of things that were constantly asked about.

## Python Compatibility

## Checklist


docs/src/Guides/20 Extensions.md
```diff
@@ -1,248 +1,546 @@
 # Extensions
 
-Damn, your code is getting pretty messy now, huh? Wouldn't it be nice if you could organise your commands and listeners into separate files?
+## Introduction
 
-Well let me introduce you to `Extensions`<br>
-Extensions allow you to split your commands and listeners into separate files to allow you to better organise your project,
-as well as that, they allow you to reload Extensions without having to shut down your bot.
+Your code's getting pretty big and messy being in a single, huh? Wouldn't it be nice if you could organise your commands and listeners into separate files?
```
mAxYoLo01:
```suggestion
Your code's getting pretty big and messy being in a single file, huh? Wouldn't it be nice if you could organise your commands and listeners into separate files?
```

AstreaTSS:
I'm going to be adding one more section to this explaining how to load all extensions in a folder, since it seems to be a common thing to do.

i0bs (approved):

docs/src/Guides/20 Extensions.md
```diff
@@ -1,248 +1,581 @@
...
+
+```python
+bot = Client(...)
 bot.start("token")
 ```
 
-Now, for the cool bit of Extensions, reloading. Extensions allow you to edit your code, and reload it, without restarting the bot.
-To do this, simply run `bot.reload_extension("Filename_here")` and your new code will be used. Bare in mind any tasks your extension
-is doing will be abruptly stopped.
+To load the extension, you just need to use `bot.load_extension("filename_in_import_style")` before `bot.start`. So, in this case, it would look like this:
```
LordOfPolls:
```suggestion
To load the extension, you just need to use `bot.load_extension("filename.in_import_style")` before `bot.start`. So, in this case, it would look like this:
```
Suggest adding a dot, to differentiate between snake_case and import

docs/src/Guides/20 Extensions.md
```diff
@@ -1,248 +1,581 @@
...
+bot.load_extension("extension")
+bot.start("token")
+```
+
+And that's it! Your extension is now loaded and ready to go.
+
+#### "Import Style"
+
+You may have noticed that the filename is passed to `load_extension` without the `.py` extension. This is because interactions.py actually does an
+*import* when loading the extension, so whatever string you give it needs to be a valid Python import path. This means that if you have a file structure like this:
```
LordOfPolls:
```suggestion
In the example above, the filename is passed to `load_extension` without the `.py` extension. This is because interactions.py actually does an
*import* when loading the extension, so whatever string you give it needs to be a valid Python import path. This means that if you have a file structure like this:
```

docs/src/Guides/20 Extensions.md
```diff
@@ -1,248 +1,581 @@
...
-from interactions import Extension
+class MyExtension(Extension):
+    def __init__(self, bot):
+        # do some initialization here
+        pass
+```
+
+#### Asynchronous Initialization
+
+You may have noticed `__init__` is synchronous. This may pose problems if you're trying to do something asynchronous in it, so there are various ways of solving it.
```
LordOfPolls:
```suggestion
As normal, `__init__` is synchronous. This may pose problems if you're trying to do something asynchronous in it, so there are various ways of solving it.
```

docs/src/Guides/20 Extensions.md
```diff
@@ -1,248 +1,581 @@
...
+The only difference is the `self` variable - this is the instance of the extension that the command is being called in, and is
+standard for functions inside of classes. Events follow a similar principal.
+
+interactions.py will automatically add all commands and events to the bot when you load the extension (discussed later),
+so you don't need to worry about that.
+
+#### Accessing the Bot
+
+You may have noticed that the `self` variable is being used in the example above. This is because the bot is passed to the extension
+when it's loaded and is stored in the `self.bot` variable. This means you can access the bot from anywhere in your extension,
```
LordOfPolls:
```suggestion
When an extension is loaded, the library automatically sets the `bot` for you. With this in mind, to access your client, use `self.bot`.

Please do note, if your convention is to use `self.client` this is fine and supported natively as well. 
```

Removed the blatant spoonfeed and added useful info... lets be real, we shouldnt explain basic class usage, our policy is we expect basic python knowledge

docs/src/Guides/20 Extensions.md
```diff
@@ -1,248 +1,581 @@
...
+You may have some logic to do while unloading a specific extension. For that, you can override the `drop` method in your extension:
+
+```python
+class MyExtension(Extension):
+    def drop(self):
+        # do some cleanup here
+        super().drop()  # important - this part actually does the unloading
+```
+
+You may have noticed that `drop` is synchronous. If you need to do something asynchronous, you can create a task with `asyncio` to do it:
```
LordOfPolls:
```suggestion
The `drop` method is synchronous. If you need to do something asynchronous, you can create a task with `asyncio` to do it:
```

docs/src/Guides/20 Extensions.md
```diff
@@ -1,248 +1,581 @@
...
+```python
+class MyExtension(Extension):
+    ...
+
+def setup(bot):
+    # insert logic here
+    MyExtension(bot)
+```
+
+Notice how the `Extension` subclass is created inside the `setup` function, and does not need to do any special function to add the extension in beyond being created using the instance.
```
LordOfPolls:
```suggestion
Here the `Extension` subclass is instantiated inside the `setup` function, and does not need to do any special function to add the extension in beyond being created using the instance.
```
It's not being created, that was done 3 lines above

docs/src/Guides/20 Extensions.md
```diff
@@ -1,248 +1,581 @@
...
+class MyExtension(Extension):
+    ...
+
+def setup(bot, some_arg: int = 0):
+    MyExtension(bot, some_arg)
+```
+
+## Extension-Wide Checks
+
+Sometimes, it is useful to have a check ran before running any command in an extension. Thankfully, all you need to do is use `add_ext_check`:
```
LordOfPolls:
```suggestion
Sometimes, it is useful to have a check run before running any command in an extension. Thankfully, all you need to do is use `add_ext_check`:
```

docs/src/Guides/20 Extensions.md
```diff
@@ -1,248 +1,581 @@
...
+        # handle the error here
+        ...
+```
 
-bot.load_extension("Filename_here", default_extension=False, another_arg=3.14)
-# OR
-bot.load_extension("Filename_here", default_extension=True, some_arg=555)
+??? note "Error Handling Priority"
+    Command error handlers will be ran over extension error handlers, and extension error handlers will be ran over event error handlers.
+    Only one error handler will be ran.
```
LordOfPolls:
```suggestion
    Only one error handler will run. Similar to css, the most specific handler takes precedence. So in order of priority command error handlers -> extension -> listeners
```

LordOfPolls (changes requested):

LordOfPolls:
You may have noticed I have made some changes

AstreaTSS:
You may have noticed that I made my own commit addressing your review and also changing other things too.

LordOfPolls (approved):

PR #1509, opened 2023-07-30
feat: add ext.load and a_load

LordOfPolls:
## Pull Request Type

- [x] Feature addition
- [ ] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description

Adds ext.load and a_load methods for loading hooks

## Test Scenarios

```py
from interactions import Extension


class TestExt(Extension):
    def load(self, *args, **kwargs):
        print("Test extension hooked!")

    async def a_load(self, *args, **kwargs):
        print("Test extension async hooked!")
```

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/models/internal/extension.py
+++ b/interactions/models/internal/extension.py
@@ -155,7 +155,7 @@ class Extension:
         """Run on extension load"""
         ...
 
-    async def a_load(self, *args, **kwargs):
+    async def async_load(self, *args, **kwargs) -> Callable[[], ...]:
         """Run on extension load, if there is a valid event loop"""
         ...
 
--- a/interactions/models/internal/extension.py
+++ b/interactions/models/internal/extension.py
@@ -132,7 +132,7 @@ class Extension:
         instance.load(*args, **kwargs)
         with contextlib.suppress(asyncio.CancelledError):
             asyncio.get_running_loop()  # guard condition
-            _ = asyncio.create_task(instance.a_load(*args, **kwargs))
+            _ = asyncio.create_task(instance.async_load(*args, **kwargs))
 
         bot.dispatch(events.ExtensionLoad(extension=instance))
         return instance
--- a/interactions/models/internal/extension.py
+++ b/interactions/models/internal/extension.py
@@ -151,7 +151,7 @@ class Extension:
         """Get the listeners from this Extension."""
         return self._listeners
 
-    def load(self, *args, **kwargs):
+    def load(self, *args, **kwargs) -> Callable[[], ...]:
         """Run on extension load"""
         ...
 
--- a/interactions/models/internal/extension.py
+++ b/interactions/models/internal/extension.py
@@ -1,4 +1,5 @@
 import asyncio
+import contextlib
 import inspect
 import typing
 from typing import Awaitable, Dict, List, TYPE_CHECKING, Callable, Coroutine, Optional
@@ -128,6 +129,11 @@ class Extension:
             else:
                 raise TypeError("async_start is a reserved method and must be a coroutine")
 
+        instance.load(*args, **kwargs)
+        with contextlib.suppress(asyncio.CancelledError):
+            asyncio.get_running_loop()  # guard condition
+            _ = asyncio.create_task(instance.a_load(*args, **kwargs))
+
         bot.dispatch(events.ExtensionLoad(extension=instance))
         return instance
 
@@ -144,6 +150,14 @@ class Extension:
     def listeners(self) -> List["Listener"]:
         """Get the listeners from this Extension."""
         return self._listeners
+
+    def load(self, *args, **kwargs):
+        """Run on extension load"""
+        ...
+
+    async def a_load(self, *args, **kwargs):
+        """Run on extension load, if there is a valid event loop"""
+        ...
 
     def drop(self) -> None:
         """Called when this Extension is being removed."""


AstreaTSS (commented):
I have... many questions:
- Why does `load` even need to exist when `__init__` quite literally does the same thing?
- Why is `a_load` called `a_load` instead of something more intuitive like `async_load`?
- What's the point of using `a_load` if it only runs in certain situations? Most people are going to use it thinking it's a solution to using `load_extension` before the event loop exists, when it really doesn't work like that. I'm guessing what you intend for users to do is (somewhat awkwardly) use both `async_start` *and* `a_load`?

interactions/models/internal/extension.py
```diff
@@ -128,6 +129,11 @@ def __new__(cls, bot: "Client", *args, **kwargs) -> "Extension":
             else:
                 raise TypeError("async_start is a reserved method and must be a coroutine")
 
+        instance.load(*args, **kwargs)
+        with contextlib.suppress(asyncio.CancelledError):
```
i0bs, line 133: I very much dislike us suppressing this, we're ignoring potential cancellation and suspension of async call sites when we're specifically wanting to determine sync/async behaviour. I think only one of these states should be allowed, depending on the validity of the guard condition.

interactions/models/internal/extension.py
```diff
@@ -145,6 +151,14 @@ def listeners(self) -> List["Listener"]:
         """Get the listeners from this Extension."""
         return self._listeners
 
+    def load(self, *args, **kwargs):
+        """Run on extension load"""
+        ...
+
+    async def a_load(self, *args, **kwargs):
```
i0bs:
```suggestion
    async def async_load(self, *args, **kwargs):
```

interactions/models/internal/extension.py
```diff
@@ -128,6 +129,11 @@ def __new__(cls, bot: "Client", *args, **kwargs) -> "Extension":
             else:
                 raise TypeError("async_start is a reserved method and must be a coroutine")
 
+        instance.load(*args, **kwargs)
+        with contextlib.suppress(asyncio.CancelledError):
+            asyncio.get_running_loop()  # guard condition
+            _ = asyncio.create_task(instance.a_load(*args, **kwargs))
```
i0bs: `a_load` is a vague declaration name, I propose changing to `async_load` to clarify without having people guess or infer off of IntelliSense for async colouring

```suggestion
            _ = asyncio.create_task(instance.async_load(*args, **kwargs))
```

interactions/models/internal/extension.py
```diff
@@ -145,6 +151,14 @@ def listeners(self) -> List["Listener"]:
         """Get the listeners from this Extension."""
         return self._listeners
 
+    def load(self, *args, **kwargs):
```
i0bs: PEP 585 supports `[]` subscripting for generic aliasing returns, I propose modifying:

```suggestion
    def load(self, *args, **kwargs) -> Callable[[], ...]:
```

interactions/models/internal/extension.py
i0bs: Extending the proposal from `load()`, choose this over and resolve if the former is applied

```suggestion
    async def async_load(self, *args, **kwargs) -> Callable[[], ...]:
```

i0bs (changes requested):

LordOfPolls:
@AstreaTSS 

> Why does load even need to exist when __init__ quite literally does the same thing?

Completeness, and accounting for the fact people seem adverse to using `__init__`. Fully agree, it's a layer of separation without much gain, but its more intuitive for the aforementioned  - ux takes priority even if redundant 

> Why is a_load called a_load instead of something more intuitive like async_load?

I actually went back and forth on this, i landed on `a_load` for shortness but im really not married to it. If anything i dont even like the `load` phrasing. in retrospect `async_load` will be fine

> What's the point of using a_load if it only runs in certain situations? Most people are going to use it thinking it's a solution to using load_extension before the event loop exists, when it really doesn't work like that. I'm guessing what you intend for users to do is (somewhat awkwardly) use both async_start and a_load?

Your assertion here is weird, but I'll focus on the premise instead. I toyed without putting handling it under the same paradigm as a `async_start` - which is basically a psuedo listener, however `async_start` had backlash with its "cheating the event loop" by allowing you to queue without an event loop being created. If you need the asynchronous load call, you should be loading the extension from a coroutine - ie `asyncio.run(start_bot)` where start bot instantiates and loads extensions 

If deemed preferable, I can mirror `async_start`s behavior, but either way someone is going to be grumpy.

interactions/models/internal/extension.py
LordOfPolls: Was supposed to be `RuntimeError` - i guess that was removed during a rebase locally. 

Regardless, marking as resolved because im going to switch this out and only handle the loop check

interactions/models/internal/extension.py
LordOfPolls: Per conversation with Astrea. Approving this and related

i0bs:
> If anything i dont even like the `load` phrasing.

Agreed, can we name it `eat` or `consume`? I like snake terms instead, keep it 'hip.

AstreaTSS:
> Completeness, and accounting for the fact people seem adverse to using `__init__`.


Is... this actually a thing? Genuinely asking - I've never encountered a person who didn't want to do it, especially since discord.py does the same.

> `async_start` had backlash with its "cheating the event loop" by allowing you to queue without an event loop being created.


I'm going to be frank with you - the only people who care are a bit too much into asynchronous programming. For most users that just don't care, not making it queue-esque will make it not helpful.

> If you need the asynchronous load call, you should be loading the extension from a coroutine - ie `asyncio.run(start_bot)` where start bot instantiates and loads extensions


Fully agree here, but users seem to be adverse to that.

i0bs (approved):
Next reply SLA will be 48 hours, LGTM as-is for current track

i0bs:
> I'm going to be frank with you - the only people who care are a bit too much into asynchronous programming. For most users that just don't care, not making it queue-esque will make it not helpful.

As one of those people who "care a bit too much," I'm going to have to side with that consensus. I didn't really understand how important it was until after I started working with async call sites in lower level languages, it's absolutely crucial for us to respect the event loop for scheduling. I prefer us keeping a sane guideline for schedulers over convenience, we can always write facade layers to nicely interface with the async logic we write.

AstreaTSS:
It... does respect the event loop though. It's not a violation, it's simply a convenient queue. I understand lower-level languages make it important, but this is Python out of all things.

LordOfPolls:
PR reverted to draft - either way changes are required. 

`async_start` does not respect the event loop, that's not up for dbate - but that's user-friendliness I added for the sake of helping people newer to async programming. 

Under the same ethos, I will update this to do the same

i0bs:
Just to note, apparently `Callable` dislikes passing ellipsis as the return type, (pytest results) we could make the param spec the ellipsis and change return to `Any`, but this may violate our MyPy config

eightween:
Proposing this PR to be closed for being stale, continued interest should lead to a new branch/PR being set forth

PR #1523, opened 2023-08-08
fix: properly fix component/modal callback

AstreaTSS:
## Pull Request Type

- [ ] Feature addition
- [x] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
`component_callback` and `modal_callback` was *always* using the function name after the last patch due to some weirdness. This PR fixes that by simplifying the patch and... fixing parts of the patch itself, funny enough.

Frankly, I'm disappointed in myself how easy this fix was 😅. 

## Changes
- Pass in `custom_id` to the unpack helper/validator, not the empty `resolved_custom_id`.
- Removed the unused (other than for the mistake above) `resolved_custom_id` initial declaration - I think we all forgot that `resolved_custom_id` does not get passed into there without a `nonlocal`. No idea how `custom_id` does in this specific case, but it does.
  - Yes, this patch makes the code very similar to the code before the first patch. However, since the variable name inside the wrapper function is different from just `custom_id`, everything works now (for some reason).

## Test Scenarios
```python
@interactions.slash_command(name="test", description="My command")
async def test(ctx: interactions.SlashContext):
    await ctx.send(
        components=interactions.Button(
            style=interactions.ButtonStyle.PRIMARY, label="Test", custom_id="test"
        )
    )


@interactions.component_callback("test")
async def test_callback(ctx: interactions.ComponentContext):
    await ctx.edit_origin(content="Test")
```

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/models/internal/application_commands.py
+++ b/interactions/models/internal/application_commands.py
@@ -1170,7 +1170,6 @@ def component_callback(*custom_id: str | re.Pattern) -> Callable[[AsyncCallable]
         *custom_id: The custom ID of the component to wait for
 
     """
-    resolved_custom_id: tuple[str | re.Pattern, ...] | list[str] = []
 
     def wrapper(func: AsyncCallable) -> ComponentCommand:
         resolved_custom_id = custom_id or [func.__name__]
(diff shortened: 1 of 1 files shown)


silasary (approved):
nonlocal was only needed because we were redefining a name used in a higher scope.

This is much nicer

PR #1533, opened 2023-08-16
Fix bot voice state connection issues

tequa:
## Pull Request Type

- [ ] Feature addition
- [x] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
I encountered a bug where connecting and disconnecting the bot itself across different guilds caused weird state issues to happen. Like if you disconnect the bot from a voice channel in one guild, the bot in the other guild wouldn't disconnect properly. See linked-issue for full context.

The issue with this in voice_events.py. The problem is that that it's grabbing cached before data for the user ID, which is fine if you're doing voice updates for actual non-bot users, but if this event is happening on a bot, it's gonna pull the "user" data for that bot, which will be global across guilds. I updated the code so that the bot updates ignore the before data and don't dispatch a voice state update, since this voice state update would be global across all guilds and not useful in any way. This fixed the bug I was encountering.

## Related Issues
[Issue 1531](https://github.com/interactions-py/interactions.py/issues/1531)

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/api/events/processors/voice_events.py
+++ b/interactions/api/events/processors/voice_events.py
@@ -13,17 +13,17 @@ __all__ = ("VoiceEvents",)
 class VoiceEvents(EventMixinTemplate):
     @Processor.define()
     async def _on_raw_voice_state_update(self, event: "RawGatewayEvent") -> None:
-        after = await self.cache.place_voice_state_data(event.data)
-
         if str(event.data["user_id"]) == str(self.user.id):
             # User is the bot itself
-            before = self.cache.get_bot_voice_state(event.data["guild_id"])
+            before = copy.copy(self.cache.get_bot_voice_state(event.data["guild_id"])) or None
+            after = await self.cache.place_voice_state_data(event.data, update_cache=False)
             if vc := before:
                 # noinspection PyProtectedMember
                 await vc._voice_state_update(before, after, event.data)
         else:
             # User is not the bot
             before = copy.copy(self.cache.get_voice_state(event.data["user_id"])) or None
+            after = await self.cache.place_voice_state_data(event.data)
 
         self.dispatch(events.VoiceStateUpdate(before, after))
 
--- a/interactions/api/events/processors/voice_events.py
+++ b/interactions/api/events/processors/voice_events.py
@@ -14,18 +14,17 @@ class VoiceEvents(EventMixinTemplate):
     @Processor.define()
     async def _on_raw_voice_state_update(self, event: "RawGatewayEvent") -> None:
         after = await self.cache.place_voice_state_data(event.data)
-        is_bot = event.data["member"]["user"]["bot"]
 
-        if is_bot:
-            before_bot_guild_voice_state = self.cache.get_bot_voice_state(event.data["guild_id"])
-            self.dispatch(events.VoiceStateUpdate(before_bot_guild_voice_state, after))
-            if vc := before_bot_guild_voice_state:
+        if str(event.data["user_id"]) == str(self.user.id):
+            # User is the bot itself
+            before = self.cache.get_bot_voice_state(event.data["guild_id"])
+            if vc := before:
                 # noinspection PyProtectedMember
-                await vc._voice_state_update(after, event.data)
-            return
+                await vc._voice_state_update(before, after, event.data)
+        else:
+            # User is not the bot
+            before = copy.copy(self.cache.get_voice_state(event.data["user_id"])) or None
 
-        # Following code is for updates to non-bot users
-        before = copy.copy(self.cache.get_voice_state(event.data["user_id"])) or None
         self.dispatch(events.VoiceStateUpdate(before, after))
 
         if before and after:
--- a/interactions/api/events/processors/voice_events.py
+++ b/interactions/api/events/processors/voice_events.py
@@ -14,7 +14,7 @@ class VoiceEvents(EventMixinTemplate):
     @Processor.define()
     async def _on_raw_voice_state_update(self, event: "RawGatewayEvent") -> None:
         after = await self.cache.place_voice_state_data(event.data)
-        is_bot = event.data['member']['user']['bot']
+        is_bot = event.data["member"]["user"]["bot"]
 
         if is_bot:
             if vc := self.cache.get_bot_voice_state(event.data["guild_id"]):
--- a/interactions/api/events/processors/voice_events.py
+++ b/interactions/api/events/processors/voice_events.py
@@ -13,15 +13,18 @@ __all__ = ("VoiceEvents",)
 class VoiceEvents(EventMixinTemplate):
     @Processor.define()
     async def _on_raw_voice_state_update(self, event: "RawGatewayEvent") -> None:
-        before = copy.copy(self.cache.get_voice_state(event.data["user_id"])) or None
         after = await self.cache.place_voice_state_data(event.data)
+        is_bot = event.data['member']['user']['bot']
 
-        self.dispatch(events.VoiceStateUpdate(before, after))
-
-        if before and before.user_id == self.user.id:
+        if is_bot:
             if vc := self.cache.get_bot_voice_state(event.data["guild_id"]):
                 # noinspection PyProtectedMember
-                await vc._voice_state_update(before, after, event.data)
+                await vc._voice_state_update(after, event.data)
+            return
+
+        # Following code is for updates to non-bot users
+        before = copy.copy(self.cache.get_voice_state(event.data["user_id"])) or None
+        self.dispatch(events.VoiceStateUpdate(before, after))
 
         if before and after:
             if (before.mute != after.mute) or (before.self_mute != after.self_mute):
--- a/interactions/api/events/processors/voice_events.py
+++ b/interactions/api/events/processors/voice_events.py
@@ -17,7 +17,9 @@ class VoiceEvents(EventMixinTemplate):
         is_bot = event.data['member']['user']['bot']
 
         if is_bot:
-            if vc := self.cache.get_bot_voice_state(event.data["guild_id"]):
+            before_bot_guild_voice_state = self.cache.get_bot_voice_state(event.data["guild_id"])
+            self.dispatch(events.VoiceStateUpdate(before_bot_guild_voice_state, after))
+            if vc := before_bot_guild_voice_state:
                 # noinspection PyProtectedMember
                 await vc._voice_state_update(after, event.data)
             return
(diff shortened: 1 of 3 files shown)


interactions/api/events/processors/voice_events.py
```diff
@@ -13,15 +13,18 @@
...
+        if is_bot:
             if vc := self.cache.get_bot_voice_state(event.data["guild_id"]):
                 # noinspection PyProtectedMember
-                await vc._voice_state_update(before, after, event.data)
+                await vc._voice_state_update(after, event.data)
+            return
+
+        # Following code is for updates to non-bot users
+        before = copy.copy(self.cache.get_voice_state(event.data["user_id"])) or None
+        self.dispatch(events.VoiceStateUpdate(before, after))
```
tequa, line 28: We don't want to dispatch this voice state update for bots -- this would affect global state for the bot across all guilds.

AstreaTSS: Um, we do though... people may listen to this update, and it's important bots get dispatched.

AstreaTSS (changes requested):
As I mentioned, it's unacceptable for bots not to be dispatched - that would both be arguably breaking and removing functionality from interactions.py. I believe this solution needs to be redone in some manner.

interactions/api/events/processors/voice_events.py
tequa: @AstreaTSS Do you understand the problem though? This dispatch event is sending a before object for event data from the user ID. The user ID is not an individual user. It's the the bot itself. So this before data will always be completely wrong if the bot's being used in multiple guilds. I'm not really sure wat the use case is for the `VoiceStateUpdate`, but maybe just sending `self.cache.get_bot_voice_state(event.data["guild_id"])` as before data if the user is a bot will work

```python
before = copy.copy(self.cache.get_voice_state(event.data["user_id"])) or None
```

AstreaTSS: Well, the before attribute being the bot... makes sense? Unless I'm missing something, that seems perfectly reasonable.

tequa: I think it might help if you look at my original issue I linked. The bot does not connect and disconnect from voice channels properly across guilds because this code is affecting and checking global voice state of the bot (`self.cache.get_voice_state(event.data["user_id"]`). There should be absolutely no global voice state for the bot, this makes no sense. There should be a voice state of the bot for each guild it's active in.

AstreaTSS: Of course, that part makes sense... but this fix is completely wrong. The bot shouldn't *not* dispatch if it doesn't find a voice state (or if it's a bot), it should dispatch the specific state it's in.

tequa: OK I'm sorry I was unaware of the purpose of the dispatch events, I haven't used those listeners before. I updated the `is_bot` block to dispatch the `VoiceUpdate` event, but now it's using the guild voice state as the before.

interactions/api/events/processors/voice_events.py
```diff
@@ -13,15 +13,20 @@
 class VoiceEvents(EventMixinTemplate):
     @Processor.define()
     async def _on_raw_voice_state_update(self, event: "RawGatewayEvent") -> None:
-        before = copy.copy(self.cache.get_voice_state(event.data["user_id"])) or None
         after = await self.cache.place_voice_state_data(event.data)
+        is_bot = event.data["member"]["user"]["bot"]
```
AstreaTSS: This should likely be checking if the user is the specific bot running, rather than any bot. `bot_voice_state` is really only meant for the running bot - yes, this means the voice state implementation is flawed as it doesn't handle other bots correctly, there's not much we can do besides for a breaking release 😅.

interactions/api/events/processors/voice_events.py
```diff
@@ -13,15 +13,20 @@
...
+            self.dispatch(events.VoiceStateUpdate(before_bot_guild_voice_state, after))
+            if vc := before_bot_guild_voice_state:
                 # noinspection PyProtectedMember
-                await vc._voice_state_update(before, after, event.data)
+                await vc._voice_state_update(after, event.data)
+            return
+
+        # Following code is for updates to non-bot users
+        before = copy.copy(self.cache.get_voice_state(event.data["user_id"])) or None
+        self.dispatch(events.VoiceStateUpdate(before, after))
```
AstreaTSS: This logic seems a bit redundant - I'm pretty sure you can do something like:

```python
if is_bot:
    before = self.cache.get_bot_voice_state(event.data["guild_id"])
    if vc := before:
        # noinspection PyProtectedMember
        await vc._voice_state_update(after, event.data)
else:
    # Following code is for updates to non-bot users
    before = copy.copy(self.cache.get_voice_state(event.data["user_id"])) or None

self.dispatch(events.VoiceStateUpdate(before, after))
```

This would also make sure the other events below it get dispatched correctly.

AstreaTSS (changes requested):
Looks mostly good after that fix - just a few things.

interactions/models/internal/active_voice_state.py
```diff
@@ -289,15 +289,13 @@ async def _voice_server_update(self, data) -> None:
...
         self,
-        before: Optional[VoiceState],
         after: Optional[VoiceState],
         data: Optional[VoiceStateData],
     ) -> None:
         """
         An internal receiver for voice server state events.
 
         Args:
-            before: The previous voice state
```
AstreaTSS: I'm sure changing around internals isn't breaking, but just in case another contributor hops it and disagrees: I see no reason that this removal is needed, per se, so it can probably be reverted 😅.

tequa: gotcha I reverted the change

AstreaTSS (dismissed):

interactions/api/events/processors/voice_events.py
tequa: Changed this to something similar to how it was done before -- checking if `self.user.id` is the same as the event user id

interactions/api/events/processors/voice_events.py
tequa: 👍

interactions/api/events/processors/voice_events.py
```diff
@@ -13,15 +13,19 @@
 class VoiceEvents(EventMixinTemplate):
     @Processor.define()
     async def _on_raw_voice_state_update(self, event: "RawGatewayEvent") -> None:
-        before = copy.copy(self.cache.get_voice_state(event.data["user_id"])) or None
         after = await self.cache.place_voice_state_data(event.data)
```
tequa: @AstreaTSS This after = `place_voice_state_data` is still sorta a bug if the user is the bot itself, it's setting global state for the bot. I didn't see any methods for like `get_voice_state_data` that fetch the voice state object without updating the global state. I'm not sure this really matters though and I didn't want to touch that code unless it's necessary.

AstreaTSS: That's actually a good point. We do already have branching if statements... it should be easy enough to define two different after behaviors, yeah?

tequa: yeah it's easy to define two different behaviors but I'm just not sure if there's already a method somewhere that translates `VoiceStateData` to `VoiceState` without actually updating the cache.

AstreaTSS: I don't think so unfortunately 😅.

Now looking at this logic, I do realize that perhaps making `before` in the bot state the thing being updated may not be a great idea. I... suppose you could do something like:

```python
# after still defined as before at the top of the function

# if statement stuff
cached_state = self.cache.get_bot_voice_state(event.data["guild_id"])
before = copy.copy(cached_state)

if cached_state:
    await cached_state._voice_state_update(before, after, event.data)
    after = cached_state
```

I... think that would work? A bit complex, but still. It doesn't solve the problem of the bot having a global state either, though I suppose you could try to pop it off the voice state cache dict?

tequa: I'm kinda confused, since `_voice_state_update` isn't even using the `before` object, wouldn't it not matter?

I made an update just to add an `update_cache` param to the `place_voice_state_data` method which hopefully should have the least side effect possible.

AstreaTSS: > I'm kinda confused, since `_voice_state_update` isn't even using the `before` object, wouldn't it not matter?

Oops, yeah, forgot about that. Still preferable to copy the state just in case.

tequa: yuh copied it in the last changed, good call

AstreaTSS (commented):
Looks good, though because this PR went through a lot, I'm going to wait for someone else's approval instead of doing it myself.

silasary:
So, I hate to do this, but... This still doesn't correctly handle a non-bot user being in two voice channels at once 😛

silasary:
It definitely fixes a bug short-term, but it doesn't fix the root bug.

AstreaTSS:
> So, I hate to do this, but... This still doesn't correctly handle a non-bot user being in two voice channels at once 😛 

Well, yes, but there's no solution to that without a breaking change.

tequa:
> So, I hate to do this, but... This still doesn't correctly handle a non-bot user being in two voice channels at once 😛

The code around non-bot users shouldn't have changed at all with this MR -- the before, after, and dispatch events should remain the same.

Also, I didn't think a non-bot user could be in 2 voice channels at once? When I join a new voice channel in the same guild or even a new guild, it leaves the old voice channel.

AstreaTSS:
Indeed, a *non-bot* user can't be in two voice channels at once... but a *bot* user can be in multiple. With the way interactions.py is designed, it only accounts for this fact for the *running bot*, not any other bot.

Again, this is unfixable without changing the cache of interactions.py, which would be breaking.

silasary:
You can be, you just need two instances of the client open (Either Desktop app and Browser/PTB/Canary, or two computers, but apparently not PC + Mobile?)

AstreaTSS:
> You can be, you just need two instances of the client open (Either Desktop app and Browser/PTB/Canary, or two computers, or PC + Mobile... etc)

Um, Discord disconnects you if you do that.

EDIT: Turns out it sometimes doesn't. Discord moment.

silasary (approved):
This is the best we'll get before 6.0, so I'm happy enough

LordOfPolls (approved):
Happy with this, but the voice state cache needs an overhaul due to oversights in its design

PR #1542, opened 2023-08-25
fix: check for guild for role processors

AstreaTSS:
## Pull Request Type

- [ ] Feature addition
- [x] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
This PR fixes role processors from erroring out when the guild cache is limited or made `NullCache`. Previously, the processors assumed the guild would always exist, when this isn't always the case.

Worth noting: many more processors use a `fetch_guild` to work around this. That... isn't a good idea here, as if the guild isn't in the cache, we likely don't care about it at all.
(Honestly, many processors could benefit from switching from `fetch_guild` to `get_guild`, but that would be breaking.)

## Changes
See the diff.

## Test Scenarios
- Disable guild cache.
- Create or delete a role.

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/api/events/processors/role_events.py
+++ b/interactions/api/events/processors/role_events.py
@@ -17,8 +17,8 @@ class RoleEvents(EventMixinTemplate):
         g_id = int(event.data.get("guild_id"))
         r_id = int(event.data["role"]["id"])
 
-        guild = self.cache.get_guild(g_id)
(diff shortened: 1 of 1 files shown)


silasary (approved):
yay, edge case scenarios!

LordOfPolls (approved):

PR #1543, opened 2023-08-25
refactor!: Auto mod refactor

mAxYoLo01:
## Pull Request Type

- [x] Feature addition
- [x] Bugfix
- [ ] Documentation update
- [x] Code refactor
- [ ] Tests improvement
- [ ] CI/CD pipeline enhancement
- [ ] Other: [Replace with a description]

## Description

Big PR to make modifications to the Auto Mod classes that currently are missing attributes, have wrong names etc..

## Changes

AutoModUpdate was triggered by the AutoModDelete event and vice versa (typo in the processors)
Added missing Triggers and Actions and added/modified some of its attributes
Renamed the AutoModLanguageType enum to KeywordPresetType (as it is called in DDocs)

## Related Issues

#1540 

## Test Scenarios

```py
@interactions.listen()
async def my_function(event: interactions.events.AutoModUpdated):
    print("UPDATED:", event)

@interactions.listen()
async def my_function2(event: interactions.events.AutoModCreated):
    print("CREATED:", event)

@interactions.listen()
async def my_function3(event: interactions.events.AutoModDeleted):
    print("DELETED:", event)
```

## Python Compatibility

- [x] I've ensured my code works on Python `3.10.x`
- [x] I've ensured my code works on Python `3.11.x`

## Checklist

- [x] I've run the `pre-commit` code linter over all edited files
- [x] I've tested my changes on supported Python versions
- [ ] I've added tests for my code, if applicable
- [x] I've updated / added  documentation, where applicable

Issue #1540, opened 2023-08-24
[BUG] Unknown trigger type 3


LawMixer:
### Library Version

Release 5.9.2

### Describe the Bug

```
Unknown trigger type for {'trigger_type': 3, 'trigger_metadata': {}, 'name': 'Block Spam Content', 'id': '1021862597634556006', 'guild_id': '864557936068395018', 'exempt_roles': [], 'exempt_channels': [], 'event_type': 1, 'enabled': True, 'creator_id': '247036279420747776', 'actions': [BlockMessage()]}
```

### Steps to Reproduce

Use anything with ``Block Spam Content`` that Discord added;

### Expected Results

Not to error

### Minimal Reproducible Code

```python
from interactions import listen, Extension, events, Embed, AuditLogEventType 

class automodDM(Extension):
    def __init__(self, client: interactions.Client):
        self.client = client
    @listen()
    async def my_function(self, event: events.AutoModExec):
        pass

        

def setup(client):
    automodDM(client)
```

### Traceback

No.

### Checklist

- [X] I have searched the open issues for duplicates.
- [X] I have shown the entire traceback, if possible.
- [X] I have removed my token from display, if visible.
- [X] I have attempted to debug this myself, and I believe this issue is with the library

PR #1546, opened 2023-08-26
fix: non-breaking auto mod fix

mAxYoLo01:
## Pull Request Type

- [x] Feature addition
- [x] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description

Similar to #1543 but since the previous PR is breaking (and won't be merged before v6.0), the bugfix needed for #1540 has been added here in a non-breaking PR. Also, some docs, typehint and arguments modified (not breaking).

## Python Compatibility

## Checklist


interactions/models/discord/auto_mod.py
```diff
@@ -97,16 +105,22 @@ def _keyword_converter(filter: str | list[str]) -> list[str]:
 class KeywordTrigger(BaseTrigger):
     """A trigger that checks if content contains words from a user defined list of keywords"""
 
-    type: AutoModTriggerType = attrs.field(
-        default=AutoModTriggerType.KEYWORD,
-        converter=AutoModTriggerType,
```
silasary: Why are you removing the type declarations from all the triggers?

mAxYoLo01: It's already defined in the BaseTrigger class. The only thing that you lose is the default value, but since type should always be passed I don't see how that's an issue. Plus, some Triggers and Actions actually did not have it re-declared.

LordOfPolls: It's intentional. Don't remove them

mAxYoLo01: > It's intentional. Don't remove them

Sure I can add them back, but then shouldn't all classes have it? Why, for example, `MentionSpamTrigger` didn't have one?

LordOfPolls: In case you haven't noticed the automod implementation is second rate. 

As for why type is explicitly defined here, it was at request

Either way, please don't remove them - feel free to draft a complete rewrite ahead for V6, but not in this PR

interactions/models/discord/auto_mod.py
```diff
@@ -369,7 +409,12 @@ def member(self) -> "Optional[Member]":
...
     AutoModTriggerType.HARMFUL_LINK: HarmfulLinkFilter,
+    AutoModTriggerType.SPAM: SpamTrigger,
     AutoModTriggerType.KEYWORD_PRESET: KeywordPresetTrigger,
     AutoModTriggerType.MENTION_SPAM: MentionSpamTrigger,
     AutoModTriggerType.MEMBER_PROFILE: MemberProfileTrigger,
 }
+
+TYPE_ALL_TRIGGER = Union[KeywordTrigger, SpamTrigger, KeywordPresetTrigger, MentionSpamTrigger, MemberProfileTrigger]
+
+TYPE_ALL_ACTION = Union[BlockMessage, AlertMessage, TimeoutUser, BlockMemberInteraction]
```
LordOfPolls, line 420: You probably want to add these to `__all__`

mAxYoLo01: Sure, but to what level should they be added? Only in this file, or all `__init__`? I also just noticed that the Actions and Triggers are not added either in there...

LordOfPolls: leave those for now, just add your new objects

LordOfPolls (approved):

LordOfPolls:
LGTM

pinging @silasary before i merge, as she was looking at this too

silasary (approved):

PR #1549, opened 2023-08-28
Minor improvements to the guides

daremon:
## Pull Request Type

- [ ] Feature addition
- [ ] Bugfix
- [x] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description

This PR includes changes to the docs and specifically the Guides. These changes were based on things I couldn't understand easily as a newcomer to interactions.py and the world of Discord bots. While the docs are quite good I feel more hand-holding is needed for first time users.

## Changes

Each commit is small and hopefully self explanatory.

## Python Compatibility

## Checklist

### DIFF
--- a/docs/src/Guides/01 Getting Started.md
+++ b/docs/src/Guides/01 Getting Started.md
@@ -56,12 +56,13 @@ Now let's get the library installed.
 Now let's get a basic bot going, for your code, you'll want something like this:
 
 ```python
-from interactions import Client, Intents, listen
+from interactions import Client, listen
 
-bot = Client(intents=Intents.DEFAULT)
-# intents are what events we want to receive from discord, `DEFAULT` is usually fine
+bot = Client(
+    token="Put your token here"
+)
 
-@listen()  # this decorator tells snek that it needs to listen for the corresponding event, and run this coroutine
+@listen()  # listen for the corresponding event, and run this coroutine
 async def on_ready():
     # This event is called when the bot is ready to respond to commands
     print("Ready")
@@ -74,7 +75,7 @@ async def on_message_create(event):
     print(f"message received: {event.message.content}")
 
 
-bot.start("Put your token here")
+bot.start()
 ```
 
 Congratulations! You now have a basic understanding of this library.
--- a/docs/src/Guides/03 Creating Commands.md
+++ b/docs/src/Guides/03 Creating Commands.md
@@ -27,7 +27,7 @@ async def my_long_command_function(ctx: SlashContext):
     await ctx.send("Hello World")
 ```
 ???+ note
-    Command names must be lowercase and can only contain `-` and `_` as special symbols and must not contain spaces.
+    Command names must be lowercase and can only contain `-` and `_`. Do not use other symbols or spaces.
 
 When testing, it is recommended to use non-global commands, as they sync instantly.
 For that, you can either define `scopes` in every command or set `debug_scope` in the bot instantiation which sets the scope automatically for all commands.
@@ -154,6 +154,9 @@ Always make sure to define all required options first, this is a Discord require
 async def my_command_function(ctx: SlashContext, integer_option: int = 5):
     await ctx.send(f"You input {integer_option}")
 ```
+
+???+ note
+    Option names must be lowercase and can only contain `-` and `_`. Do not use other symbols or spaces.
 
 For more information, please visit the API reference [here](/interactions.py/API Reference/API Reference/models/Internal/application_commands/#interactions.models.internal.application_commands.slash_option).
 
--- a/docs/src/Guides/03 Creating Commands.md
+++ b/docs/src/Guides/03 Creating Commands.md
@@ -9,6 +9,8 @@ To create an interaction, simply define an asynchronous function and use the `@s
 
 Interactions need to be responded to within 3 seconds. To do this, use `await ctx.send()`.
 If your code needs more time, don't worry. You can use `await ctx.defer()` to increase the time until you need to respond to the command to 15 minutes.
+
+Add this code to the [Basic bot](/Guides/01 Getting Started/#basic-bot).
 ```python
 from interactions import slash_command, SlashContext
 
@@ -18,22 +20,29 @@ async def my_command_function(ctx: SlashContext):
 
 @slash_command(name="my_long_command", description="My second command :)")
 async def my_long_command_function(ctx: SlashContext):
-    # need to defer it, otherwise, it fails
+    # request a time extension since we won't respond in 3 seconds
     await ctx.defer()
 
-    # do stuff for a bit
-    await asyncio.sleep(600)
+    # put your time consuming code here
+    await asyncio.sleep(10)  # we just sleep for 10 seconds here
 
     await ctx.send("Hello World")
 ```
 ???+ note
     Command names must be lowercase and can only contain `-` and `_`. Do not use other symbols or spaces.
 
-When testing, it is recommended to use non-global commands, as they sync instantly.
-For that, you can either define `scopes` in every command or set `debug_scope` in the bot instantiation which sets the scope automatically for all commands.
+All bot commands must be synced to all guilds (also called servers) that have the bot installed. This happens automatically but can take a lot of time. When developing, it is recommended to sync only to specific guilds as this happens instantly and testing is faster.
 
-You can define non-global commands by passing a list of guild ids to `scopes` in the interaction creation.
+To specify guilds you can either define `scopes` (a list of guild ids) in every command or set a `debug_scope` (just one guild id) in the bot instantiation which sets the scope automatically for all commands.
+
 ```python
+# set one guild for all commands
+bot = Client(
+    token="Put your token here",
+    debug_scope=870046872864165888
+)
+
+# or set guilds just for this command
 @slash_command(name="my_command", description="My first command :)", scopes=[870046872864165888])
 async def my_command_function(ctx: SlashContext):
     await ctx.send("Hello World")
--- a/docs/src/Guides/03 Creating Commands.md
+++ b/docs/src/Guides/03 Creating Commands.md
@@ -10,7 +10,7 @@ To create an interaction, simply define an asynchronous function and use the `@s
 Interactions need to be responded to within 3 seconds. To do this, use `await ctx.send()`.
 If your code needs more time, don't worry. You can use `await ctx.defer()` to increase the time until you need to respond to the command to 15 minutes.
 
-Add this code to the [Basic bot](/Guides/01 Getting Started/#basic-bot).
+Add this code to the [Basic bot](/interactions.py/Guides/01 Getting Started/#basic-bot).
 ```python
 from interactions import slash_command, SlashContext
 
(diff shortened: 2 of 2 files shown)


AstreaTSS:
What... exactly does this PR do? You've given us no details.

daremon:
> What... exactly does this PR do? You've given us no details.

Yeah, submitted too soon. I have now edited the description.

docs/src/Guides/01 Getting Started.md
```diff
@@ -56,12 +56,13 @@ Now let's get the library installed.
...
 
 ```python
-from interactions import Client, Intents, listen
+from interactions import Client, listen
 
-bot = Client(intents=Intents.DEFAULT)
-# intents are what events we want to receive from discord, `DEFAULT` is usually fine
+bot = Client(
+    token="Put your token here"
+)
```
AstreaTSS, line 63: Why were the Intents removed (they're a good concept to introduce right at the beginning, I feel), and why was the token moved here? The token being at the end is fine and is usually the recommended method in a lot of other guides.

docs/src/Guides/03 Creating Commands.md
```diff
@@ -9,6 +9,8 @@ To create an interaction, simply define an asynchronous function and use the `@s
 
 Interactions need to be responded to within 3 seconds. To do this, use `await ctx.send()`.
 If your code needs more time, don't worry. You can use `await ctx.defer()` to increase the time until you need to respond to the command to 15 minutes.
+
+Add this code to the [Basic bot](/interactions.py/Guides/01 Getting Started/#basic-bot).
```
AstreaTSS, line 13: I disagree with such changes as these. This isn't necessarily true - you can add this to your own code, and the guides almost never assume you're reading them in order. Plus, none of the other guides do this.

docs/src/Guides/03 Creating Commands.md
```diff
@@ -18,22 +20,29 @@ async def my_command_function(ctx: SlashContext):
 
 @slash_command(name="my_long_command", description="My second command :)")
 async def my_long_command_function(ctx: SlashContext):
-    # need to defer it, otherwise, it fails
+    # request a time extension since we won't respond in 3 seconds
```
AstreaTSS, line 23: I'd still prefer "defer" be used in this context - yes, it's a time extension, but more specifically you are deferring your response to allow up to 15 minutes to do a proper one.

docs/src/Guides/03 Creating Commands.md
```diff
@@ -18,22 +20,29 @@ async def my_command_function(ctx: SlashContext):
...
-    # do stuff for a bit
-    await asyncio.sleep(600)
+    # put your time consuming code here
+    await asyncio.sleep(10)  # we just sleep for 10 seconds here
 
     await ctx.send("Hello World")
 ```
 ???+ note
-    Command names must be lowercase and can only contain `-` and `_` as special symbols and must not contain spaces.
+    Command names must be lowercase and can only contain `-` and `_`. Do not use other symbols or spaces.
```
AstreaTSS, line 32: Minor thing, but "Do not use other symbols or spaces" feels a bit weird to say. I feel like adding ", and cannot contain other symbols or spaces" is better.

docs/src/Guides/03 Creating Commands.md
```diff
@@ -18,22 +20,29 @@ async def my_command_function(ctx: SlashContext):
...
 
-When testing, it is recommended to use non-global commands, as they sync instantly.
-For that, you can either define `scopes` in every command or set `debug_scope` in the bot instantiation which sets the scope automatically for all commands.
+To specify guilds you can either define `scopes` (a list of guild ids) in every command or set a `debug_scope` (just one guild id) in the bot instantiation which sets the scope automatically for all commands.
 
-You can define non-global commands by passing a list of guild ids to `scopes` in the interaction creation.
 ```python
+# set one guild for all commands
+bot = Client(
+    token="Put your token here",
```
AstreaTSS, line 41: This is unnecessary - using `...,` to represent "your other arguments for the init here" is fine, and also avoids the whole token debate.

docs/src/Guides/03 Creating Commands.md
```diff
@@ -284,6 +296,38 @@ async def autocomplete(self, ctx: AutocompleteContext):
...
+        await ctx.send('Message outside of embed',
+                    embed=Embed(title="Command failed",
+                                description="You don't have permission to kick members",
+                                color=Color("#ff0000")))
+```
+
+To dive a bit deeper read the docs on [SlashContent](/interactions.py/API Reference/API Reference/models/Internal/context/#interactions.models.internal.context.SlashContext) from where we can access the `.guild` model and the `.member` model.
+
+In the same fashion read the docs on the [guild model](/interactions.py/API Reference/API Reference/models/Discord/guild/#interactions.models.discord.guild.Guild) where we find the actual [`.kick`](/API Reference/API Reference/models/Discord/guild/#interactions.models.discord.guild.Guild.kick) method.
+
```
AstreaTSS, line 330: I... don't like giving examples like these because of how subjective they are. For example, I personally would have made it so that a check is used to check the authors permissions instead of adding it to the actual function, and even barring that, I would have raised an error for an error handler to get rather than sending an embed. This is a useful example of how useful the attributes in the context are... but to be frank, the API Reference already has all of this.

I think another set of eyes on this example would be a good thing because I'm very mixed about it as a whole.

docs/src/Guides/03 Creating Commands.md
```diff
@@ -284,6 +296,38 @@ async def autocomplete(self, ctx: AutocompleteContext):
...
+@slash_command(name='kick', description='Kicks a user')
+@slash_option(name="user",
+              description="The user that you want to kick",
+              required=True,
+              opt_type=interactions.OptionType.USER)
+@slash_option(name="reason",
+              description="The reason for the kick",
+              required=False,
+              opt_type=interactions.OptionType.STRING)
+async def kick(ctx: SlashContext, user, reason=None):
```
AstreaTSS, line 316: Okay, but if this example stays, this should absolutely be typehinted properly

docs/src/Guides/03 Creating Commands.md
```diff
@@ -284,6 +296,38 @@ async def autocomplete(self, ctx: AutocompleteContext):
...
 
+## An example `/kick` command
+
+Let's use some of the above features to create a real-world command that kicks a member out of the guild. Remember to add this code to the [Basic bot](/interactions.py/Guides/01 Getting Started/#basic-bot).
+
+```python
+from interactions import SlashContext, Permissions, Embed, Color
+from interactions import slash_command, slash_option
+
+@slash_command(name='kick', description='Kicks a user')
```
AstreaTSS, line 307: Why isn't this command disabled for DMs or enabled only for users with kick permissions? Both of those are easy to set up here.

docs/src/Guides/01 Getting Started.md
```diff
@@ -56,12 +56,13 @@ Now let's get the library installed.
...
 
-@listen()  # this decorator tells snek that it needs to listen for the corresponding event, and run this coroutine
+@listen()  # listen for the corresponding event, and run this coroutine
```
AstreaTSS, line 65: Now thinking about this, this should probably be changed to a `@listen(Ready)` to match up with the rest of the docs. Good catch with the comments though.

docs/src/Guides/03 Creating Commands.md
```diff
@@ -521,7 +565,7 @@ There also is `CommandCompletion` which you can overwrite too. That fires on eve
 
 If your bot is complex enough, you might find yourself wanting to use custom models in your commands.
 
-To do this, you'll want to use a string option, and define a converter. Information on how to use converters can be found [on the converter page](/Guides/08 Converters).
+To do this, you'll want to use a string option, and define a converter. Information on how to use converters can be found [on the converter page](/interactions.py/Guides/08 Converters).
```
AstreaTSS, line 568: Good catch.

AstreaTSS (changes requested):
Overall, I really like the stuff this PR tries to do, but I'm not so sure about the final result. Another set of eyes would do good here.

docs/src/Guides/01 Getting Started.md
LawMixer: i like when intents was apart of the client ngl

daremon: Thanks for the detailed feedback. I'll try to respond based on my experience as a total newcomer to all this. I admit I consider the first 3 chapters (Introduction, Creating bot, Creating Slash commands) as something to be read in order by total interactions.py newcomers. Personally I would do even more work to make these chapters a complete hand-holding experience targeting newcomers.

---

So mentioning intents so early made me think it's something important to get the basics running, so I started reading about it until I realized it is not "important". At least not important to run a basic hello world command. I feel it offers nothing at this point 😅

---

I don't understand why you can give the token both in the constructor and `.start()`. I feel it's better suited to the constructor but I also followed this -> https://github.com/interactions-py/boilerplate/blob/5.0.0/main.py#L31

docs/src/Guides/03 Creating Commands.md
daremon: If you already have "your own code" with interactions.py then I'm pretty sure you're not reading the "Hello World" example.

On the other hand when it really is your first interaction with creating a bot it's easier to forget you also need the initialization code.

docs/src/Guides/03 Creating Commands.md
daremon: Agreed, Engligh is not my native lang so I tend to prefer smaller, imperative sentences.

docs/src/Guides/01 Getting Started.md
LawMixer: ``client = Client(token=IS_PROD, intents=Intents.ALL)``

This is the proper way you do it.

AstreaTSS: For the intents: it may offer nothing *at that moment*, but having it there both explains that intents are important in interactions.py (as they *should* be) and should be set by the user whenever possible. At the very least, I see no harm to their addition - perhaps a link to the events guide would be a good idea in case someone was curious, though.

For the token... it's partially because that's how other libraries do it, and partially because they allow for making the bot in scenarios where you may not have the token yet, IE if you're making it outside of the main file and then starting the bot in the main one. The example in the repository was... well, frankly, I'm not sure why it's there, because it should be removed 😅. That was mostly for testing v5 when we were making it.

Honestly, I think my biggest complaint is that the init was changed for little to no gain - it isn't exactly unclear in its intent as is. Though moving the token there is... alright, I suppose.

docs/src/Guides/03 Creating Commands.md
daremon: The example isn't meant as a pinnacle of good code and practices. It is meant to showcase a real example of how to use the rest of the documentation. How to go from `ctx` (`SlashContext`) to the corresponding guild to the required method.

mAxYoLo01: I would personally use the `@slash_default_member_permission()` but outside of that, I feel like basic examples are cool and should be somewhere in the guides. Maybe we should totally redo the current Examples guide with examples that... make sense. Right now it's not any useful information

docs/src/Guides/03 Creating Commands.md
daremon: Should be `MISSING` instead of `None`?

mAxYoLo01: No, should be `user: Member, reason: str = None`

docs/src/Guides/03 Creating Commands.md
AstreaTSS: There's a lot of people who use these guides when they've already made a small blueprint - perhaps "add this to your existing bot file, IE the one from the previous guide" is a good compromise.

docs/src/Guides/03 Creating Commands.md
AstreaTSS: I do like the idea of revamping the examples guide.

docs/src/Guides/03 Creating Commands.md
AstreaTSS: Er `reason: str | None = None`. Typehints are a good thing to drill into newcomers early.

daremon:
@AstreaTSS I think you should decide if it makes sense to take the Guides (at least the first 3 chapters) to a more "newbie" friendly direction. I will be happy to work more on this ie breaking the overly long "Creating Slash Commands" chapter into two. One part up until the (modified if need be) `kick` example and another with the more advanced stuff.

mAxYoLo01:
Personnally, I think that the guides do need an entire overhaul. When I try to put myself in a newcomer's shoes, I feel like the guides have so many things that need to be modified. The titles don't make any sense, the "familiar" approach in some sentences (example: "Buttons are, you guessed right, buttons. Users can click them, and they can be disabled if you wish. That's all really.") doesn't feel right, and I think we could explain some more in-depth parts of how the lib or Discord API works (for example, so many people don't know how interactions work, that you need to acknowledge them).
If I weren't too lazy I would start working on that lol

AstreaTSS:
I generally agree. I'm not sure if splitting up the slash command docs is the right decision, but in general I've been working slowly on the guides for that very reason - I'd like to think my contributions at least make those parts nice.

At the same time though, we're not a library for Python newcomers - teaching a newcomer asynchronous programming is... not a good idea, if the support server can say anything about it. That being said, we can't expect users to be too familiar with the Discord API, as everyone else here has basically said.

daremon:
I think a better [Examples guide page](https://interactions-py.github.io/interactions.py/Guides/90%20Example/) would have been most helpful for me if it had some more examples.

discord.py has a great collection of examples in https://github.com/Rapptz/discord.py/tree/master/examples

This led me to discover the interactions.py examples directory: https://github.com/interactions-py/interactions.py/blob/stable/examples/ It only has one example but it's pretty good - too bad I didn't find it earlier.

I think someone should decide if examples should go to the Examples Guide page or the examples directory and add some basic, medium and more advanced examples. I would gladly try to add some basic examples.

AstreaTSS:
Hello - I think I'm going to close this PR for now as it's quickly got to become something huge (at least idea wise) and this needs a lot more discussion, especially on the Discord server. That being said, I agree that the example do need more... examples - definitely something to talk about there.

PR #1550, opened 2023-08-29
docs: Revamp guides

mAxYoLo01:
The docs generated from this PR can be seen [here](https://maxyolo01.github.io/interactions.py/).

## Pull Request Type

- [ ] Feature addition
- [ ] Bugfix
- [x] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description

This PR's goal is to enhance the current guides section of the docs. Some of its key aspects are:
- Fix typos or incorrect code examples
- Remove some of the "jokes" that could be present in the guides, especially in the titles of the sections
- Rewrite some parts of the documentation, add more meaningful examples

## Python Compatibility

## Checklist

### DIFF
--- a/docs/src/Guides/03 Creating Commands.md
+++ b/docs/src/Guides/03 Creating Commands.md
@@ -1,4 +1,4 @@
-# Creating Slash Commands
+# Slash Commands
 
 So you want to make a slash command (or interaction, as they are officially called), but don't know how to get started?
 Then this is the right place for you.
@@ -107,7 +107,7 @@ For all of these, the "group" parts are optional, allowing you to do `/base comm
     You cannot mix group subcommands and non-group subcommands into one base command - you must either use all group subcommands or normal subcommands.
 
 
-## But I Need More Options
+## Options
 
 Interactions can also have options. There are a bunch of different [types of options](/interactions.py/API Reference/API Reference/models/Internal/application_commands/#interactions.models.internal.application_commands.OptionType):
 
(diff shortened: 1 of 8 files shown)


docs/src/Guides/03 Creating Commands.md
```diff
@@ -235,9 +237,9 @@ async def my_command_function(ctx: SlashContext, integer_option: int):
 
 For more information, please visit the API reference [here](/interactions.py/API Reference/API Reference/models/Internal/application_commands/#interactions.models.internal.application_commands.SlashCommandChoice).
 
-## I Need More Than 25 Choices
+## Autocompleting Choices
```
silasary: This feels like a regression. 

Unless people already intimately know how discord works, there's an assumption that choices and autocomplete are the same thing. (UX-wise, to the end-user, they are)

"More than 25" is the most important part of this heading.  Feel free to rephrase it, but changing it to "Autocompleting choices" just makes people confused as to what choices without autocompletion does.

silasary (changes requested):

docs/src/Guides/03 Creating Commands.md
mAxYoLo01: I understand your point, however I'm not sure I agree with it. I think that the most important part of autocomplete choices is that they are dynamic, technically you're still limited to providing 25 max choices, so I'm not sure if "advertising" as allowing more than 25 choices is the best idea. Maybe others can join on this debate?

LordOfPolls: I agree with Sil, this is a regression. At the very least an alternate like the following should be considered

> Autocomplete - when you need over 25 chouces

LordOfPolls (approved):

PR #1558, opened 2023-10-03
feat: allow editing for guild role icons

example-git:
## Pull Request Type

- [X] Feature addition
- [ ] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
I have implemented handling into the role.edit method to allow for guild.icon updating, as well as adding the required converter methods into the utils.misc_utils.

## Changes
- added icon and unicode_emoji to the role.edit method. If a unicode emoji is selected it will use that value instead of an icon and vice versa.
- added bytes_to_base64_data and get_mime_type_for_image to utils.misc_utils to handle the bytes imput from the icon input into the proper formatting in the api request.

## Python Compatibility

- [X] I've ensured my code works on Python `3.11.x`

## Checklist

- [X] I've added tests for my code, if applicable
- [X] I've updated / added  documentation, where applicable


i0bs:
I don't want to give a review without understanding how this works. Are you able to provide valid client-facing code with your proposal for me?

example-git:
Here is a snippet of a listener I have that will update a role to match the icon (if one is present) on a role it's tracking. @i0bs 

I also updated the branch because I found that there was already a method in interactions.client.utils.serializer that did exactly what I added to misc_utils. Let me know if the snippet below will suffice.

```py
    async def color_check(self, user):
        roles = user.roles  # remove @everyone
        # highest order of the colour is the one that gets rendered.
        # if the highest is the default colour then the next one with a colour
        # is chosen instead
        role_color = None
        # print(user.top_role.color.value)
        for role in reversed(roles):
            if role.color.value:
                role_color = role.color
                role_name = role.name
        return role_color, role_name
        
    @listen()
    async def on_role_update(self, ctx: RoleUpdate):
        if int(ctx.after.guild.id) != int(self.GuildID):
            return
        user = await ctx.after.guild.fetch_member(self.UserID)
        roles = user.roles
        if int(ctx.after.id) not in roles:
            return
        try:
            self.bot.logger.info(f"Updating Role color to match updated user roles")
            my_role = await ctx.after.guild.fetch_role(self.RoleID)
            role_color, role_name = await self.color_check(user)
            role_icon = None
            try:
                for role in user.roles:
                    role = await ctx.guild.fetch_role(role.id)
                    if role.icon:
                        role_icon = await role.icon.fetch()
                        break
            except:
                role_icon = None
            if not role_icon:
                await my_role.edit(name=role_name, color=role_color)
            else:
                await my_role.edit(name=role_name, color=role_color, icon=role_icon)

**Review by i0bs — CHANGES_REQUESTED** [review]
Please revert type signatures for `icon` and `unicode_emoji` within the `.edit()` docstring. This is currently inconsistent with the way we already document argument signatures for methods.

Please rewrite both the documentation for `unicode_emoji` to be clearer, simpler and shorter in synopsis. Better choice of verbiage and structure lets us keep things streamlined and straight to the point.

**Comment by example-git** [comment]
> Please revert type signatures for `icon` and `unicode_emoji` within the `.edit()` docstring. This is currently inconsistent with the way we already document argument signatures for methods.
> 
> Please rewrite both the documentation for `unicode_emoji` to be clearer, simpler and shorter in synopsis. Better choice of verbiage and structure lets us keep things streamlined and straight to the point.

How do the changes in the latest commit look

**Review by i0bs — APPROVED** [review]

File: interactions/models/discord/role.py [hunk 1]
```diff
@@ -187,6 +186,8 @@ async def edit(
         color: Color | COLOR_TYPES | None = None,
         hoist: bool | None = None,
         mentionable: bool | None = None,
+        icon: bytes | None = None,
```
silasary: `to_image_data` takes UPLOADABLE_TYPE, so you may as well accept it here

silasary (commented):
Other than an overly narrow type hint, looks good to me

LordOfPolls (approved):
lgtm

Issue #1582, opened 2023-11-30
[BUG] If an autocomplete function is a method of a Client subclass, the object binding is missing.


fractal13:
### Library Version

5.11.0

### Describe the Bug

Autocomplete functions implemented as methods of a Client subclass are not called with binding to the client object. This causes the autocomplete call to raise an exception.

### Steps to Reproduce

The attached bot will demonstrate the failure when the user tries to issue the command: /ichart c4
[autocomplete-failure.zip](https://github.com/interactions-py/interactions.py/files/13518373/autocomplete-failure.zip)

### Expected Results

I would expect to have a method of the Client class be usable as an autocomplete routine.

### Minimal Reproducible Code

```python
I'm sorry it's so long, but please see the attached bot in the Steps to Reproduce.
```

### Traceback

$ ./bot6-autocomplete-fails.py
This bot is owned by @fractal13
Task exception was never retrieved
future: <Task finished name='Task-56' coro=<Client._dispatch_interaction() done, defined at /home/cgl/repos/interactions.py/interactions/client/client.py:1773> exception=TypeError("MyBot.chart_c4_integer_autocomplete() missing 1 required positional argument: 'ctx'")>
Traceback (most recent call last):
  File "/home/cgl/repos/interactions.py/interactions/client/client.py", line 1818, in _dispatch_interaction
    callback=callback(ctx),
TypeError: MyBot.chart_c4_integer_autocomplete() missing 1 required positional argument: 'ctx'

### Checklist

- [X] I have searched the open issues for duplicates.
- [X] I have shown the entire traceback, if possible.
- [X] I have removed my token from display, if visible.
- [X] I have attempted to debug this myself, and I believe this issue is with the library

### Additional Information

I have a PR that fixes this issue. There may be a better way to fix it. Please feel free to use your own solution.

LordOfPolls:
Please don't use zips for minimal reproducibles - it's incredible bad practice and a security concern for the person downloading it. 

Please just share the relevant code - if you can't do that, just the relevant file

fractal13:
Thanks for letting me know. The UI won't let me upload a .py file. So, here's the file contents:

#!/usr/bin/env python3
from interactions import Client, Intents, listen
from interactions import slash_command, SlashContext, SlashCommand
from interactions import OptionType, slash_option, SlashCommandChoice
from interactions import AutocompleteContext
import os

class MyBot(Client):

    def __init__(self, **kwargs):
        super().__init__(**kwargs)
[...]

PR #1583, opened 2023-11-30
fix: consider binding when running autocomplete functions

fractal13:
## Pull Request Type

- [ ] Feature addition
- [ X] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
Autocomplete functions implemented as methods of a Client subclass are not called with binding to the client object. This causes the autocomplete call to raise an exception. This PR fixes the issue.

## Changes
- Call autocomplete methods with object binding.

## Related Issues
#1582

## Test Scenarios
Issue #1582 has a script that demonstrates the failure before this patch, and success after this patch.

## Python Compatibility

- [X ] I've ensured my code works on Python `3.10.x`

## Checklist

- [ X] I've run the `pre-commit` code linter over all edited files
- [ X] I've tested my changes on supported Python versions

There were no autocomplete tests prior to this, and I did not add any to the tests/ directory.
No documentation should need to be updated.

### DIFF
--- a/interactions/client/client.py
+++ b/interactions/client/client.py
@@ -1808,9 +1808,7 @@ class Client(
                 if auto_opt := getattr(ctx, "focussed_option", None):
                     if autocomplete := ctx.command.autocomplete_callbacks.get(str(auto_opt.name)):
                         if ctx.command.has_binding:
-                            callback = CallbackObject()
-                            callback.callback = autocomplete
-                            callback._binding = ctx.command._binding
+                            callback = functools.partial(ctx.command.call_with_binding, autocomplete)
                         else:
                             callback = autocomplete
                     elif autocomplete := self._global_autocompletes.get(str(auto_opt.name)):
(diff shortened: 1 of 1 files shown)


fractal13:
@interactions-py/core

interactions/client/client.py
```diff
@@ -1807,7 +1807,12 @@ async def _dispatch_interaction(self, event: RawGatewayEvent) -> None:  # noqa:
 
                 if auto_opt := getattr(ctx, "focussed_option", None):
                     if autocomplete := ctx.command.autocomplete_callbacks.get(str(auto_opt.name)):
-                        callback = autocomplete
+                        if ctx.command.has_binding:
+                            callback = CallbackObject()
+                            callback.callback = autocomplete
+                            callback._binding = ctx.command._binding
+                        else:
+                            callback = autocomplete
```
AstreaTSS: This is more overhead than it needs to. Instead of making a new callback object, using `ctx.command.call_with_binding` (with `functools.partial`) is likely much better.

AstreaTSS (changes requested):

interactions/client/client.py
fractal13: Great. I'll try to figure out that syntax and resubmit.

silasary (approved):

eightween (approved):

PR #1602, opened 2024-02-12
feat: ability to define which shards connect

Wolfhound905:
## Pull Request Type

- [x] Feature addition
- [ ] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
Having one AutoShardedClient can be risky in the case of a fatal error in a users code. Leading for the entire bot to crash. With this feature, a user can call a microservice or other to determine what shards the AutoShardedClient should be running. 

## Changes
- added `shard_ids` to AutoShardedClient classs

## Test Scenarios
```py
bot = AutoShardedClient(
    fetch_members=False,
    delete_unused_application_cmds=False,
    sync_interactions=False,
    total_shards=112,
    shard_ids=[12, 13, 14, 15, 16],
    basic_logging=True,
    log_level=logging.DEBUG,
)


@listen(Startup, delay_until_ready=True)
async def on_startup(event: Startup):
    print("Bot is ready")
    print(f"Logged in as {bot.user}")
    print(f"Bot ID: {bot.user.id}")
    print(f"Guilds: {len(bot.guilds)}")

```

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/client/auto_shard_client.py
+++ b/interactions/client/auto_shard_client.py
@@ -35,7 +35,7 @@ class AutoShardedClient(Client):
         self.auto_sharding = "total_shards" not in kwargs
         super().__init__(*args, **kwargs)
 
-        self.only_shards: Optional[List[int]] = kwargs.get("only_shards", None)
+        self.shard_ids: Optional[List[int]] = kwargs.get("shard_ids", None)
 
         self._connection_state = None
 
@@ -247,8 +247,8 @@ class AutoShardedClient(Client):
 
         self.logger.debug(f"Starting bot with {self.total_shards} shard{'s' if self.total_shards != 1 else ''}")
 
-        if self.only_shards:
-            self._connection_states = [ConnectionState(self, self.intents, shard_id) for shard_id in self.only_shards]
+        if self.shard_ids:
+            self._connection_states = [ConnectionState(self, self.intents, shard_id) for shard_id in self.shard_ids]
         else:
             self._connection_states = [
                 ConnectionState(self, self.intents, shard_id) for shard_id in range(self.total_shards)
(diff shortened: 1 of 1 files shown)


interactions/client/auto_shard_client.py
```diff
@@ -35,6 +35,8 @@ def __init__(self, *args, **kwargs) -> None:
         self.auto_sharding = "total_shards" not in kwargs
         super().__init__(*args, **kwargs)
 
+        self.only_shards: Optional[List[int]] = kwargs.get("only_shards", None)
```
AstreaTSS: Why not use `Optional[list[int]]` and avoid an import?

AstreaTSS: Not sure how I feel about the naming either - it's rather confusing. Why not just name it `shard_ids`?

AstreaTSS (changes requested):

interactions/client/auto_shard_client.py
Wolfhound905: > Not sure how I feel about the naming either - it's rather confusing. Why not just name it `shard_ids`?

I didn't want anyone to confuse it with "shard_id." But either way is fine and I can change it if you feel only_shards is confusing.

Wolfhound905: > Why not use `Optional[list[int]]` and avoid an import?

This is nitpicking, we use "List" hundreds of times in other files, I believe that if I am using the Typing library for Optional, I might as well use it for List as well.

AstreaTSS: > I didn't want anyone to confuse it with "shard_id." But either way is fine and I can change it if you feel only_shards is confusing.

`only_shards` feels even more confusing than `shard_ids` honestly. Perhaps we should look into renaming attributes in `AutoShardedClient` in general, though that's for a different PR.

AstreaTSS (approved):

PR #1605, opened 2024-02-20
Update voice recording example

jasondilworth56:
## Pull Request Type

- [ ] Feature addition
- [ ] Bugfix
- [x] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
The existing example does not work due to missing deferral and missing message body, this fixes both.

## Changes
- Added a called to `ctx.defer()` at beginning of example
- Added message body to output from example

## Python Compatibility

## Checklist

### DIFF
--- a/docs/src/Guides/23 Voice.md
+++ b/docs/src/Guides/23 Voice.md
@@ -102,13 +102,14 @@ import interactions
 
 @interactions.slash_command("record", "record some audio")
 async def record(ctx: interactions.SlashContext):
+    await ctx.defer()
     voice_state = await ctx.author.voice.channel.connect()
 
     # Start recording
     await voice_state.start_recording()
     await asyncio.sleep(10)
     await voice_state.stop_recording()
-    await ctx.send(files=[interactions.File(file, file_name="user_id.mp3") for user_id, file in voice_state.recorder.output.items()])
+    await ctx.send("Here are your recordings", files=[interactions.File(file, file_name="user_id.mp3") for user_id, file in voice_state.recorder.output.items()])
 ```
 This code will connect to the author's voice channel, start recording, wait 10 seconds, stop recording, and send a file for each user that was recorded.
 


docs/src/Guides/23 Voice.md
```diff
@@ -102,13 +102,14 @@ import interactions
...
 async def record(ctx: interactions.SlashContext):
+    await ctx.defer()
     voice_state = await ctx.author.voice.channel.connect()
 
     # Start recording
     await voice_state.start_recording()
     await asyncio.sleep(10)
     await voice_state.stop_recording()
-    await ctx.send(files=[interactions.File(file, file_name="user_id.mp3") for user_id, file in voice_state.recorder.output.items()])
+    await ctx.send("Here are your recordings", files=[interactions.File(file, file_name="user_id.mp3") for user_id, file in voice_state.recorder.output.items()])
```
AstreaTSS, line 112: The content here shouldn't be needed. What error are you getting here?

docs/src/Guides/23 Voice.md
```diff
@@ -102,13 +102,14 @@ import interactions
 
 @interactions.slash_command("record", "record some audio")
 async def record(ctx: interactions.SlashContext):
+    await ctx.defer()
```
AstreaTSS, line 105: Good catch.

docs/src/Guides/23 Voice.md
jasondilworth56: I cannot test again until a bit later, but there was an error about the message not being empty.

eightween:
Putting on hold until requests are met and satisfied

docs/src/Guides/23 Voice.md
TyrantKingBen: It's because you didn't actually send/capture any audio. You need to talk and should visually see that your microphone is activating. Without any text or file attachments, the message was empty, hence the error.

```python
files=[interactions.File(file, file_name=f"{user_id}.mp3") for user_id, file in voice_state.recorder.output.items()]
if len(files) > 0:
    await ctx.send(files=files)
else:
    await ctx.send("No audio was recorded.")
```

AstreaTSS:
Closing, PR seems to be dead.

PR #1661, opened 2024-04-19
feat: option to defer without raising errors

mifuyutsuki:
## Pull Request Type

- [x] Feature addition
- [ ] Bugfix
- [x] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description

Add a bool kwarg `suppress_error` (default `False`) to `defer()` (often `ctx.defer()`) in different instances of `InteractionContext`. If enabled, deferring an interaction response again will not raise an error, which is the current behavior of `ctx.defer()`.

One example use case of this feature, which I use in a bot of mine, is using `auto_defer(time_until_defer=...)` and `ctx.defer()` together, making the command defer after either a certain point or a set time, whichever is reached first.

### Technical Description
This feature moves/renames the current `defer()` to `_defer()` (note the underscore). The new `defer()` is a wrapper for `_defer()`, which would suppress relevant errors (`AlreadyDeferred`, `AlreadyResponded`, and `HTTPException`) using `contextlib.suppress()`. A similar method of error suppression is already used by `AutoDefer`.

## Changes

* Add bool kwarg `suppress_error` to `defer()`
  * Default is `False` in line with current behavior
* Add error suppression logic in `defer()`
  * Affects `InteractionContext`, `ComponentContext`, `ContextMenuContext`, `ModalContext`, and `HybridContext`
* Bonus: add docstring note in `defer()` regarding the method overriding ephemeral response settings

## Test Scenarios

Use the following callback coroutine to an interaction command (slash command, component, context menu, modal, hybrid slash command). Tested on all five.
```python
async def test_cmd(ctx):
  await ctx.defer()
  try:
    await ctx.defer(suppress_error=True)
  except Exception as e:
    await ctx.send(f"Error raised: {type(e)}.")
  else:
    await ctx.send(f"No error.")
```
If `suppress_error=False` or unspecified: Command responds with "Error raised" for `AlreadyDeferred`.
If `suppress_error=True`: Command responds with "No error".

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/ext/hybrid_commands/context.py
+++ b/interactions/ext/hybrid_commands/context.py
@@ -188,6 +188,7 @@ class HybridContext(BaseContext, SendMixin):
 
         Args:
             ephemeral: Should the response be ephemeral? Only applies to responses for interactions.
+            suppress_error: Should errors on deferring be suppressed than raised.
 
         """
         if self._slash_ctx:
--- a/interactions/ext/hybrid_commands/context.py
+++ b/interactions/ext/hybrid_commands/context.py
@@ -183,9 +183,6 @@ class HybridContext(BaseContext, SendMixin):
             If using this method, whether the interaction response is ephemeral or not will be determined by this
             method regardless of ephemeral settings in send().
 
-            If used alongside auto defer, you may want to use `suppress_error` as auto defer may defer first before
-            this method is called manually.
-
         Args:
             ephemeral: Should the response be ephemeral? Only applies to responses for interactions.
             suppress_error: Should errors on deferring be suppressed than raised.
--- a/interactions/ext/hybrid_commands/context.py
+++ b/interactions/ext/hybrid_commands/context.py
@@ -180,8 +180,10 @@ class HybridContext(BaseContext, SendMixin):
         Either defers the response (if used in an interaction) or triggers a typing indicator for 10 seconds (if used for messages).
 
         ???+ note "Interaction Note"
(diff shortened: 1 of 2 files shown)


interactions/ext/hybrid_commands/context.py
```diff
@@ -175,16 +175,23 @@ def typing(self) -> Typing | DeferTyping:
...
         Args:
             ephemeral: Should the response be ephemeral? Only applies to responses for interactions.
+            suppress_error: Should errors on deferring be suppressed than raised.
 
         """
         if self._slash_ctx:
-            await self._slash_ctx.defer(ephemeral=ephemeral)
+            await self._slash_ctx.defer(ephemeral=ephemeral, suppress_error=suppress_error)
         else:
             await self.channel.trigger_typing()
```
AstreaTSS, line 201: To be consistent, you probably should also suppress errors for the `trigger_typing` variant.

AstreaTSS (changes requested):

interactions/ext/hybrid_commands/context.py
mifuyutsuki: I didn't think of adding one because `trigger_typing()` doesn't have the sort of errors expected by an extra call to `defer()`.

That said, I added suppression (on only `HTTPException`) near that snippet of code.

AstreaTSS (approved):

silasary (approved):

PR #1668, opened 2024-04-27
fix: Don't throw confusing error when using unsuppressed autocomplete in Hybrid Commands

silasary:
## Pull Request Type

- [ ] Feature addition
- [x] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
I wasted far too much time trying to figure out why my hybrid commands didn't have any options defined, and ignored the "Autocomplete doesn't work on text commands" message because I wasn't even getting as far as having options to autocomplete.

It was only when I dug far too much into the code that I realized that the "warning" there was effectively removing all parameters from my command (in both Slash and Prefixed modes), that I figured out that I needed to pass `_silence_autocomplete_errors=True`, something that A) is a private parameter, and B) implies that it hides the message but otherwise doesn't change behaviour. 

This PR makes the combination of `autocomplete=True, _silence_autocomplete_errors=False` behave the way I expected it to.

## Changes
* Changed `throw "Autocomplete is unsupported in hybrid commands."` into a warning.

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/ext/hybrid_commands/hybrid_slash.py
+++ b/interactions/ext/hybrid_commands/hybrid_slash.py
@@ -31,6 +31,7 @@ from interactions.client.const import AsyncCallable, GLOBAL_SCOPE
 from interactions.client.utils.serializer import no_export_meta
 from interactions.client.utils.misc_utils import maybe_coroutine, get_object_name
 from interactions.client.errors import BadArgument
+from interactions.client import get_logger
 from interactions.ext.prefixed_commands import PrefixedCommand, PrefixedContext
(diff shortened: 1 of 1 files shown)


AstreaTSS (commented):
We've talked about this in the contributor chat, but to give a recap: while there is a bug involved with this error because of hybrid command processing happening in an listener that cannot completely stop the bot, and also agree the naming and wording of the variable isn't great, I disagree with this solution.

That being said, I'm not going to stop this PR from being merged (hence commenting instead of requesting changes) - I'm just stating my opinion here. I also recognize alternative solutions get hacky pretty fast, as listeners do suppress errors by their very nature.

PR #1672, opened 2024-04-30
fix/refactor: add missing imports and reorganize some alls

AstreaTSS:
## Pull Request Type

- [ ] Feature addition
- [x] Bugfix
- [ ] Documentation update
- [x] Code refactor
- [ ] Tests improvement

## Description
I've recently developed a tool that goes in and makes sure that each `__init__.py` matches up with the `__all__`s in files/folders in the same directory as it. For example, if there was a statement like `from .channel import GuildText`, the program will go in and make sure the `__all__` of `channel` is `("GuildText",)`, and fixes the import if the `__all__` doesn't match up. It also detects if the `__init__` isn't importing a file with a `__all__` in it and adds it to the import.

This tool is *far* from ready, but I've gotten it to a state where I could run a run on the interactions.py library. Wouldn't you know it, we *are* missing some imports, so I made this PR to add them back in. This PR also fixes some other issues I was running into involving the codebase, like `GuildMedia` not being properly exposed in `channel.py`.

As a (hopefully good) consequence of this program, `__all__` for `__init__.py`s that have been affected have been put in alphabetical order.

## Changes
- Add `GuildMedia` to `__all__` in `channel.py` for models.
- Export `CronTrigger`.
- Export `LocalisedField` and `LocalizedField`.
- Export `ClientObject` and `DiscordObject`. You may argue against it, but I'd argue that there's no good reason to *not* export them, and they can have their uses.
- Oh right, this program also picked up `deserialize_app_cmds`, interestingly enough.
- Do an alphabetical sort for all `__all__`s that have been adjusted.

## Python Compatibility

## Checklist


LawMixer (approved):

LordOfPolls:
Your code sounds cool, but out of interest how are you validating for objects that should _not_ be exposed to the top level API of a library? Its not always desirable to expose everything under the root namespace. 

Or is this just a expose-everything situation?

AstreaTSS:
> Your code sounds cool, but out of interest how are you validating for objects that should _not_ be exposed to the top level API of a library? Its not always desirable to expose everything under the root namespace.
> 
> Or is this just a expose-everything situation?

Sorry, couldn't look at the code I made for a while 😅. Anyways, the code goes in and exposes everything in a file's `__all__`, so it's not the most robust thing in the world. Hence why the PR here was limited to only running in places I deemed "safe", and even then, in retrospect, it may have been too zealous.

I'm not sure if there's a way of making some objects not exposed to the top level without adding it to some ignore option or the like to the script.

AstreaTSS:
For transparency's sake, since it's been some months and this PR would have to be recreated if we want to continue with it, here's the script that generated everything: https://gist.github.com/AstreaTSS/9032753c1ae1b8ece431a87d31ca166f

LordOfPolls:
per @AstreaTSS's last. suggest re-creating if this is required. However this is going to cause some nasty conflicts with #1774

PR #1678, opened 2024-05-06
fix: hook onto add_command to propagate errors correctly

AstreaTSS:
## Pull Request Type

- [ ] Feature addition
- [x] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
Errors generated when adding prefixed and hybrid commands were not being propagated as they should, as due to being a separated extension, they would error out like a listener instead of like `add_command`. This caused some fun situations, like the one noted in #1668.

This PR fixes that by adding a way of adding "hooks" onto `add_command` directly, and then making prefixed and hybrid commands use that hook for its register command logic.

## Changes
- Add `Client._add_command_hook`, which is a list of functions to run during `add_command`.
- Make prefixed commands and hybrid commands use the hook.
  - This required modifying the register command logic a bit to make the function not async. This would be *technically* breaking for hybrid commands due to me not using an underscore, so the old function was kept as a wrapper for the new one.

## Test Scenarios
```python
@hybrid_slash_command()
@slash_option(
    "test", "The test option", OptionType.STRING, required=True, autocomplete=True
)
async def test(ctx: HybridContext, test: str):
    await ctx.send(f"Test: {test}")
```

```python
@prefixed_command()
async def test(ctx: PrefixedContext):
    await ctx.send("Hello!")

@prefixed_command(name="test")
async def test2(ctx: PrefixedContext):
    await ctx.send("Hello!")
```

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/ext/prefixed_commands/manager.py
+++ b/interactions/ext/prefixed_commands/manager.py
@@ -9,7 +9,6 @@ from interactions.api.events.discord import MessageCreate
 from interactions.api.events.internal import (
     CommandError,
     CommandCompletion,
-    CallbackAdded,
     ExtensionUnload,
 )
 from interactions.client.client import Client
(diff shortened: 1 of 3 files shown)


interactions/ext/prefixed_commands/manager.py
```diff
@@ -229,17 +228,13 @@ def remove_command(self, name: str, delete_parent_if_empty: bool = False) -> Opt
...
-    @listen("callback_added")
-    async def _register_command(self, event: CallbackAdded) -> None:
+    def _register_command(self, callback: Callable) -> None:
         """Registers a prefixed command, if there is one given."""
-        if not isinstance(event.callback, PrefixedCommand):
+        if not isinstance(callback, PrefixedCommand):
             return
 
-        cmd = event.callback
-        cmd.extension = event.extension
```
AstreaTSS, line 239: Note to any reviewer: this line was redundant, since extensions do this themselves before even adding the command.

silasary (approved):

Issue #1681, opened 2024-05-12
[FEAT] Add deactivate_buttons() function to Paginators


Donbur4156:
### Problem Description

To reduce lines of code for switching from buttons to select menu I would recommend to implement a function in Paginators Class.

before:
```py
paginator.show_select_menu = True
paginator.show_back_button = False
paginator.show_first_button = False
paginator.show_next_button = False
paginator.show_last_button = False
```
after:
```py
paginator.show_select_menu = True
paginator.deactivate_buttons()
```

### Proposed Solution

add following function to Paginators class:

```py
class Paginator:
...
    def deactivate_buttons(self) -> None:
        """
        deactivate all control buttons for `First`, `Back`, `Next`, `Last`.
        """
        self.show_first_button = False
        self.show_back_button = False
        self.show_next_button = False
        self.show_last_button = False
...
```

### Additional Information

PR is ready and will add asap

### Code of Conduct

- [X] I agree to follow the contribution requirements.

PR #1682, opened 2024-05-12
add deactivate_buttons() to Paginators

Donbur4156:
## Pull Request Type

- [X] Feature addition
- [ ] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
add function to deactivate Control Buttons in a Paginator. To reduce 5 Lines of Code to 2 to switch from Buttons to select.

## Changes

- add function "deactivate_buttons()" to Paginator Class

## Related Issues
#1681

## Test Scenarios
1. create Paginator
2. activate Select Menu
3. deactivate Buttons via function
4. send Paginator

## Python Compatibility

not testet with 3.11, but changes should not break this.

## Checklist

### DIFF
--- a/interactions/ext/paginators.py
+++ b/interactions/ext/paginators.py
@@ -259,6 +259,15 @@ class Paginator:
             pages.append(Page(page, prefix=prefix, suffix=suffix))
         return cls(client, pages=pages, timeout_interval=timeout)
 
+    def deactivate_buttons(self) -> None:
+        """
+        deactivate all control buttons for `First`, `Back`, `Next`, `Last`.
+        """
+        self.show_first_button = False
+        self.show_back_button = False
+        self.show_next_button = False
+        self.show_last_button = False
+
     def create_components(self, disable: bool = False) -> List[ActionRow]:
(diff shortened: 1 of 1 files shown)


AstreaTSS:
I'm... not actually sure if this is really useful, honestly. This seems like a niche usecase that isn't done terribly often, and it's only a few lines of code.

mifuyutsuki:
If the purpose of this feature is to change a few default attributes... Yeah, I don't think it warrants creating a whole method. And the method name `deactivate_buttons()` itself appears like it's meant to disable the buttons of an already sent paginator, which is the purpose of `Paginator.stop()`.

If you use that "show select menu only" pattern often in your bot code, you might have better luck creating a subclass of `Paginator` with the relevant attributes overridden. Otherwise, I would be perfectly fine with the extra few lines for explicit readability.

Donbur4156:
Ok, it was just an idea to simplify the bot code.
If you don't want to implement this I would be fine with it and you can close PR and Feature request.

PR #1699, opened 2024-06-12
feat: make context have generic client types

AstreaTSS:
## Pull Request Type

- [x] Feature addition
- [ ] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
This PR makes the contexts a generic type holding `ClientT`, a `TypeVar` that represents clients and subclasses. Taking advantage of [PEP 696](https://peps.python.org/pep-0696/), this PR ensures a grand total of 0 typings are broken by using `default` (thanks to `typing_extensions`).

This allows for people to type out their command functions in a pretty clean way if they make a subclass of `Client` - previously, you would have to make your own `Context` class that replaced the `Client` references with your own, but now it's as easy as the test scenario.

This doesn't cover every possible thing that could use `ClientT`, and especially not everything that could use `TypeVar(default=X)`, but it's a start

## Changes
- Introduce `ClientT` in `const`. To get around circular imports, during runtime, all `ClientT` is is `ClientT = TypeVar("ClientT")`, but when type checking, it's far more accurate.
- Use `ClientT` in the `Context`s as a generic.
- Adjust `Client`'s custom contexts to use the new generic property.

## Test Scenarios
```python
import interactions as ipy

class MyBotClient(ipy.Client):
    def say_hi(self) -> None:
        print("Hello!")

@ipy.slash_command()
async def test(ctx: ipy.SlashContext[MyBotClient]):
    ctx.bot.say_hi()  # typehints perfectly, instead of being unknown
```

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/models/internal/context.py
+++ b/interactions/models/internal/context.py
@@ -9,7 +9,7 @@ import discord_typings
 from aiohttp import FormData
 
 from interactions.client import const
-from interactions.client.const import get_logger, MISSING
+from interactions.client.const import get_logger, MISSING, ClientT
 from interactions.models.discord.components import BaseComponent
 from interactions.models.discord.file import UPLOADABLE_TYPE
 from interactions.models.discord.sticker import Sticker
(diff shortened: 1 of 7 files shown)


interactions/models/internal/context.py
```diff
@@ -172,8 +169,10 @@ class BaseContext(metaclass=abc.ABCMeta):
     guild_id: typing.Optional[Snowflake]
     """The id of the guild this context was invoked in, if any."""
 
-    def __init__(self, client: "interactions.Client") -> None:
-        self.client = client
+    def __init__(self, client: ClientT) -> None:
+        self.client: ClientT = client
+        """The client that created this context."""
```
AstreaTSS, line 174: If you're wondering about this - I had a hard time convincing typehinting that `self.client` wasn't `Unknown` without resorting to this. No idea why.

silasary (approved):

PR #1716, opened 2024-07-13
fix: command lookup error when scope is cached

sondevra:
Fixed lookup to use .get() on the dictionary to prevent large log of error in the circumstance of a cache command showing on a changed scope

## Pull Request Type

- [ ] Feature addition
- [x] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
Fixed lookup to use .get() on the dictionary to prevent large log of error in the circumstance of a cache command showing on a changed scope.

What happens is when no commands have been loaded to a Discord guild's scope, but they are cached, if a user tries to request one of these cached commands, an error occurs because of the way the lookup was handled (index notation)

## Changes

- Changed index notation to `.get()` on the dictionary.

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/models/internal/context.py
+++ b/interactions/models/internal/context.py
@@ -346,7 +346,7 @@ class BaseInteractionContext(BaseContext[ClientT]):
 
     @property
     def command(self) -> InteractionCommand:
-        return self.client._interaction_lookup[self._command_name]
+        return self.client._interaction_lookup.get(self._command_name)
 
     @property
     def expires_at(self) -> Timestamp:
--- a/interactions/models/internal/context.py
+++ b/interactions/models/internal/context.py
@@ -345,7 +345,7 @@ class BaseInteractionContext(BaseContext[ClientT]):
         return Permissions(0)
 
     @property
-    def command(self) -> InteractionCommand:
+    def command(self) -> typing.Optional[InteractionCommand]:
         return self.client._interaction_lookup.get(self._command_name)
 
     @property


sondevra:
@AstreaTSS need your help adding reviewers.

AstreaTSS:
Reviewers don't need to be manually added - the contributors of interactions.py will see this and review whenever they can on their own.

interactions/models/internal/context.py
```diff
@@ -346,7 +346,7 @@ def author_permissions(self) -> Permissions:
 
     @property
     def command(self) -> InteractionCommand:
-        return self.client._interaction_lookup[self._command_name]
+        return self.client._interaction_lookup.get(self._command_name)
```
AstreaTSS, line 349: This does make this breaking, right? I'm going to let another contributor look over this to give their own judgment, but this definitely changes how people should interface with the property.

sondevra: Not necessarily. Original functionality, if the command doesn't exist it will throw a `KeyError` whilst .get() will just return None.

The logic is already in place to handle as if it's supposed to return None.

See:

https://github.com/interactions-py/interactions.py/blob/2a896b7a34a42bf4abe63b6a7b4af17a1a60b722/interactions/client/client.py#L1913-L1951

```python
if ctx.command:
                self.logger.debug(f"{ctx.command_id}::{ctx.command.name} should be called")
```
but this never is not true because it will return a KeyError before it reaches there.

AstreaTSS: I saw that too, but that doesn't mean changing the property still wouldn't be breaking. It just points out how this is not how it was meant to be designed, but I know I've been told before that fixes to things like that would still be breaking.

Again, I'm going to let another contributor decide on a judgment since I don't have the best judgment on what's breaking or not.

syntaxsdev: You are right in the sense that this change could be breaking. I'm sure people have also written counter logic from the errors that pertain from this issue. The thing about this issue is that it's impossible to try-catch this error since it comes from the an independent callback - to my knowledge. However, there does need to be a proper resolution.

AstreaTSS: I'm a bit confused about not being able to try catch the error - you should absolutely be able to. If you mean how the command dispatcher will error... well, yes, you can't really error handle things like that from a bot developer POV.

That being said, I'm pretty sure you could easily modify the library itself to make the interaction command dispatcher use a try catch itself.

silasary: If I'm understanding it right, this only comes up when you're running multiple clients on the same token?

If so, I'm happy to say the current situation is a bug related to undefined behaviour, and therefore non-breaking.

AstreaTSS: Not necessarily - it can happen if your commands are out of sync in general, which... is also undefined, granted.

I suppose the only remaining question then is if the typehint needs to be changed.

silasary: Yeah, definitely needs to be marked as Optional

sondevra: Thanks for feedback. If you all have anymore feedback please let me know.

AstreaTSS (approved):

silasary (approved):

PR #1728, opened 2024-08-05
fix: properly parse aliases for base commands

AstreaTSS:
## Pull Request Type

- [ ] Feature addition
- [x] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
Previously, aliases like the below wouldn't be properly registered for the prefixed command variant of hybrid commands:
```python
message = HybridSlashCommand(
    name="message",
    description="Hosts public-facing messaging commands.",
    dm_permission=False,
    aliases=["msg"],
)
```

In this case, trying to use `<prefix>msg` would not make the bot respond. The hybrid manager was just ignoring any commands without a callback or that were only meant as a jumping point for subcommands, meaning it would never read the aliases from them.

For the first part, this PR properly parses these types of commands (and makes it so it doesn't add improper aliases for base commands). However, this introduced a new problem - the command hooks were run after an empty callback check, which means that the above example would *still* not properly get parsed. That was fixed by moving the hook above the callback check.

## Changes
- Parse and use aliases from "empty" base commands.
- Don't add subcommand aliases as aliases to base command.
- Move add command hook logic above the callback check.

## Test Scenarios
See example above.

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/client/client.py
+++ b/interactions/client/client.py
@@ -1489,6 +1489,9 @@ class Client(
         elif not isinstance(func, BaseCommand):
             raise TypeError("Invalid command type")
 
+        for hook in self._add_command_hook:
+            hook(func)
+
         if not func.callback:
             # for group = SlashCommand(...) usage
             return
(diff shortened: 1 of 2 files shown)


silasary (approved):

LawMixer (approved):
Looks & feels good to my dev discord bot!

PR #1730, opened 2024-08-08
feat: Add poll result embed type

retr0-init:
## Pull Request Type

- [ ] Feature addition
- [x] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description

The new poll feature results in a new `EmbedType`, `poll_result`. Missing this will lead to error to be:
```
ValueError: 'poll_result' is not a valid EmbedType
```

## Changes

Add the `poll_result` to the `EmbedType` in the `enum.py`

## Related Issues

N/A

## Test Scenarios

Create a poll and end the poll.
Try to use the following code in the channel where the poll was created.
```python
@interactions.slash_command(
    "test", description="test command", scopes=[DEV_GUILD] if DEV_GUILD else None
)
async def test_cmd(self, ctx: interactions.SlashContext):
    """Register as an extension command"""
    async for msg in ctx.channel.history(10):
        ...
```
It should not generate any error

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/models/discord/enums.py
+++ b/interactions/models/discord/enums.py
@@ -448,6 +448,7 @@ class EmbedType(Enum):
     LINK = "link"
     AUTOMOD_MESSAGE = "auto_moderation_message"
     AUTOMOD_NOTIFICATION = "auto_moderation_notification"
+    POLL_RESULT = "poll_result"
 
 
 class MessageActivityType(CursedIntEnum):


retr0-init:
This yields another message:
```
Class `MessageType` received an invalid and unexpected value `46`, a new enum item will be created to represent this value. Please update interactions.py or report this issue on GitHub - https://github.com/interactions-py/interactions.py/issues
```
I think this is the `poll_result` message type. However, it is not shown on the official documentation on https://discord.com/developers/docs/resources/message#message-object-message-types

Do you think I need to create a message type in the `MessageType` enum like `POLL_RESULT` for the value of `46`?

AstreaTSS:
We usually don't add undocumented features.

retr0-init:
OK. I will document this into an issue. And if there is any update about this on the official documentation, we can proceed with the addition

AstreaTSS (approved):
Approving this PR because of how prevalent errors are without it, but it's worth noting that [the PR for this embed type](https://github.com/discord/discord-api-docs/pull/7050) notes a lot more fields for it. These fields aren't necessary to add right now as interactions.py will ignore them, but they should be added at some point.

PR #1732, opened 2024-08-13
Add `name` as an alias for `display_name`

AutonomousCat:
## Pull Request Type
- Feature addition

## Description
Purpose of the PR is to give a shorter version of `display_name`. I was considering other options first, like "dname", but was thinking what of this name actually is. It's just someone's name and `name` isn't used by anything else. I can't think of a better purpose for it, and both "name" and "username" together makes sense to me.

## Changes
- Added `name` property to `interactions/models/discord/user.py` (and `user.pyi`)


AstreaTSS:
I'd argue `name` is extremely vague if you didn't know it was an alias of `display_name` - I know many people would just consider it an alias of `username` or `tag` instead.

Plus, I'm against added aliases like this in general. There's a purpose to aliases (largely when names can be spelt differently across different variants of English or for syntax matching discord.py), but short aliases like this feel unnecessary when you can do `name = user.display_name` if you really needed a short name.

I'm against this PR, but I'll let other contributors decide.

silasary:
> I'd argue `name` is extremely vague if you didn't know it was an alias of `display_name` - I know many people would just consider it an alias of `username` or `tag` instead.
> 
> Plus, I'm against added aliases like this in general. There's a purpose to aliases (largely when names can be spelt differently across different variants of English), but short aliases like this feel unnecessary when you can do `name = user.display_name` if you really needed a short name.
> 
> I'm against this PR, but I'll let other contributors decide.

Agreed

mifuyutsuki:
I'm also against this alias since `name` alone would be ambiguous here. It could be any of `nickname`, `global_name`, and `username` (actual fields in the Discord API). Or library-defined fields `display_name` and `tag`. The current `display_name` is good enough because it states its specific meaning: *name* as *displayed* by the Discord client.

AstreaTSS:
I think these responses are enough to close this PR. Apologies - I understand `display_name` is wordy, but it's for the best.

PR #1733, opened 2024-08-16
fix: correctly serialize slash cmd with same names from different scope

AstreaTSS:
## Pull Request Type

- [ ] Feature addition
- [x] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
Trying to make two non-sub slash commands with the same name but a different `scopes` results in an error message when trying to sync these slash commands.

This is because our `application_commands_to_dict`, if multiple slash commands under the same name were found, only used the scope of the first command in the list for serialization purposes.

This later causes issues between the dict generated by this command and `bot.interactions_by_scope` since `interactions_by_scope` now has a command in a scope that `application_commands_to_dict` did not account for, causing an error when syncing.

This PR fixes all of that by making sure all slash commands with the same name are properly serialized. Because I was there already, this PR also refactors `application_commands_to_dict` a tiny bit to be nicer.

## Changes
- Include all commands in `cmd_list` (a list representing slash commands with the same name) in the final, serialized result instead of only taking the scope of the first command and ignoring other commands.
- Did a little refactoring to `application_commands_to_dict` since I was here anyways.

## Related Issues
N/A

## Test Scenarios
```python
import interactions as ipy

bot = ipy.Client(sync_interactions=True, delete_unused_application_cmds=True, ...)

@ipy.slash_command(name="test", scopes=[GUILD_1])
async def test(ctx: ipy.SlashContext):
    await ctx.send("Hello, world!")

@ipy.slash_command(name="test", scopes=[GUILD_2])
async def test2(ctx: ipy.SlashContext):
    await ctx.send("Hello, world!")

bot.start()
```

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/models/internal/application_commands.py
+++ b/interactions/models/internal/application_commands.py
@@ -1,4 +1,5 @@
 import asyncio
+from collections import defaultdict
 import inspect
 import re
 import typing
@@ -1466,9 +1467,9 @@ def application_commands_to_dict(  # noqa: C901
     `Client.interactions` should be the variable passed to this
 
     """
-    cmd_bases = {}  # {cmd_base: [commands]}
+    cmd_bases: defaultdict[str, list[InteractionCommand]] = defaultdict(list)  # {cmd_base: [commands]}
     """A store of commands organised by their base command"""
-    output = {}
+    output: defaultdict["Snowflake_Type", list[dict]] = defaultdict(list)
     """The output dictionary"""
 
     def squash_subcommand(subcommands: List) -> Dict:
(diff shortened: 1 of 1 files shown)


AstreaTSS:
Subcommands do not have this fix. Frankly, I'm unsure how this would even be fixed with them, and it might be impossible without some major refactorings, so it has been left as is.

syntaxsdev (approved):
thanks for this fix, I've tested this for this behavior in my environment and it performs as intended.

syntaxsdev:
bump, need approval :)

silasary (approved):

PR #1743, opened 2024-11-14
feat: add guild_count property

AstreaTSS:
## Pull Request Type

- [x] Feature addition
- [ ] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
This PR adds a `guild_count` property to both `Client` and `ClientUser`. It does exactly as you may expect - it gets the number of guilds the client is in. Compared to `len(bot.guilds)`, this method avoids using cache lookups, which takes up a significant amount of time as the number of guilds a bot is in grows.

The debug extension has been modified to use this property too, although the websocket ready instance was left as-is (sans a comment) since that could result in differing behavior - there are cases where `len(bot.guilds)` is not `bot.guild_count`, most notably if the guild is not in the cache. For normal users, though, this difference likely doesn't matter.

## Changes
- Added `bot.guild_count`.
- Added `bot.user.guild_count`.
- Make the debug extension use `guild_count`.
- Add a TODO to potentially make the websocket ready code use this new property, if only after extensive testing.

## Related Issues
N/A

## Test Scenarios
```python
import interactions as ipy

@ipy.slash_command()
async def test(ctx: ipy.SlashContext):
    await ctx.send(ctx.bot.guild_count)
```

## Python Compatibility
A variant of this code has been working for over a year for the [Realms Playerlist Bot](https://github.com/AstreaTSS/RealmsPlayerlistBot).

## Checklist


silasary (approved):

interactions/client/client.py
```diff
@@ -867,7 +877,9 @@ async def _on_websocket_ready(self, event: events.RawGatewayEvent) -> None:
         self._user._add_guilds(expected_guilds)
 
         if not self._startup:
-            while len(self.guilds) != len(expected_guilds):
+            while len(self.guilds) != len(
+                expected_guilds
+            ):  # TODO: potentially use self.guild_count instead of len(self.guilds)
```
silasary: We explictly don't want to use guild_count here, as this behaviour relies on unloaded guilds not being listed.

AstreaTSS: Comment has been removed.

PR #1751, opened 2025-01-25
Fix timeout for ephemeral paginators and hybrid commands

AutonomousCat:
## Pull Request Type
- [ ] Feature addition
- [x] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
While I was writing a command that uses ephemeral paginators, I noticed the timeout resulted in a 404 NotFound. While looking for ways to resolve this, I noticed that HybridContext required a `message` object in it's `edit()` function, so I used `_prefixed_ctx` to determine HybridContext. Probably not the best way to handle this, but I did what I could.

## Changes
- Added `edit` field to `Timeout` class to store the message edit coroutine
- Updated `send()` and `reply()` to use the appropriate edit method based on command type

## Test Scenarios
I tested with ephemeral, non-ephemeral, send, reply, slash, hybrid, and prefixed. Now works in all cases without errors.

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/ext/paginators.py
+++ b/interactions/ext/paginators.py
@@ -381,9 +381,11 @@ class Paginator:
         self._message = await ctx.send(**self.to_dict(), **kwargs)
         self._author_id = ctx.author.id
 
-        if hasattr(ctx, "_prefixed_ctx"): # HybridContext
+        if hasattr(ctx, "token"):  # SlashContext
+            edit = ctx.edit
+        elif hasattr(ctx, "_prefixed_ctx"):  # HybridContext
             edit = partial(ctx.edit, message=self._message)
-        else:
+        else:  # PrefixedContext
             edit = self._message.edit
 
         if self.timeout_interval > 1:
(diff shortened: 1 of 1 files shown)


interactions/ext/paginators.py
```diff
@@ -38,6 +39,8 @@ class Timeout:
         repr=False,
     )
     """The paginator that this timeout is associated with."""
+    edit: Callable[..., Coroutine] = attrs.field(repr=False, default=None)
```
mifuyutsuki, line 42: Any reason to pass `edit` to `Timeout` instead of, say, `ctx`? At least that way you can forgo the if-else chain added in `send()` below to determine the right `edit()` by just using `ctx.edit()` regardless of command type, as have been done by the `_on_button()` callback. (Alternatively, add `ctx` to `Paginator` attributes instead of `Timeout`, then use `self.paginator.ctx.edit()`.)

AutonomousCat: @mifuyutsuki HybridContext requires `message`. Storing `ctx` was the first thing I tried, but when testing, HybridContext types would error out since the `message` param was missing.

silasary: I'm trying to figure out why HybridContext needs the message parameter in the first place.

AstreaTSS: This is likely because said context was made before `SlashContext` had a default for `message`. It appears when a default was added, `HybridContext` was missed.

PR #1783, opened 2025-09-18
feat: allow list operations on ContainerComponent

AstreaTSS:
## Pull Request Type

- [x] Feature addition
- [ ] Bugfix
- [ ] Documentation update
- [ ] Code refactor
- [ ] Tests improvement

## Description
This PR allows using `ContainerComponent` similarly to a list.

Previously, if someone wanted to edit the components in a container after it was created, they would have to do something like:
```python
container = ContainerComponent(...)
container.components.append(...)
```

This is cumbersome for a common operation. This PR makes it so a shorthand can be used instead:
```python
container = ContainerComponent(...)
container.append(...)
```

The old method is still able to be used.

## Changes
- Make `ContainerComponent` into a subtype of `collections.UserList`.
  - To keep the `__init__` the same, `UserList` requires some function overriding have to be done to handle the multi-argument init.
  - `ContainerComponent` now has a `data` field because `UserList` uses that under the hood for list operations. `components` is now an alias of `data`.

## Test Scenarios
```python
container = ContainerComponent(...)
container.append(...)
```

## Python Compatibility

## Checklist

### DIFF
--- a/interactions/models/discord/components.py
+++ b/interactions/models/discord/components.py
@@ -1,7 +1,9 @@
+from collections import UserList
 import contextlib
 import uuid
 from abc import abstractmethod
-from typing import Any, Dict, Iterator, List, Optional, Sequence, Union, TYPE_CHECKING
+from typing import Any, Dict, Iterator, List, Optional, Sequence, Union, TYPE_CHECKING, overload
+from typing_extensions import Self
 
 import attrs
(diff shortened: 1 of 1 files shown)


silasary (approved):
I had to look up how UserList works, but yeah, this looks good 😀

